PR #53, opened 2024-08-23
Replace model / data with public HF path, update readme

JasonZhu1313:
## Summary

Replace model / data with HF path, update readme

Replace model / data with HF path, update readme
## Testing Done

Done


ByronHsu (commented):
Great! Do we want to update with bs=6 benchmark result?

ByronHsu (approved):

Issue #74, opened 2024-08-24
[bug fix] Gemma needs RMSNorm with 1.0 offset


yundai424:
### 🐛 Describe the bug

Right now liger straightly applies llama-style rmsnorm impl to gemma. However gemma has a `1.0` offset to the weight https://github.com/huggingface/transformers/blob/v4.44.2/src/transformers/models/gemma/modeling_gemma.py#L122 so it's not correct right now :/

### Versions

0.1.1

PR #85, opened 2024-08-25
Fix Gemma RMSNorm

davidgonmar:
## Summary
Fixes https://github.com/linkedin/Liger-Kernel/issues/74. Allows Gemma's RMSNorm by adding a generic offset parameter to RMSNorm.

## Testing Done
Parametrize the RMSNorm tests to test both Llama's and Gemma's versions (as per Hugging Face's transformers library).

- Hardware Type: tested on NVIDIA L4/L40

## Test outputs
```
test/transformers/test_rms_norm.py ................................                                                                                                                                                                                            [100%]

========================================================================================================================= 32 passed in 3.74s =========================================================================================================================

test/convergence/test_mini_models.py ......                                                                                                                                                                                                                    [100%]

========================================================================================================================= 6 passed in 32.04s =========================================================================================================================
```

### DIFF
--- a/test/transformers/test_rms_norm.py
+++ b/test/transformers/test_rms_norm.py
@@ -24,6 +24,22 @@ class LlamaRMSNorm(nn.Module):
         return self.weight * hidden_states.to(input_dtype)
 
 
+# https://github.com/huggingface/transformers/blob/v4.44.2/src/transformers/models/gemma/modeling_gemma.py#L122
+class GemmaRMSNorm(nn.Module):
+    def __init__(self, hidden_size: int, eps: float = 1e-6):
+        super().__init__()
+        self.eps = eps
+        self.weight = nn.Parameter(torch.ones(hidden_size))
+
+    def _norm(self, x):
+        return x * torch.rsqrt(x.pow(2).mean(-1, keepdim=True) + self.eps)
+
+    def forward(self, x):
+        output = self._norm(x.float())
+        output = output * (1.0 + self.weight.float())
+        return output.type_as(x)
+
+
 @pytest.mark.parametrize(
     "bs, sl, hd",
     [
@@ -45,7 +61,14 @@ class LlamaRMSNorm(nn.Module):
         (torch.bfloat16, 5.0, 1e-5),
     ],
 )
-def test_correctness(bs, sl, hd, dtype, atol, rtol):
+@pytest.mark.parametrize(
+    "reference, offset",
+    [
+        (LlamaRMSNorm, 0.0),
+        (GemmaRMSNorm, 1.0),
+    ],
+)
+def test_correctness(bs, sl, hd, dtype, atol, rtol, reference, offset):
     # h
     _tensor = torch.randn(bs, sl, hd, device="cuda", dtype=dtype)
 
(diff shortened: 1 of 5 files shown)


ByronHsu (commented):
Awesome! Our first kernel related PR :-) Left a comment

test/transformers/test_rms_norm.py
```diff
@@ -24,6 +24,22 @@ def forward(self, hidden_states):
...
+    def __init__(self, hidden_size: int, eps: float = 1e-6):
+        super().__init__()
+        self.eps = eps
+        self.weight = nn.Parameter(torch.ones(hidden_size))
+
+    def _norm(self, x):
+        return x * torch.rsqrt(x.pow(2).mean(-1, keepdim=True) + self.eps)
+
+    def forward(self, x):
+        output = self._norm(x.float())
```
ByronHsu, line 38: @yundai424 I am curious if we need to do the upcasting too in our kernel? or we don't because anyway it gets casted to x.dtype. Our old impl doesn't have that but the result is still consistent

ByronHsu:
May you join our discord and say hi? The PR looks solid and we would love more contribution from you https://discord.gg/CX2YmNmn

ByronHsu:
Also can you add your hardware type in the PR description? I just updated PR template

test/transformers/test_rms_norm.py
davidgonmar: I agree that it might not be the exact same, but in the tests, the results are consistent with Gemma too (up to the testing tolerance). I think it's a tradeoff between exact reproduction and complexity (because I suppose different models might have different small details like this casting thing (llama casting != gemma casting), so handling every variation might end up with a lot of code in the future.

davidgonmar: > @yundai424 I am curious if we need to do the upcasting too in our kernel? or we don't because anyway it gets casted to x.dtype. Our old impl doesn't have that but the result is still consistent

The test results seem to be consistent with the reference too. I think it's a tradeoff between complexity (because some models have slight inconsistencies) and exact reproduction. Let me know what you think and I can modify it (or even on another PR, to keep things clean)

ByronHsu: lgtm! waiting @yundai424 to final check

ByronHsu: can we add gemma to convergence test as well?

davidgonmar: > can we add gemma to convergence test as well?

Sure. Do you think it's better to do it in another PR?

davidgonmar: > can we add gemma to convergence test as well?

adding it

davidgonmar: > can we add gemma to convergence test as well?

done. as mentioned in the discord channel, I made the tolerance for bfloat16 convergence test a bit bigger, since the casting difference seems to be slightly affecting it (for the regular tests it's fine, but it seems to add up in the end-to-end training situation)

yundai424 (approved):
LGTM, thanks for the contribution! Let's create a follow-up issue for the dtype thing

austin362667 (approved):

PR #94, opened 2024-08-26
Update test_cross_entropy.py

lancerts:
Update test_cross_entropy.py, make the threshold more readable

### DIFF
--- a/test/transformers/test_cross_entropy.py
+++ b/test/transformers/test_cross_entropy.py
@@ -194,7 +194,7 @@ def _full_pass_once(B, T, V):
     ],
 )
 @pytest.mark.skipif(
-    torch.cuda.get_device_properties(0).total_memory < 64000000000,
+    torch.cuda.get_device_properties(0).total_memory < 64 * 1000 * 1000 * 1000, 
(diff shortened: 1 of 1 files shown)


misbahkhalilaz (approved):
LGTM!

test/transformers/test_cross_entropy.py
```diff
@@ -194,7 +194,7 @@ def _full_pass_once(B, T, V):
     ],
 )
 @pytest.mark.skipif(
-    torch.cuda.get_device_properties(0).total_memory < 64000000000,
+    torch.cuda.get_device_properties(0).total_memory < 64 * 1000 * 1000 * 1000,
```
pchng, line 197: If readability is the concern, you can also use the underscore separator to delineate large numbers:

```
>>> 64_000_000_000 == 64 * 1000 * 1000 * 1000
True
```

Obviously just subjective though.

yundai424 (approved):

test/transformers/test_cross_entropy.py
ByronHsu: TIL!

PR #95, opened 2024-08-26
pytest without need to dealing with PYTHONPATH

helloworld1:
## Summary
pytest without need to dealing with PYTHONPATH

According to [pytest doc](https://docs.pytest.org/en/7.1.x/explanation/pythonpath.html#invoking-pytest-versus-python-m-pytest), `python -m pytest` will put current dir to sys.path, eliminating the need to add `PYTHONPATH=.` to make test passing

## Testing Done
make

- Hardware Type: AMD 7900 XTX

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,11 @@
-.PHONY: test checkstyle test-convergence
+.PHONY: test checkstyle test-convergence all
+
+
+all: test test-convergence checkstyle
 
 # Command to run pytest for correctness tests
 test:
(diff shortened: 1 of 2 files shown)


lancerts (approved):

Makefile
```diff
@@ -1,8 +1,11 @@
-.PHONY: test checkstyle test-convergence
+.PHONY: test checkstyle test-convergence all
+
+
+all: test test-convergence checkstyle
```
ByronHsu, line 4: nice

PR #78, opened 2024-08-25
Remove unused images in `examples/medusa/docs/images/`

pchng:
## Summary
- There are four images I believe are unused in `examples/medusa/docs/images/`
- These images were replaced by updated benchmark results in this PR: https://github.com/linkedin/Liger-Kernel/pull/59/
- Furthermore these images have filenames that only differ by one letter case, causing issues on cloning the repo on case-insensitive filesystems like those used on Windows and MacOS. (Linux not affected) You would see an error like this:

```
git clone git@github.com:pchng/Liger-Kernel.git
Cloning into 'Liger-Kernel'...
remote: Enumerating objects: 735, done.
remote: Counting objects: 100% (342/342), done.
remote: Compressing objects: 100% (213/213), done.
remote: Total 735 (delta 211), reused 213 (delta 121), pack-reused 393 (from 1)
Receiving objects: 100% (735/735), 8.18 MiB | 11.39 MiB/s, done.
Resolving deltas: 100% (306/306), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'examples/medusa/docs/images/Memory_Stage1_num_head_3.png'
  'examples/medusa/docs/images/memory_Stage1_num_head_3.png'
  'examples/medusa/docs/images/Memory_Stage1_num_head_5.png'
  'examples/medusa/docs/images/memory_Stage1_num_head_5.png'
  'examples/medusa/docs/images/Memory_Stage2_num_head_3.png'
  'examples/medusa/docs/images/memory_Stage2_num_head_3.png'
  'examples/medusa/docs/images/Memory_Stage2_num_head_5.png'
  'examples/medusa/docs/images/memory_Stage2_num_head_5.png'
```

## Testing Done
- Ran `grep -r "memory_Stage" ./*`, didn't find references to the deleted image files.
- The relevant `README.md` file still looks OK after removing these unused images: https://github.com/pchng/Liger-Kernel/blob/50345cfb881388ce4de44210e08bd8c43cf1ed04/examples/medusa/README.md

PR #107, opened 2024-08-26
Remove duplicate images

qingquansong:
## Summary

Same problem still appears after #78 

## Testing Done

- Hardware Type: <BLANK>


JasonZhu1313 (approved):
Thanks, the upper case is what we needed, so we could delete the unused image for bs = 4 setup

Issue #70, opened 2024-08-24
Model Agnostic Patching?


lapp0:
### 🚀 The feature, motivation and pitch

Liger is missing support for a number of language model architectures. Additionally, I'd like to be able to patch any model without consideration for the model architecture using a simple `apply_liger_kernel_to_all_modules(pretrained_model)`

### Additional context

The transformers library isn't very good at sharing code between identical implementations. For example [`LlamaRMSNorm`](https://github.com/huggingface/transformers/blob/0a7af19f4dc868bafc82f35eb7e8d13bac87a594/src/transformers/models/llama/modeling_llama.py#L112-L129) is exactly the same as [`MistralRMSNorm`](https://github.com/huggingface/transformers/blob/0a7af19f4dc868bafc82f35eb7e8d13bac87a594/src/transformers/models/mistral/modeling_mistral.py#L61-L78)

Provided the lack of code sharing in transformers, perhaps a hack like this would work for checking whether to apply a patch to a module?

```
>>> from transformers.models.llama import modeling_llama
>>> from transformers.models.mistral import modeling_mistral
>>> from transformers.models.phi3 import modeling_phi3
>>> 
>>> def is_equivalent_module(module0, module1):
...     return (
...         module0.__init__.__code__.co_code == module1.__init__.__code__.co_code
...         and module0.forward.__code__.co_code == module1.forward.__code__.co_code
...     )

>>> is_equivalent_module(modeling_llama.LlamaRMSNorm, modeling_mistral.MistralRMSNorm)
True
>>> is_equivalent_module(modeling_llama.LlamaRMSNorm, modeling_phi3.Phi3RMSNorm)
True
```

ByronHsu:
@robertgshaw2-neuralmagic has proposed a cool idea. We can have a `AutoLigerModelForCausalLM` class which auto detects the model type and apply the kernels, like in [AutoAWQ](https://github.com/casper-hansen/AutoAWQ)

```
model = transformers.AutoLigerModelForCausalLM.from_pretrained("<some llama model>")
```

cc @yundai424 @qingquansong @robertgshaw2-neuralmagic  thoughts?

ByronHsu:
https://github.com/casper-hansen/AutoAWQ/blob/6f14fc7436d9a3fb5fc69299e4eb37db4ee9c891/awq/models/auto.py#L50

yundai424:
[...]

PR #115, opened 2024-08-27
Adding AutoLigerKernelForCausalLM

shimizust:
## Summary
- Adding AutoLigerKernelForCausalLM to automatically apply relevant liger kernels to corresponding model to address #70 
- Args passed to auto model that are part of `apply_liger_kernel_to_{model_type}` supported args are passed through
- Refactored the monkey_patch.py so the mapping and patching functions are all in the same file so it's easier to not miss updating. Also added test to ensure the patching functions match the auto model mapping.

Example:
```python
model = AutoLigerKernelForCausalLM.from_pretrained(
        custom_args.model_name,
        trust_remote_code=True,
        use_cache=False,
        torch_dtype=torch.bfloat16,
        cross_entropy=True,
        fused_linear_cross_entropy=False,
    )

> INFO:liger_kernel.transformers.monkey_patch:Applying Liger kernels for model type: llama with kwargs: {'cross_entropy': True, 'fused_linear_cross_entropy': False}
```

## Testing Done
- Run huggingface example

```python
assert modeling_llama.LlamaRMSNorm != LigerRMSNorm
model = AutoLigerKernelForCausalLM.from_pretrained(
    custom_args.model_name,
    trust_remote_code=True,
    use_cache=False,
    torch_dtype=torch.bfloat16,
    # These args will get passed to the appropriate apply_liger_kernel_to_* function
    # to override the default settings
    cross_entropy=True,
    fused_linear_cross_entropy=False,
)
assert modeling_llama.LlamaRMSNorm == LigerRMSNorm
```

- Hardware Type: A100


```
$ make checkstyle
flake8 .; flake8_status=$?; \
isort .; isort_status=$?; \
black .; black_status=$?; \
if [ $flake8_status -ne 0 ] || [ $isort_status -ne 0 ] || [ $black_status -ne 0 ]; then \
        exit 1; \
fi
Skipped 3 files
All done! ✨ 🍰 ✨
54 files left unchanged.

$ make test
pytest --disable-warnings test/ --ignore=test/convergence
==================================================================================================================================== test session starts ====================================================================================================================================
platform linux -- Python 3.10.14, pytest-7.1.2, pluggy-1.0.0
rootdir: /home/jobuser/Liger-Kernel
plugins: lipy-config-base-32.0.27, lipy-fabric-35.2.19, lipy-test-8.0.66, datadir-1.3.1
collected 117 items                                                                                                                                                                                                                                                                         

test/transformers/test_auto_model.py .                                                                                                                                                                                                                                                [  0%]
test/transformers/test_cross_entropy.py ..........................................................                                                                                                                                                                                    [ 50%]
test/transformers/test_fused_linear_cross_entropy.py ......                                                                                                                                                                                                                           [ 55%]
test/transformers/test_geglu.py ........                                                                                                                                                                                                                                              [ 62%]
test/transformers/test_monkey_patch.py .....                                                                                                                                                                                                                                          [ 66%]
test/transformers/test_rms_norm.py ................                                                                                                                                                                                                                                   [ 80%]
test/transformers/test_rope.py ............                                                                                                                                                                                                                                           [ 90%]
test/transformers/test_swiglu.py ........                                                                                                                                                                                                                                             [ 97%]
test/transformers/test_trainer_integration.py .                                                                                                                                                                                                                                       [ 98%]
test/triton/test_triton_monkey_patch.py ..                                                                                                                                                                                                                                            [100%]

=================================================================================================================================== 117 passed in 50.75s ====================================================================================================================================

$ make test-convergence
HF_DATASETS_OFFLINE=1 pytest --disable-warnings test/convergence
==================================================================================================================================== test session starts ====================================================================================================================================
platform linux -- Python 3.10.14, pytest-7.1.2, pluggy-1.0.0
rootdir: /home/jobuser/Liger-Kernel
plugins: lipy-config-base-32.0.27, lipy-fabric-35.2.19, lipy-test-8.0.66, datadir-1.3.1
collected 8 items                                                                                                                                                                                                                                                                           

test/convergence/test_mini_models.py ......                                                                                                                                                                                                                                           [ 75%]
test/convergence/test_mini_models_no_logits.py ..                                                                                                                                                                                                                                     [100%]

==================================================================================================================================== 8 passed in 47.97s =====================================================================================================================================

```


JasonZhu1313 (commented):
Thanks for contribution, love the tests here!

src/liger_kernel/transformers/monkey_patch.py
```diff
@@ -212,3 +216,54 @@ def apply_liger_kernel_to_phi3(
...
+        - model_type: the model types as defined in transformers/models/auto/modeling_auto.py
+          and specified in the model's config.json
+        - kwargs: keyword arguments that are passed to the corresponding apply_liger_kernel_to_* function.
+    """
+
+    if not model_type:
+        logger.info("Model type was not provided. No Liger kernels will be applied.")
+        return
+
+    if model_type not in MODEL_TYPE_TO_APPLY_LIGER_FN.keys():
```
JasonZhu1313, line 248: Correct me if I am wrong. This still doesn't fully resolve the original request in the issue "Liger is missing support for a number of language model architectures. Additionally, I'd like to be able to patch any model without consideration for the model architecture using a simple apply_liger_kernel_to_all_modules(pretrained_model)" where users want to automatically apply this to appropriate layers / modules if that module is identical to one of supported models.

lapp0 (commented):
Am I reading the changeset correctly - patches will now be applied to objects rather than globally applied to classes?

That's great!

src/liger_kernel/transformers/monkey_patch.py
shimizust: That's correct, discussing with @ByronHsu, comparing module code diff to determine if we can patch is a bit hacky, so want to hold off on that and just maintain the supported model list in the library. In any case, this will be a wrapper on top of any dynamic patching of models if we decide to implement in the future.

Also, this change doesn't provide a function to apply any and all liger kernels globally using default settings. I think that could be valuable so users could just globally set this at the beginning of their training script and not worry if they are using different AutoModels*/different model types. If we do, we can have it in a separate PR.

I don't believe there would be conflicts? For example, even though we throw an error if `fused_linear_cross_entropy` and `cross_entropy` are both set to True, I don't think it would actually cause a conflict to patch both of them. Thoughts @ByronHsu ?

shimizust:
> Am I reading the changeset correctly - patches will now be applied to objects rather than globally applied to classes?
> 
> That's great!

That's a great point, this still globally applies the changes to classes--it just will automatically patch the correct model. It would make more sense for AutoModel to only patch the model objects to remove any potential global side effects. I think this may require some more involved refactoring of the existing code, so we can look into it in a subsequent PR.

cc @ByronHsu

ByronHsu:
@shimizust @lapp0 i prefer using the current monkey patch way rather than object swapping, the later one is quite involved and need nasty for loop. i don't see major issues for the current approach

src/liger_kernel/transformers/monkey_patch.py
ByronHsu: After discussion w steven. We want to have two high-level APIs now

1. existing per model patching
2. autoligerclass

which emphasizes on ease of use and convenience since this is the reason people like liger kernel.

for others, we can revisit based on community's demand. we don't want to push everything out at time or users will get confused.

tyler-romero:
Maybe this is out of scope for this PR, but the AutoLigerKernel approach misses a couple of use cases, for example [LlavaNext](https://github.com/huggingface/transformers/blob/v4.44.2/src/transformers/models/llava_next/modeling_llava_next.py#L356), where a Llama model is being loaded during the construction of the outer Llava model.

Perhaps the `AutoModelForCausalLM` HF class itself could be monkeypatched with `AutoLigerKernelForCausalLM` so that there is a single line of code that converts existing HF code to use liger kernel if applicable.

shimizust:
> Maybe this is out of scope for this PR, but the AutoLigerKernel approach misses a couple of use cases, for example [LlavaNext](https://github.com/huggingface/transformers/blob/v4.44.2/src/transformers/models/llava_next/modeling_llava_next.py#L356), where a Llama model is being loaded during the construction of the outer Llava model.
> 
> Perhaps the `AutoModelForCausalLM` HF class itself could be monkeypatched with `AutoLigerKernelForCausalLM` so that there is a single line of code that converts existing HF code to use liger kernel if applicable.

That's an interesting use case where the model has sub-models which could be patched. This PR is meant for the most common use cases and also misses folks who are using some other AutoModel variant (e.g. AutoModelForTokenClassification).

I think for the time being you'd have to just put `apply_liger_kernel_to_llama()` at the top of your training code. Also noting this would be another use case to have an `apply_all_liger_kernels()` function.

ByronHsu:
yes. for any advanced use case. please compose the model your own. lol, or write your own patching

ByronHsu (approved):

ByronHsu (approved):

PR #117, opened 2024-08-27
Update rms_norm.py

lancerts:
## Summary
Correct the comment in the code.

### DIFF
--- a/src/liger_kernel/ops/rms_norm.py
+++ b/src/liger_kernel/ops/rms_norm.py
@@ -203,5 +203,6 @@ class LigerRMSNormFunction(torch.autograd.Function):
             num_warps=ctx.num_warps,
         )
         dX = dY.view(*shape)
+        # summation dW over BxT dimension
         dW = torch.sum(dW, dim=0)
         return dX, dW, None, None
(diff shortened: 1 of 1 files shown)


src/liger_kernel/ops/rms_norm.py
```diff
@@ -88,7 +88,7 @@ def _rms_norm_backward(
 ):
     """
     dx = (1 / RMS) * [dy * w  - (1 / N) * (1 / RMS^2) * ((dy * w) dot x) * x]. * means element-wise multiplication, whileas dot means dot product
-    dw = sum(dy * (x / RMS)). summation over BxT dimension
+    dw = dy * (x / RMS)
```
ByronHsu, line 91: should we emphasize the summation too? important too

lancerts: https://github.com/linkedin/Liger-Kernel/pull/117/files#diff-73d2c774174032bab7bb8bb965069f1d3345993ad0943b2d50c18d091da70dbeR122 we are not doing summation in the triton kernel @ByronHsu

ByronHsu: i got your point, but it is important to let users know why we need sum https://github.com/linkedin/Liger-Kernel/blob/8f0c8413096a813099ddaa097aba285e2b311d00/src/liger_kernel/ops/rms_norm.py#L206. maybe we can add at line 206

lancerts: done

PR #111, opened 2024-08-26
Add FusedLinearCrossEntropy to Gemma

Luke-Chesley:
## Summary

This PR adds FusedLinearCrossEntropy support for gemma to resolve
issue https://github.com/linkedin/Liger-Kernel/issues/101.

## Details
I based the code in this PR off of https://github.com/linkedin/Liger-Kernel/pull/93 which does the same thing for Mistral. 

Since the parameters fused_linear_cross_entropy and cross_entropy cannot both be true, fused_linear_cross_entropy has to be explicitly set to False in test_mini_models.py, to avoid failing convergence
test due to setting both True.
In test_mini_models_no_logits.py, I add two extra tests for gemma to ensure that FusedLinearCrossEntropy part in gemma is tested.

## Testing Done

- Hardware Type: NVIDIA A100-SXM4-80GB 
- [x] run `make test` to ensure correctness
- [x] run `make checkstyle` to ensure code style
- [x] run `make test-convergence` to ensure convergence

```
Fixing ~/Liger-Kernel/test/convergence/test_mini_models_no_logits.py
Skipped 1 files
reformatted /home/luke/Desktop/Liger-Kernel/src/liger_kernel/transformers/model/gemma.py
reformatted /home/luke/Desktop/Liger-Kernel/src/liger_kernel/transformers/monkey_patch.py
reformatted /home/luke/Desktop/Liger-Kernel/test/convergence/test_mini_models_no_logits.py

All done! ✨ 🍰 ✨
3 files reformatted, 51 files left unchanged.
make: *** [Makefile:14: checkstyle] Error 1
luke@luke:~/Desktop/Liger-Kernel$ 


root@C.12106734:~/Liger-Kernel$ make test
python -m pytest --disable-warnings test/ --ignore=test/convergence
======================================================= test session starts =======================================================
platform linux -- Python 3.11.9, pytest-8.3.2, pluggy-1.5.0
rootdir: /root/Liger-Kernel
plugins: hypothesis-6.108.4, anyio-4.4.0
collected 130 items                                                                                                               

test/transformers/test_cross_entropy.py ..........................................................                          [ 44%]
test/transformers/test_fused_linear_cross_entropy.py ......                                                                 [ 49%]
test/transformers/test_geglu.py ........                                                                                    [ 55%]
test/transformers/test_rms_norm.py ................................                                                         [ 80%]
test/transformers/test_rope.py ............                                                                                 [ 89%]
test/transformers/test_swiglu.py ........                                                                                   [ 95%]
[...]

PR #120, opened 2024-08-27
Add gemma lightning example for single L40 GPU

qingquansong:
## Summary

Add gemma lightning example for single L40 GPU

## Testing Done

- Hardware Type: <BLANK>


qingquansong:
@yundai424 @ByronHsu Waiting for #111 to check in and then can patch fused CE to test memory

yundai424:
@qingquansong do you think if we should just replace llama with gemma2 in this lightning example, and fix the config so that it works with L40 GPU, just to minimize the code 🤔  -- and the example can be used out of the box as lightning studio example

qingquansong:
> @qingquansong do you think if we should just replace llama with gemma2 in this lightning example, and fix the config so that it works with L40 GPU, just to minimize the code 🤔 -- and the example can be used out of the box as lightning studio example

Or maybe set the default one to be Gemma and provide a llama example in readme with config changes?

ByronHsu (approved):

PR #121, opened 2024-08-27
[WIP] Fix confusion on Gemma

yundai424:
## Summary

**why**

1. From config point of view, Gemma1 is doing exact GeLU [[ref](https://huggingface.co/google/gemma-2b/blob/main/config.json#L10)] but gemma 1.1 and 2 are doing approximate gelu ([gemma2](https://huggingface.co/google/gemma-2-2b/blob/main/config.json#L14), [gemma1](https://huggingface.co/google/gemma-1.1-2b-it/blob/main/config.json#L10))
2. also, gemma uses `hidden_activation` config field and `hidden_act` is ignored ([gemma1 code](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma/modeling_gemma.py#L170), [gemma2 code](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma2/modeling_gemma2.py#L202))

That being said, we should be fine to claim that all of gemma 1, 1.1 and 2 are supported. But for safety i think we can first go with 1.1 and 2 first

**what**
1. adjust the monkey patch so it works properly with gemma2 code base too
2. checkstyle
3. **[TODO] add final logit softcapping for gemma2 https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma2/modeling_gemma2.py#L1054**

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/transformers/monkey_patch.py
+++ b/src/liger_kernel/transformers/monkey_patch.py
@@ -129,8 +129,8 @@ def apply_liger_kernel_to_gemma(
     geglu: bool = True,
 ) -> None:
     """
-    Apply Liger kernels to replace original implementation in HuggingFace Gemma2 models
-    to make GPU go burrr.
+    Apply Liger kernels to replace original implementation in HuggingFace Gemma models
+    (Gemma 1.1 and Gemma 2 supported) to make GPU go burrr.
 
     Args:
         rope (bool): Whether to apply Liger's rotary position embedding. Default is True.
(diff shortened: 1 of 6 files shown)


test/convergence/test_mini_models_no_logits.py
```diff
@@ -241,16 +270,6 @@ def test_mini_model(
         rtol=loss_rtol,
     )
 
-    # No logits are materialized
```
ByronHsu, line 244: minor suggestion: we can keep this line since this is quite important

src/liger_kernel/transformers/monkey_patch.py
```diff
@@ -141,20 +141,28 @@ def apply_liger_kernel_to_gemma(
...
+        )
     if cross_entropy:
         modeling_gemma.CrossEntropyLoss = LigerCrossEntropyLoss
+        modeling_gemma2.CrossEntropyLoss = LigerCrossEntropyLoss
     if geglu:
         modeling_gemma.GemmaMLP = LigerGEGLUMLP
+        modeling_gemma2.Gemma2MLP = LigerGEGLUMLP
     if fused_linear_cross_entropy:
         modeling_gemma.GemmaForCausalLM.forward = gemma_lce_forward
+        modeling_gemma2.Gemma2ForCausalLM.forward = gemma_lce_forward
```
DocShotgun, line 165: It looks like the same recently added lce_forward is being used for both Gemma and Gemma2 (https://github.com/linkedin/Liger-Kernel/pull/111).

It appears that there is a slight difference in the `forward` between Gemma and Gemma2 in the modeling code in transformers, specifically regarding logit softcapping (https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma2/modeling_gemma2.py#L1054-L1057), which doesn't seem to be accounted for in the new lce code as far as I can tell. Would this potentially lead to incompatibility?

yundai424: yeah exactly, just figured out we need to add softcapping :/ i'll have a follow up PR to clarify that right now this only works for gamma 1 and 1.1 and we'll work on changing the kernel correspondingly asap

PR #125, opened 2024-08-27
Fix gemma 1 and 2 support

yundai424:
## Summary

Right now for gemma:
- Gemma1 and 1.1 can be fully supported (FLCE included). Gemma 1, though using `hidden_act="gelu"` in their config, is actually gonna use tanh approximation because this is ignored in [model code](https://github.com/huggingface/transformers/blob/v4.40.1/src/transformers/models/gemma/modeling_gemma.py#L175)
- Gemma2 can be supported without FLCE. That's because Gemma2 has a soft capping after `lm_head`. Thanks a lot @DocShotgun for pointing this out! We have a TODO item for doing it properly.
- 

## Testing Done

- Hardware Type: <BLANK>


qingquansong (approved):
LGTM! (maybe later when having more models, we can remove some gamma testing to reduce test time?)

yundai424:
yeah eventually we can just keep gemma2 🤔

PR #128, opened 2024-08-27
fallback to torch native linear+CE when without label

### LINKED ISSUE #113: Issue Using TRL DPOTrainer with Liger Kernels
### 🐛 Describe the bug

The concatenated forward in the standard `DPOTrainer` in TRL does not work with Liger Kernels. This is because the concatenated forward call in this function seen [here](https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L1306-L1311) does not supply the labels field. This leads to the following error:

```bash
liger_kernel/transformers/model/llama.py", line 102, in lce_forward
    shift_labels = labels[..., 1:].contiguous()
TypeError: 'NoneType' object is not subscriptable
```

### Reproduce

Take the standard [DPO script in TRL](https://github.com/huggingface/trl/blob/main/examples/scripts/dpo.py) and set the model_id to `meta-llama/Meta-Llama-3.1-8B-Instruct` or something. Next apply the Liger kernel monkey patch and run the script. It will crash for the labels being `None` in the `concatenated_forward` call.

### Versions

Liger stable

yundai424:
## Summary

fixes #113 

## Testing Done

- Hardware Type: <BLANK>


ByronHsu (commented):
maybe self.training is True and labels is not None is better?

src/liger_kernel/transformers/model/gemma.py
```diff
@@ -97,7 +97,7 @@ def lce_forward(
     loss = None
     logits = None
 
-    if self.training:
+    if self.training and labels is not None:
```
ByronHsu, line 100:
```suggestion
    if self.training and (labels is not None):
```

truly the **last** suggestion!

ByronHsu (commented):
last minor suggestion!

yundai424 (commented):
update

ByronHsu (approved):
critical bug fixes!

Issue #131, opened 2024-08-27
Add support for encoder-only transformers (e.g. BERT)


OxxoCodes:
### 🚀 The feature, motivation and pitch

Liger Kernel is currently incompatible with encoder-only transformer architectures such as BERT, DistilBERT, RoBERTa, XLM-R, and DeBERTa.

Given the importance these models still have in research and industry use-cases, it would be great to see support added to further decrease memory requirements and increase training throughput.

roy6n23:
Is encoder-only support (BERT/RoBERTa/DeBERTa-class) in scope for Liger?
I don't see any encoder entries in monkey_patch.py, so I want to check before
investing. If it is in scope I'm interested in taking it — I run BERT-class
encoder inference in production and have a concrete use case.

PR #135, opened 2024-08-28
custom Embedding kernel

AndreSlavescu:
## Summary
- Added Embedding forward/backwards kernels + LigerEmbedding class which maps to nn.Embedding
- nn.Embedding is useful for encoder-only models such as BERT
- ref: https://github.com/linkedin/Liger-Kernel/issues/131

## Testing Done

- tested against nn.Embedding for correctness on various inputs
- tested with and without padding_idx

- Hardware Type: RTX 3090 + RTX 4090


ByronHsu:
very cool! i am wondering how is the perf compared with torch

benchmark/benchmark_embedding.py
```diff
@@ -0,0 +1,119 @@
...
+            styles=[
+                ("blue", "solid"),
+                ("orange", "solid"),
+            ],
+            ylabel="time (ms)",
+            plot_name="embedding-fwd-speed-benchmark",
+            args={
+                "B": 32, 
+                "T": 512, 
+                "D": 768,
```
ByronHsu: very impressive! can you use the param of llama3?

ByronHsu:
just an idea: we can use torch.compile to dump kernel and compare with our impl

src/liger_kernel/ops/embedding.py
```diff
@@ -0,0 +1,143 @@
...
+
+    grad_output = tl.load(
+        grad_output_ptr + offsets_m[:, None] * embedding_dim + offsets_n[None, :],
+        mask=mask_m[:, None] & mask_n[None, :],
+        other=0.0,
+    )
+
+    grad_weight_offsets = indices[:, None] * embedding_dim + offsets_n[None, :]
+
+    tl.atomic_add(
```
qingquansong, line 71: i think this should work otherwise we need to pre-group and routing similar to the layer norm written in flash attn issue. but have a concern here about the multi-gpu case. The atomic add doc scope default to be gpu https://triton-lang.org/main/python-api/generated/triton.language.atomic_add.html so not sure if it will always happen on single host gpu or has other behaviors. May need a test on this.

yundai424: do we need to care about multi GPU case since each GPU is on its own process and the only chance they get communicated is in some nccl operations?

lancerts: @ByronHsu @shimizust can we test the kernel using multi-GPU to see if the default lock scope is fine?

src/liger_kernel/transformers/embedding.py
```diff
@@ -0,0 +1,28 @@
...
+import torch.nn as nn
+
+from liger_kernel.ops.embedding import LigerEmbeddingFunction
+
+
+class LigerEmbedding(nn.Module):
+    def __init__(
+        self, num_embeddings, embedding_dim, padding_idx: Optional[int] = None
+    ):
+        super().__init__()
```
ByronHsu, line 13: We can merge it for now, but add a warning here say this is still not fully optimized

ByronHsu: actually, let's create `src/liger_kernel/ops/experimental`, just like https://github.com/google/jax/tree/main/jax/experimental/mosaic

AndreSlavescu: okay, I'll make that change quickly

AndreSlavescu: made changes @ByronHsu

lancerts (approved):

PR #149, opened 2024-08-28
updated sl to be non-constexpr

AndreSlavescu:
## Summary

updated seqlen for rope to be non-constexpr

sl from constexpr to non-constexpr

## Testing Done

- Hardware Type: RTX 3090


ByronHsu (approved):

yundai424 (approved):
Thanks for the quick fix! Tested with a mini script running on different seq len and validate that this works.

PR #155, opened 2024-08-28
fix convergence test, phi3 import and update benchmark

yundai424:
## Summary

1. bf16 loss rtol is a bit too loose, tighten it by 1 digit
2. slightly loosen gemma1 atol, it's been failing
3. old `transformers` version doesn't carry phi3 source code (testing on 4.40.1), since we claim support for >= 4.40.1, change the import a bit so things still work on older HF ver
4. rerun all benchmark to reflect latest performance in preparation for new release

## Testing Done

- Hardware Type: <BLANK>


ByronHsu (approved):

ByronHsu (commented):
https://github.com/linkedin/Liger-Kernel/blob/c3c56e7f2cc6ce5e1a49f6222a7596368c213a24/setup.py#L33

btw, we actually bump transformers to >= `4.41.0`

src/liger_kernel/transformers/monkey_patch.py
```diff
@@ -268,6 +267,8 @@ def apply_liger_kernel_to_phi3(
     if cross_entropy:
         modeling_phi3.CrossEntropyLoss = LigerCrossEntropyLoss
     if fused_linear_cross_entropy:
+        from liger_kernel.transformers.model.phi3 import lce_forward as phi3_lce_forward
```
ByronHsu, line 270: maybe we can move this back. we don't want to support too old transformers

yundai424: oh! apparently i swapped the `0` and `1`, lemme revert these phi3 related stuff

ByronHsu (approved):

PR #185, opened 2024-08-31
gemm fp8 e4m3

AndreSlavescu:
## Summary

Implemented FP8 gemm with E4M3 representation for FP8.

## Testing Done

tested square matrices of varying sizes (64, 256, 512, 1024) and compared against torch matmul with appropriate casting for backward (torch.matmul doesn't support fp8_e4m3 dtype for backward). 

FP8 gemm will only work on SM_89+

- Hardware Type: RTX 4090


benchmark/gemm_split_k_fp8_speed/gemm-split-k-fp8-full-speed-benchmark.png
lancerts, line 1: why are there multiple lines for each configuration? should be 1 line per case?

src/liger_kernel/ops/experimental/gemm_split_k_fp8_e4m3.py
```diff
@@ -0,0 +1,302 @@
...
+    pid_in_group = pid % width
+    group_size = tl.minimum(grid_m - group_id * group_m, group_m)
+    pid_m = group_id * group_m + (pid_in_group % group_size)
+    pid_n = pid_in_group // group_size
+
+    return pid_m, pid_n
+
+
+@triton.jit
+def gemm_split_k_kernel_forward(
```
lancerts, line 36: can we add some comments on the kernel to help border audience understand what each kernel is doing?

src/liger_kernel/ops/experimental/gemm_split_k_fp8_e4m3.py
```diff
@@ -0,0 +1,302 @@
...
+    tl.atomic_add(c_ptrs, acc, mask=mask)
+
+
+def forward_kernel(ctx, a, b):
+    ctx.save_for_backward(a, b)
+
+    # cast to FP8
+    # structure:
+    #   | 1 bit sign | 4 bit exponent | 3 bit mantissa |
+    a, b = a.to(torch.float8_e4m3fn), b.to(torch.float8_e4m3fn)
```
qingquansong, line 106: Directly casting to FP8 is not a good option since it will cause huge performance drop when our input is not fp8 originally. A preferred way is to do dynamic (or static fp8 quantization) so you can extract a scale factor (in precision fp32) to conduct scaled matmul.

Note that: ultimately we wanna fuse the scaling part into the kernel as well so we can reduce the overhead of quantization and dequantization.

src/liger_kernel/ops/experimental/gemm_split_k_fp8_e4m3.py
```diff
@@ -0,0 +1,302 @@
...
+
+    total_blocks_m = triton.cdiv(m, block_m)
+    total_blocks_n = triton.cdiv(n, block_n)
+    total_programs_mn = total_blocks_m * total_blocks_n
+    total_programs_k = split_k
+
+    grid = (total_programs_mn, total_programs_k)
+
+    c = torch.zeros((m, n), device=a.device, dtype=torch.float16)
+    gemm_split_k_kernel_forward[grid](
```
qingquansong, line 123: Could you add some comments to explain the reason of using split_k implementation such as in which scenario it's preferred?

benchmark/benchmark_gemm_split_k_fp8_e4m3.py
```diff
@@ -0,0 +1,184 @@
...
+    b_fp8 = torch.randn((k, n), device=device, dtype=dtype).to(torch.float8_e4m3fn)
+
+    a_float = a_fp8.float().requires_grad_()
+    b_float = b_fp8.float().requires_grad_()
+
+    def fwd_liger():
+        return LigerFP8GemmSplitKFunction.apply(a_fp8, b_fp8)
+
+    def fwd_torch():
+        return torch.matmul(a_float, b_float)
```
qingquansong, line 72: Computing the fp8 kernel with torch matmul on fp32 is not a quite fair comparison. A better comparison would be to compare with torch._scaled_mm with fp8 matmul such as the example here: https://gist.github.com/malfet/7874d96b99670c3da83cbb779ab770c6

benchmark/benchmark_gemm_split_k_fp8_e4m3.py
```diff
@@ -0,0 +1,184 @@
...
+            line_names=["Liger", "PyTorch", "Torch Compile"],
+            styles=[("blue", "solid"), ("orange", "solid"), ("green", "solid")],
+            ylabel="GPU memory usage (MB)",
+            plot_name="gemm-split-k-fp8-memory-benchmark",
+            args={"dtype": torch.float32},
+        )
+    ]
+)
+def bench_memory_gemm_split_k_fp8(m, k, n, provider, dtype, device="cuda"):
+    a_fp8 = torch.randn((m, k), device=device, dtype=dtype).to(torch.float8_e4m3fn)
```
qingquansong, line 141: ditto. Let's try to create bf16 input and compare the speed/memory of torch._scaled_mm v.s. the fp8 kernel and then compare the joint time of quant + dequant + matmul (with fp8 scaled factor) Thanks!

qingquansong (changes requested):

qingquansong:
Thanks for the efforts! Provided some comments and please take a look and let me know if there're any questions and we can discuss more here or on discord.

src/liger_kernel/ops/experimental/gemm_split_k_fp8_e4m3.py
```diff
@@ -0,0 +1,302 @@
+# adapted from: https://github.com/pytorch-labs/applied-ai/blob/main/kernels/triton/inference/fp8/splitk_gemm_fp8.py
```
yundai424, line 1: should we adapt the BSD3 license header from the original repo? @ByronHsu

src/liger_kernel/ops/utils.py
```diff
@@ -41,6 +42,96 @@ def calculate_settings(n):
...
+
+
+def check_compute_capability_for_fp8(fn):
+    @functools.wraps(fn)
+    def wrapper(ctx, *args, **kwargs):
+        if torch.cuda.is_available():
+            device = torch.cuda.current_device()
+            compute_capability = torch.cuda.get_device_capability(device)
+            if compute_capability[0] >= 9:  # SM_90+
+                os.environ["ENABLE_TMA"] = "1"
```
yundai424, line 119: i'm wondering if we should care about `ENABLE_MMA_V3` too 🤔

src/liger_kernel/ops/experimental/gemm_split_k_fp8_e4m3.py
ByronHsu: we can check with legal internally

src/liger_kernel/ops/utils.py
AndreSlavescu: Yeah I think that's a good idea. The H100 has wgmma units which has much higher throughput than previous generations.

ByronHsu:
@qingquansong can you take another look

qingquansong:
Hey @AndreSlavescu @ByronHsu  considering this is a bit different from the original feature request, maybe we can change the PR description a bit to clarify this and also correct the test cases to compare with torch fp8 matmul? (or at least torch bf16 matmul rather than fp32 before checking in) Let me know your thoughts and happy to discuss more.

src/liger_kernel/ops/experimental/gemm_split_k_fp8_e4m3.py
```diff
@@ -0,0 +1,302 @@
...
+        b = tl.load(b_ptrs, mask=mask_b, other=0.0)
+
+        # fp8 input dot product (supported types: [fp8e4nv, fp8e5, fp8e4b15])
+        acc = tl.dot(a, b, acc)
+
+        a_ptrs += block_k * split_k * stride_ak
+        b_ptrs += block_k * split_k * stride_bk
+
+    # cast to fp16 pre-store
+    acc = acc.to(tl.float16)
```
AdamLouly, line 89: Can we accumulate in fp16 directly to avoid the overhead of fp8>fp32>fp16 ?

AndreSlavescu: Yes, I can make that change and test.

AndreSlavescu: Re this:

I tried setting tl.dot output_dtype to tl.float16, but it doesn't seem to work. I don't think the tl.dot output dtype can be changed, so at least with this method, acc has to be float32.

qingquansong: I think it has to use fp32 here but probably should cast to the given compute_type inferred from input rather than fixed as fp16 (as bf16 or fp32 might also be used)

benchmark/benchmark_gemm_split_k_fp8_e4m3.py
AdamLouly, line 1: Can we put this script inside the scripts folder to keep it consistent like older benchmark scripts?

test/transformers/test_gemm_split_k_fp8_e4m3.py
```diff
@@ -0,0 +1,69 @@
...
+)
+@pytest.mark.parametrize(
+    "dtype, atol, rtol",
+    [
+        (torch.float32, 2e-1, 2e-1),
+    ],
+)
+@pytest.mark.skipif(
+    compute_capability not in [(8, 9), (9, 0)],
+    reason="FP8 GEMM is only supported on SM_89 and SM_90 CUDA devices.",
```
AdamLouly, line 40: we can change this to:
FP8 GEMM is only supported on SM_89 and newer architectures.

because we assume newer architectures will all support fp8

test/transformers/test_gemm_split_k_fp8_e4m3.py
AdamLouly, line 1: I tried to run this using 8 x H100 and the test failed because of memory constraints:

```
FAILED test_gemm.py::test_gemm_split_k[dtype0-0.2-0.2-1024-1024-1024] - triton.runtime.errors.OutOfResources: out of resource: shared memory, Required: 327680, Hardware limit: 232448. Reducing block sizes or `num_stages` may help.
FAILED test_gemm.py::test_gemm_split_k[dtype0-0.2-0.2-1024-2048-1024] - triton.runtime.errors.OutOfResources: out of resource: shared memory, Required: 327680, Hardware limit: 232448. Reducing block sizes or `num_stages` may help.
```

Test case should dynamically run only configurations that the machine's hardware can handle.

src/liger_kernel/ops/experimental/gemm_split_k_fp8_e4m3.py
```diff
@@ -0,0 +1,302 @@
...
+        k_remaining = k - k_ * (block_k * split_k)
+
+        mask_a = offs_k[None, :] < k_remaining
+        mask_b = offs_k[:, None] < k_remaining
+
+        a = tl.load(a_ptrs, mask=mask_a, other=0.0)
+        b = tl.load(b_ptrs, mask=mask_b, other=0.0)
+
+        # fp8 input dot product (supported types: [fp8e4nv, fp8e5, fp8e4b15])
+        acc = tl.dot(a, b, acc)
```
AdamLouly, line 83: you'd have to specify fp16 out_dtype for tl.dot as its fp32 by default.
This will make it work but make sure this does not impact numerical stability.

```suggestion
    acc = tl.zeros((block_m, block_n), dtype=tl.float16)

    for k_ in range(0, grid_k, step=2):
        k_remaining = k - k_ * (block_k * split_k)

        mask_a = offs_k[None, :] < k_remaining
        mask_b = offs_k[:, None] < k_remaining

        a = tl.load(a_ptrs, mask=mask_a, other=0.0)
        b = tl.load(b_ptrs, mask=mask_b, other=0.0)

        # fp8 input dot product (supported types: [fp8e4nv, fp8e5, fp8e4b15])
        acc = tl.dot(a, b, acc, out_dtype=tl.float16)
```

AndreSlavescu: I tried this previously, and I don't think its supported. They also don't list it as a param in the documentation, so my guess is that it's designed to be unmodified.

kadirnar:
Are you planning to work on training with the FP8 model?

PR #187, opened 2024-09-01
compute quantiles for memory usage

kvignesh1420:
## Summary
This PR adds quantile computations for memory usage during benchmarking.

## Details
Currently, the `_test_memory` function only return the mean memory usage and does not return quantile data. This PR adds the necessary functionality and keeps it in sync with the speed based stats that triton benchmarks compute.

Env:
```
torch==2.4.0+cu118
triton==3.0.0
```

## Testing Done


ByronHsu (approved):
Great! But i believe memory usage doesn't often oscillate

PR #201, opened 2024-09-03
Update the casting logic of RMSNorm

lancerts:
## Summary
Fix the logic of RMSNorm, and isolate the casting logic from the computation logic.
If casting model =None, it does not perform casting but still performs the computation.

### DIFF
--- a/src/liger_kernel/ops/rms_norm.py
+++ b/src/liger_kernel/ops/rms_norm.py
@@ -104,7 +104,6 @@ def _rms_norm_backward_kernel(
     dW_ptr,
     dW_row_stride,
     n_cols,
-    eps,
     offset,
     casting_mode: tl.constexpr,
     BLOCK_SIZE: tl.constexpr,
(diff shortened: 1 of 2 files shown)


src/liger_kernel/ops/rms_norm.py
```diff
@@ -104,7 +104,6 @@ def _rms_norm_backward_kernel(
     dW_ptr,
     dW_row_stride,
     n_cols,
-    eps,
```
lancerts, line 107: eps is not used in backward.

ByronHsu (commented):
add a unit test to catch the bug?

src/liger_kernel/ops/rms_norm.py
```diff
@@ -303,7 +302,6 @@ def backward(ctx, dY):
             X,
             W,
             r,
-            ctx.eps,
```
ByronHsu, line 317: does that mean we don't have to store eps in the forward pass?

lancerts:
> add a unit test to catch the bug?


will do after we bump the pytorch to 2.3-2.4. It was added recently https://github.com/pytorch/pytorch/pull/121364 so 2.1, 2.2 cannot find the baseline torch implementation.

src/liger_kernel/ops/rms_norm.py
lancerts: yes

yundai424:
How about just adding a manual one by removing `.to(torch.float32)` from https://github.com/linkedin/Liger-Kernel/blob/d338f4b9923e452baecff6d36775242a5319df4c/test/transformers/test_rms_norm.py#L25 ?

ByronHsu (approved):

PR #196, opened 2024-09-03
Refactored benchmark tests

shimizust:
## Summary
- Refactored kernel benchmarking tests to save data to a single CSV (`benchmark/data/all_benchmark_data.csv`) that contains more complete information on how the benchmarking test was setup:
```
kernel_name (e.g. swiglu)
kernel_provider (e.g. liger or huggingface)
kernel_operation_mode (e.g. full, forward, backward)
metric_name (e.g. speed, memory)
metric_unit (e.g. MB)
x_name
x_label (e.g. hidden size)
x_value
y_value_50 (median)
y_value_20 (20th percentile)
y_value_80 (80th percnetile)
extra_benchmark_config_str (e.g. {""B"": 32, ""T"": 512, ""D"": 768, ""dtype"": ""torch.float32""})
gpu_name (e.g. NVIDIA A100-SXM4-80GB)
timestamp
liger_version (e.g. 0.2.1)
```
- Removed existing image files to not clutter up the git history per previous discussion (https://github.com/linkedin/Liger-Kernel/pull/54)
- Added sample jupyter notebook that can be used to filter and plot the benchmark data
- Added makefile command to run all benchmarks: `make run-benchmarks`
- Don't overwrite existing data by default if same benchmark. Same benchmark means same kernel_name, kernel_provider, kernel_operation_mode, metric_name, x_label, gpu_name, and extra_benchmark_config_str

## Testing Done
- Ran all benchmarks and spot-checked values with existing benchmark data

image

- Hardware Type: A100
- [x] run `make test` to ensure correctness
- [x] run `make checkstyle` to ensure code style
- [x] run `make test-convergence` to ensure convergence

PR #209, opened 2024-09-04
Add `--decimal` flag to benchmarking

austin362667:
## Summary

Improves follow-up of https://github.com/linkedin/Liger-Kernel/pull/196#pullrequestreview-2279028288.
To make decimal of speed benchmark configurable.

## Testing Done

- Hardware Type: NVIDIA L40
- [X] run `make run-benchmarks OVERWRITE=1` to ensure benchmarking successfully


ByronHsu (commented):
this does not solve the issue because precision is configured on `do_bench`. triton now supports configurable precision for persisting csv: https://github.com/triton-lang/triton/blob/7480ef5028b724cb434b7841b016c6d6debf3b84/python/triton/testing.py#L295 (It was not before).  we can use that

also, we don't need to expose precision as an option because each kernel needs to use different precision. The way to check is to inspect the csv. If there are something like 0.9, 0.8, then we need to increase the prec.

tldr: use triton native argument, and manually tune the precision for the kernel if not precise enough

benchmark/data/all_benchmark_data.csv
```diff
@@ -1,411 +1,411 @@
...
-swiglu,huggingface,full,memory,MB,T,sequence length,1024,1294.25,1294.25,1294.25,"{""B"": 4, ""hidden_size"": 4096, ""dtype"": ""torch.bfloat16"", ""intermediate_size"": 11008, ""hidden_act"": ""silu""}",NVIDIA A100-SXM4-80GB,2024-09-03 15:40:36,0.2.1
-swiglu,huggingface,full,memory,MB,T,sequence length,2048,1992.25,1992.25,1992.25,"{""B"": 4, ""hidden_size"": 4096, ""dtype"": ""torch.bfloat16"", ""intermediate_size"": 11008, ""hidden_act"": ""silu""}",NVIDIA A100-SXM4-80GB,2024-09-03 15:40:36,0.2.1
-swiglu,huggingface,full,memory,MB,T,sequence length,4096,3452.25,3452.25,3452.25,"{""B"": 4, ""hidden_size"": 4096, ""dtype"": ""torch.bfloat16"", ""intermediate_size"": 11008, ""hidden_act"": ""silu""}",NVIDIA A100-SXM4-80GB,2024-09-03 15:40:36,0.2.1
-swiglu,huggingface,full,memory,MB,T,sequence length,8192,6372.25,6372.25,6372.25,"{""B"": 4, ""hidden_size"": 4096, ""dtype"": ""torch.bfloat16"", ""intermediate_size"": 11008, ""hidden_act"": ""silu""}",NVIDIA A100-SXM4-80GB,2024-09-03 15:40:36,0.2.1
+layer_norm,liger,forward,speed,ms,N,hidden size,1024,0.0573,0.0563,0.0594,"{""M"": 4096, ""dtype"": ""torch.float32"", ""eps"": 1e-06}",NVIDIA L40S,2024-09-04 16:00:49,0.2.1
```
yundai424, line 2: maybe we can keep the bench result on A100 -- or better off, maintain separate csv data for each hardware? 🤔 @shimizust @ByronHsu wdyt

ByronHsu: yes was asking austin to rerun on A100

ByronHsu:
https://github.com/linkedin/Liger-Kernel/pull/196#issuecomment-2329629032 i think no action is needed for now

austin362667:
The PR https://github.com/linkedin/Liger-Kernel/pull/196/files#diff-321a032a217a6c93e14d1cae93dd20922d020e0d55893220a026ac5c52374d81R151 already solved the under precision issue with its own csv writer, no action is needed for now, closing~

benchmark/data/all_benchmark_data.csv
shimizust: I'm not sure if the original file was deleted and re-run, but the behavior of `make run-benchmarks` or `make run-benchmarks OVERWRITE=1` in this case should have been to append entries to the existing CSV file since it was using different hardware spec. The `OVERWRITE=1` flag just means if test was run using the same parameters (including hardware), it will overwrite the existing record with the latest data.

This is essentially a file-based DB, so I don't think there's a need to split into different files. One can easily filter by GPU name when loading this data for analysis. Building up benchmarks on different GPU here could be useful.

PR #214, opened 2024-09-05
Add support for jamba model with Liger Kernel

yubofredwang:
## Summary
Add support for jamba model with Liger Kernel. The following ops can be patched with Liger kernel:
- RMSNorm
- cross_entropy
- swiglu
- lce_forward

## Testing Done


ByronHsu (dismissed):

yundai424:
CI is failing, we probably need to set `use_mamba_kernels=False` in the tests? Or install mamba-ssm in GPU CI

yubofredwang:
pip install . '[dev]' fails for this PR after [mamba-ssm](https://github.com/state-spaces/mamba) into the dependent. The reason is that mamba-ssm has a bug in its setup.py that makes it not PEP 517 compliant(basically torch, packaging, wheel has to be installed before `pip install mamba-ssm` runs, otherwise it complains no module found error). Similar situation applies to [causal-conv1d](https://github.com/Dao-AILab/causal-conv1d). 

There are fixes done on both repo, but never gets merged in or released:
- https://github.com/state-spaces/mamba/pull/402#issuecomment-2283915089
- https://github.com/Dao-AILab/causal-conv1d/pull/26

My current solution is to add it as a separate optional dependency and requires another pip install for the convergence test to run. I have a corresponding change in our CI server script here:
https://github.com/linkedin-sandbox/liger-ci-server/pull/2

Any other suggestion is highly welcomed.

pyproject.toml
```diff
@@ -27,6 +27,11 @@ dev = [
     "seaborn",
 ]
 
+test = [
```
ByronHsu, line 30: this is not good. these two dependencies seem very heavy. is there an alternative?

ByronHsu (changes requested):

pyproject.toml
yubofredwang: these won't be installed by default thought. only when you `pip install . '[test]'` they are installed. 

they are not heavy actually. installation takes about 15 seconds

helloworld1:
No activities for a long time. Closing this PR. Feel free to create a new PR if there are progress. Thanks!

PR #217, opened 2024-09-06
Update swiglu and geglu forward: zeros_like -> empty_like

IvanYashchuk:
## Summary

This PR improves the performance of swiglu and geglu forward by replacing `zeros_like` with `empty_like`. The difference is that `empty_like` doesn't require a separate kernel launch.

## Testing Done

Testing is covered by existing `test_geglu.py` and `test_swiglu.py`.

### DIFF
--- a/src/liger_kernel/ops/geglu.py
+++ b/src/liger_kernel/ops/geglu.py
@@ -98,7 +98,7 @@ def geglu_forward(a, b):
     n_cols = ori_shape[-1]
     a = a.view(-1, n_cols)
     b = b.view(-1, n_cols)
-    c = torch.zeros_like(a)
+    c = torch.empty_like(a)
     n_rows = a.shape[0]
 
     BLOCK_SIZE, num_warps = calculate_settings(n_cols)
--- a/src/liger_kernel/ops/swiglu.py
+++ b/src/liger_kernel/ops/swiglu.py
@@ -66,7 +66,7 @@ def swiglu_forward(a, b):
     n_cols = ori_shape[-1]
     a = a.view(-1, n_cols)
     b = b.view(-1, n_cols)
-    c = torch.zeros_like(a)
+    c = torch.empty_like(a)
     n_rows = a.shape[0]
 
     BLOCK_SIZE, num_warps = calculate_settings(n_cols)


ByronHsu (commented):
Thanks! Can you elaborate more what cost empty_like saves compared with zeros like?
in my understanding, there are two steps
1. cpu asks gpu to allocate a tensor
2. gpu kernel to init the values

And this PR can save the 2nd step. Is that correct? Also, is "cpu asks gpu to allocate a tensor" also a kernel?

ByronHsu (approved):

ByronHsu:
can you also run benchmark? https://github.com/linkedin/Liger-Kernel/blob/main/CONTRIBUTING.md#run-kernel-benchmarks. (Can just run individual file)

ByronHsu:
you need to rebase master to pass the CI

ByronHsu:
you need to rebase the master to pass the CI

IvanYashchuk:
> Can you elaborate more what cost empty_like saves compared with zeros like?
in my understanding, there are two steps

You're correct that there are two steps.

> cpu asks gpu to allocate a tensor

Conceptually, yes, it's the first step. Technically, the CPU asks the caching memory allocator to borrow a block of memory already allocated on the GPU. More details about PyTorch's caching allocator can be found in https://zdevito.github.io/2022/08/04/cuda-caching-allocator.html.

> gpu kernel to init the values
And this PR can save the 2nd step. Is that correct? 

That is correct. This PR saves the 2nd step. "zeros_like",  "ones_like", etc. all use a CUDA kernel to fill the tensor data with the numbers.

> Also, is "cpu asks gpu to allocate a tensor" also a kernel?

Requesting new memory from CUDA using `cudaMalloc` and `cudaFree` is a "kernel" in the sense that it blocks all other executions and waits for malloc and free to finish. However, in PyTorch, it's not blocking because PyTorch has its own allocator with rare calls to cudaMalloc and cudaFree. It's usually safe to assume that the "cpu asks gpu to allocate a tensor" step has no cost.

ByronHsu:
Curious if "cpu asks gpu to allocate a tensor" step has no cost, why we can still see perf gain?

IvanYashchuk:
Because now there's no kernel to fill the allocated tensor with zeros (zeros_like is removed), it's an unnecessary overhead.

ByronHsu:
ah got it. It was thinking the wrong thing. Thanks a ton!

S1ro1:
While going through the codebase, multiple ops use zeros/zeros_like, shouldn't we update all of those as well?

ByronHsu:
yes. we should use empty_like if the variable does not serve as an accumulator

PR #220, opened 2024-09-06
add repr infomation for layer_norm and rms_norm

wizyoung:
## Summary
Add repr information for layernorm and rmsnorm class so that the useful layer information can be displayed after the model is printed. Other classes are not modified because they inherit from related torch.nn classes, or there are torch.nn sub-modules.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/transformers/layer_norm.py
+++ b/src/liger_kernel/transformers/layer_norm.py
@@ -11,6 +11,8 @@ class LigerLayerNorm(nn.Module):
             "ones",
             "zeros",
         ], f"init_fn must be either 'ones' or 'zeros', got {init_fn}"
+        self.hidden_size = hidden_size
+        self.eps = eps
         self.weight = nn.Parameter(
             torch.ones(hidden_size) if init_fn == "ones" else torch.zeros(hidden_size)
         )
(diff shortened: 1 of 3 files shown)


src/liger_kernel/transformers/layer_norm.py
```diff
@@ -23,3 +25,6 @@ def forward(self, hidden_states):
         return LigerLayerNormFunction.apply(
             hidden_states, self.weight, self.bias, self.variance_epsilon
         )
+        
+    def extra_repr(self) -> str:
+        return f'{self.hidden_size}, eps={self.eps}'
```
ByronHsu, line 30: why no weight here?

wizyoung: This is in line with nn.LayerNorm whose extra_repr is:
```python
def extra_repr(self) -> str:
        return '{normalized_shape}, eps={eps}, ' \
            'elementwise_affine={elementwise_affine}'.format(**self.__dict__)
```
For rms_norm, it is in line with huggingface transformers.

wizyoung: Do we need to keep them consistent here?

ByronHsu: oh sg

ByronHsu (approved):

ByronHsu:
need to fix the checkstyle

wizyoung:
now checkstyle passed

ByronHsu (approved):

Issue #176, opened 2024-08-30
[testing bug] Many convergence tests are incorrect - comparing LigerKernel against itself


tyler-romero:
### 🐛 Describe the bug

We typically test each model twice, once with float32 and once afterwards with bfloat16. The second test for a given model never reverts the monkeypatch of the first, so they are comparing LigerKernel against itself.

A fix would look like a reverse monkeypatch.

### Versions

python -m liger_kernel.env_report                                                           
Environment Report:
-------------------
Operating System: Linux-6.5.0-44-generic-x86_64-with-glibc2.35
Python version: 3.10.13
PyTorch version: 2.4.0+cu121
CUDA version: 12.1
Triton version: 3.0.0
Transformers version: 4.45.0.dev0

ByronHsu:
critical bug catch!!

austin362667:
#take

austin362667:
Great catch on the bug! Thank you so much. I’ve drafted [PR #232](https://github.com/linkedin/Liger-Kernel/pull/232) to address this issue. However, if it resolves the problem, it may potentially introduce other value mismatch issues. We can fix them in follow-up PRs.

PR #238, opened 2024-09-09
fused_linear_cross_entropy: Move float32 cast into kernel

hansonw:
## Summary

Another small optimization :) The `logits_chunk.float()` allocation may be surprisingly large, e.g. Cohere models have 256K vocabs, so each logit chunk in float32 could be something like 1024 * 256K * 4 = 1GB VRAM (even more if the chunk size is larger.)

I actually don't think any explicit casting is even required within the Triton kernel since the intermediate softmax calculation variables like `m`, `d`, etc. are already float32 by default, so [with type promotion](https://triton-lang.org/main/python-api/triton-semantics.html) the calculations should all be float32 regardless.

However, I added explicit casts `.cast(tl.float32)` around all of the X_ptr loads to make this more obvious to the reader. In either case, the actual `liger_cross_entropy_kernel` runs so quickly that I don't think there's any performance difference - this is purely to save the float32 allocation. (It might be more efficient without the explicit casts, but I was not able to measure anything - even with a 1K x 256K logit matrix the kernel kind of runs instantly lol.)

## Testing Done

- Hardware Type: A100 80GB

### DIFF
--- a/src/liger_kernel/ops/cross_entropy.py
+++ b/src/liger_kernel/ops/cross_entropy.py
@@ -61,8 +61,8 @@ def liger_cross_entropy_kernel(
     # 3. [Online softmax] first pass: find max + sum
     m = float("-inf")  # m is the max value. use the notation from the paper
     d = 0.0  # d is the sum. use the notation from the paper
-    ori_X_y = tl.load(
-        X_ptr + y
+    ori_X_y = tl.load(X_ptr + y).cast(
+        tl.float32
     )  # we need to store the original value of X_y for the loss calculation
 
     # Label smoothing is a general case of normal cross entropy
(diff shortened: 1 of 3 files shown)


ByronHsu:
thanks! this makes sense. i did try the similar thing before but seen divergence compared with casting from the torch side (not sure why, maybe i did it wrong). also, currently bfloat16 convergence test is not actually tested due to https://github.com/linkedin/Liger-Kernel/issues/176. we can try to run on convergence tests with bf16 to see if there is any gaps.

hansonw:
I added a test with `test_float32_internal` which runs the kernel twice, once with a bfloat16 input and once with a float-upcasted version and verifies that the resulting output (in bfloat16) is exactly identical 🙂 You can also verify that the test passes even without the explicit `.cast(tl.float32)` calls, so maybe those could be removed as long as the test is present..

ByronHsu:
cool! I will take a deeper look today or tomorrow. This is exciting!

kostum123:
Can we merge this? in current form liger kernel broken.

lancerts:
@hansonw  can we resolve the conflict? ty

ByronHsu:
We can merge this once the conflict is resolved. thanks!!

winglian:
@ByronHsu I rebased this PR and fixed the conflicts, but not sure of the best way to help get the changes into this PR. see https://github.com/winglian/Liger-Kernel/pull/new/hw-fused-ce-float

ByronHsu:
@pramodith would you interested in moving this forward? maybe create a new PR based on the current main. We just have to ensure the precision is correct

pramodith:
> @pramodith would you interested in moving this forward? maybe create a new PR based on the current main. We just have to ensure the precision is correct

@ByronHsu done #406

PR #255, opened 2024-09-19
RMSNorm aggregation

### LINKED ISSUE #179: Improve the efficiency of the RMSNorm aggregation

Move this line https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/ops/rms_norm.py#L306, the sum in pytorch to partial aggregation in triton, reference 
https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/ops/layer_norm.py#L106,
which does 2 level of aggregation, first in triton and second in torch (more efficient).

### Alternatives

_No response_

### Additional context

_No response_

Tcc0403:
## Summary
Resolve #179 

WIP: solving issues with numerical stability for large hidden_size (4096)
## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/ops/rms_norm.py
+++ b/src/liger_kernel/ops/rms_norm.py
@@ -10,6 +10,7 @@ https://github.com/unslothai/unsloth/blob/fd753fed99ed5f10ef8a9b7139588d9de9ddec
 Modifications made by Yanning Chen, 2024.
 """
 
+import math
 import operator
 
 import torch
@@ -20,6 +21,7 @@ from liger_kernel.ops.utils import (
     calculate_settings,
     compare_version,
     ensure_contiguous,
+    torch_to_triton_dtype,
 )
 
 if compare_version("triton", operator.ge, "3.0.0"):
@@ -100,6 +102,9 @@ def _rms_norm_forward_kernel(
 
     Y_row = X_row * (offset + W_row)
 
+    if casting_mode == _CASTING_MODE_GEMMA:
+        Y_row = Y_row.to(X_row_dtype)
+
     tl.store(Y_ptr + col_offsets, Y_row, mask=mask)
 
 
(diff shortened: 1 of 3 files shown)


Tcc0403:
It can pass with (atol, rtol) = (1e-4, 1e-5), but I'm not sure if the change is acceptable. @ByronHsu

Tcc0403:
The reason why it didn't pass the test was mainly because of numerical instability with large BxT, not hidden_size. 
Ready for review now.

src/liger_kernel/ops/rms_norm.py
```diff
@@ -195,7 +213,7 @@ def rms_norm_forward(X, W, eps, offset, casting_mode):
 
     shape = X.shape
     dim = shape[-1]
-    X = X.view(-1, dim)
+    X = X.reshape(-1, dim)
```
ByronHsu, line 216: why view -> reshape. same question for other parts

src/liger_kernel/ops/rms_norm.py
```diff
@@ -236,33 +254,40 @@ def rms_norm_forward(X, W, eps, offset, casting_mode):
 def rms_norm_backward(dY, X, W, RSTD, offset, casting_mode, BLOCK_SIZE, num_warps):
     shape = dY.shape
     dim = shape[-1]
-    dY = dY.view(-1, dim)
+    dY = dY.view(-1, dim).contiguous()
```
ByronHsu, line 257: in most of the case this is contiguous right?

ByronHsu (commented):
@lancerts can you take another look too? thanks

src/liger_kernel/ops/rms_norm.py
Tcc0403: Oh my bad, I simply forgot to remove it after testing around.

src/liger_kernel/ops/rms_norm.py
Tcc0403: All these reshape can swap back to view. The change was initially made because I thought my very first implementation was wrong due to not contiguous, and [reshape](https://pytorch.org/docs/stable/generated/torch.reshape.html#torch.reshape) can guarantee tensor contiguous.

All tensors are already contiguous with decorator `@ensure_contiguous`, so view should be fine.

lancerts (approved):

Issue #250, opened 2024-09-16
AutoLigerKernelForCausalLM.from_pretrained discards hub_kwargs_names


uris-opti:
### 🐛 Describe the bug

AutoLigerKernelForCausalLM.from_pretrained retain only the keyword args present in the model configuration, which do not include hub_kwargs_names -> [
            "cache_dir",
            "force_download",
            "local_files_only",
            "proxies",
            "resume_download",
            "revision",
            "subfolder",
            "use_auth_token",
            "token",
        ]
therefore cannot replace AutoModelForCausalLM

### Reproduce

N/A

### Versions

N/A

tyler-romero:
+1, I notice that some models don't have `attn_implementation` in their config even though its a valid keyword arg to `AutoModelForCausalLM` - so the user-specified `attn_implementation` gets discarded as well.

shimizust:
Thanks for reporting! Currently we are only keeping kwargs that are present in the model config. I wasn't aware of this other set of valid args--let me look into it

uris-opti:
Thanks for looking into it!
I don't know what's the motivation for filtering the kwargs, but I would consider removing this logic completely,
After seeing Tyler's comment, I looked around the transformers code and noticed there are many many kwargs that are not in the config and they vary between from_pretrained implementations, 
I also saw that each from_pretrained implementation handles extra kwargs, so just passing on the kwargs seems safe
🙏

shimizust:
Hm, when I tried not removing extra kwargs (e.g. those that get passed to the liger monkey patch function) it resulted in errors, hence the filtering. Looking at this now

shimizust:
If you can take a look @yundai424 @JasonZhu1313

PR #263, opened 2024-09-20
Fix AutoLigerKernelForCausalLM to pass through original kwargs

shimizust:
## Summary
- Fixes https://github.com/linkedin/Liger-Kernel/issues/250 to correctly pass all original kwargs to .from_pretrained(). Previously we were only passing args that were part of the model config, but there are additional valid kwargs beyond that.
- We still need to filter out the kwargs passed into the apply_liger_* functions, or else will result in model init errors

## Testing Done
Tested on huggingface example with some of the args in https://github.com/linkedin/Liger-Kernel/issues/250

- Hardware Type: A100

### DIFF
--- a/src/liger_kernel/transformers/auto_model.py
+++ b/src/liger_kernel/transformers/auto_model.py
@@ -40,7 +40,6 @@ class AutoLigerKernelForCausalLM(AutoModelForCausalLM):
             if key not in apply_fn_signature.parameters
         }
 
-        print(applicable_kwargs)
         return super().from_pretrained(
             pretrained_model_name_or_path, *model_args, **applicable_kwargs
         )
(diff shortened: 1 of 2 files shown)


qingquansong (approved):
LGTM!

uris-opti:
Thanks for fixing this!
It solves the issue for me 🙏 

Just one comment,
there's a chance in the future there'll be intersecting kwargs between the apply_liger_* function and the super.from_pretrained function, in this case you'll filter out kwargs that might've been intended to the super.from_pretrained function

if you want to fix that, 
one solution that comes in mind is to rename the apply_liger_* kwargs so it'll be unlikely to have these intersecting with the from_pretrained function,
something like:
```
liger_rope: bool = True,
liger_cross_entropy: bool = False,
...
```

PR #269, opened 2024-09-25
Fix sharing a ResBlock layer for each head in Medusa example

chiwanpark:
## Summary

There is a bug of incorrect weight sharing between layers for each Medusa head. Since `nn.Module` is Python reference, the original source code creates a list containing references to the same weights. This PR fixes the bug.

## Testing Done


lancerts:
Great catch!!!

lancerts (approved):

JasonZhu1313:
@ctlllll

PR #290, opened 2024-10-01
Add beta support for jsd 

### LINKED ISSUE #278: Add general JSD (w/ beta) support

We have had a great JSD kernel implemented in this [PR](https://github.com/linkedin/Liger-Kernel/pull/264) and this request is for helping extend with an optional beta arg to create the mixed distribution between student and teacher. When beta=0, it should be equivalent to forward KL and beta=1 it should be equivalent to reverse KL. Note: the corner case could be tricky. More details can be found in paper: https://arxiv.org/abs/2306.13649

### Alternatives

_No response_

### Additional context

_No response_

Tcc0403:
## Summary
Resolve #278 .

### Forward:
```math
\begin{align}
JSD(X, Y, \beta) &= JSD_{\beta}(P \Vert Q)\\
&= \beta\ KL(P \Vert \beta P + (1-\beta)Q) + (1-\beta)\ KL(Q \Vert \beta P + (1-\beta)Q)\\
&= \sum \beta\ PY + (1-\beta)QX - M\ logM
\end{align}
```
where $X=logQ$, $Y=logP$ and $M=\beta P + (1-\beta)Q$. 

### Gradients:
```math
\frac{\partial}{\partial X_i} JSD(X, Y, \beta) = (1-\beta)Q_i(X_i - logM_i)
```
## Testing Done
jsd_memory
jsd_speed

- Hardware Type: H100


src/liger_kernel/transformers/jsd.py
```diff
@@ -4,8 +4,12 @@
...
+    def __init__(self, beta=0.5):
         super().__init__()
+        assert (
+            beta > 0 and beta < 1
+        ), f"beta must be greater than 0 and less than 1. Got: {beta}"
+        self.beta = beta
 
-    def forward(self, p, q):
-        return LigerJSDFunction.apply(p, q)
+    def forward(self, log_q, log_p):
```
Tcc0403, line 14: This is the correct order of input and target (student and teacher) respectively. would it be too confusing?

test/transformers/test_jsd.py
```diff
@@ -1,25 +1,34 @@
...
+        log_q: torch.tensor,  # input
+        log_p: torch.tensor,  # target
+    ):
+        log_p, log_q = log_p.to(torch.float), log_q.to(torch.float)
         log_p, log_q = log_p.view(-1, log_p.size(-1)), log_q.view(-1, log_q.size(-1))
-        m = 0.5 * (torch.exp(log_p) + torch.exp(log_q))
-        log_m = torch.log(m)
-        loss = 0.5 * (self.kl(log_m, log_p) + self.kl(log_m, log_q))
-        return loss
+        m = self.beta * torch.exp(log_p) + (1 - self.beta) * torch.exp(log_q)
```
qingquansong, line 27: nit: can just use https://pytorch.org/docs/stable/generated/torch.lerp.html

src/liger_kernel/transformers/jsd.py
qingquansong: yeah, the name is a bit confusing, or we can add some descriptions here to clarify

qingquansong (approved):
LGTM in general! In case you're interested, I think one good future work is to make those KL or JSD losses similar to the fuse CE loss: feed teacher and student model last projection layer to the kernel and fuse it with the losses. Here teacher weight does not need grad and student will need grad.

ByronHsu:
awesome work! waiting for the final nit review

lancerts (dismissed):

Tcc0403:
> I think one good future work is to make those KL or JSD losses similar to the fuse CE loss: feed teacher and student model last projection layer to the kernel and fuse it with the losses. Here teacher weight does not need grad and student will need grad.

@qingquansong sure, I'm in.

Tcc0403:
Forgot to add jsd in readme and liger_kernel.transformer

lancerts (approved):

PR #300, opened 2024-10-08
Add FusedLinearJSD

Tcc0403:
## Summary
similar to the fuse linear CE.

It handles the forward and backward pass of the final linear layer via JSD by avoiding the materialization of the large logits tensor. Since JSD is the last layer, we can compute the gradient at the forward pass.

## Testing Done
Hidden size: 4096, Vocab size: 128256
fused_linear_jsd_memory
fused_linear_jsd_speed

- Hardware Type: NVIDIA H100 80GB HBM3 (SXM5)


src/liger_kernel/ops/fused_linear_jsd.py
```diff
@@ -0,0 +1,201 @@
...
+
+        if grad_weight is not None:
+            torch.addmm(
+                input=grad_weight,
+                mat1=student_logits_chunk.t(),  # gradients of logits_chunk
+                mat2=student_input_chunk,
+                out=grad_weight,
+            )
+
+    loss = torch.sum(loss_1d) / BT
```
Tcc0403, line 109: I just noticed that torch.sum might overflow when BT size is large. It can be fixed by either
1. do division first before sum, i.e. torch.sum(loss_1d / BT) 
2. put division inside JSD kernel
I prefre the second solution more, but I'm not sure if it is ok to modify another kernel in this PR.

qingquansong: can the [existing division in JSD](https://github.com/linkedin/Liger-Kernel/blob/de12602d858a6e83aaacc56e5cb64ab218c75a0a/src/liger_kernel/ops/jsd.py#L77) be able to handle this? 🤔 I saw you change the [`n_row` for each chunk ](https://github.com/linkedin/Liger-Kernel/pull/300/files#diff-1da1c5423f58c2cb144e1e7d131f17bc0eaf4915d34ced8177792dfee9f6aa46R85) to be BT now.

src/liger_kernel/ops/utils.py
```diff
@@ -68,3 +68,39 @@ def compare_version(package: str, operator: Callable, target: str):
     torch.float16: tl.float16,
     torch.bfloat16: tl.bfloat16,
 }
+
+
+@triton.jit
+def element_mul_kernel(
```
qingquansong, line 74: maybe we can delete the one in the original [ce](https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/ops/cross_entropy.py#L163) and share this

qingquansong (approved):
LGTM! We can add the ignore index later similar to [here](https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/ops/cross_entropy.py#L16C5-L16C17) so it can be easily used for the SFT context. Great work!

src/liger_kernel/ops/fused_linear_jsd.py
Tcc0403: curretnly n_row parameter in JSD is only for gradients calculation.
it can be used to loss calculation wrt demanding reduction method like cross_entropy
https://github.com/linkedin/Liger-Kernel/blob/de12602d858a6e83aaacc56e5cb64ab218c75a0a/src/liger_kernel/ops/cross_entropy.py#L143



> I saw you change the [`n_row` for each chunk ](https://github.com/linkedin/Liger-Kernel/pull/300/files#diff-1da1c5423f58c2cb144e1e7d131f17bc0eaf4915d34ced8177792dfee9f6aa46R85) to be BT now.

simply passing BT can perform correct calculations without further alpha tweaking like flce does
https://github.com/linkedin/Liger-Kernel/blob/de12602d858a6e83aaacc56e5cb64ab218c75a0a/src/liger_kernel/ops/fused_linear_cross_entropy.py#L112

Tcc0403: > can the [existing division in JSD](https://github.com/linkedin/Liger-Kernel/blob/de12602d858a6e83aaacc56e5cb64ab218c75a0a/src/liger_kernel/ops/jsd.py#L77) be able to handle this? 

it has potential overflow issue too, thats one of the reason why i think moving division into kernel is better

qingquansong: Gotcha. I'm good with moving it inside the JSD kernel. For ce/flce, let's keep them for now. Thanks!

qingquansong (approved):

PR #302, opened 2024-10-11
Monkey patch layer norm in mllama

shivam15s:
## Summary

Monkey patches layer norm in mllama for conditional generation

## Testing Done

Tested monkey patching works as intended

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/transformers/monkey_patch.py
+++ b/src/liger_kernel/transformers/monkey_patch.py
@@ -152,6 +152,7 @@ def apply_liger_kernel_to_mllama(
         MllamaForCausalLM,
         MllamaForConditionalGeneration,
         MllamaTextModel,
+        MllamaVisionModel,
     )
 
     from liger_kernel.transformers.model.mllama import lce_forward as mllama_lce_forward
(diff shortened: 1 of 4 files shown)


src/liger_kernel/transformers/monkey_patch.py
```diff
@@ -151,12 +152,15 @@ def apply_liger_kernel_to_mllama(
...
         MllamaTextModel,
+        MllamaVisionModel,
     )
 
     from liger_kernel.transformers.model.mllama import lce_forward as mllama_lce_forward
 
     if rope:
         modeling_mllama.apply_rotary_pos_emb = liger_rotary_pos_emb
+    if layer_norm:
+        modeling_mllama.nn.LayerNorm = LigerLayerNorm
```
tyler-romero, line 163: FWIW I looked into this but was worried this would modify torch.nn.LayerNorm globally, not just for the targeted model

shivam15s: That probably answers why ci tests fail. LigerLayerNorm gets applied even before we monkey patch because a previous test patched layer norm globally.

qingquansong (approved):

ByronHsu (approved):

PR #306, opened 2024-10-12
Add ignore_index and label to jsd and fl-jsd

### LINKED ISSUE #277: Adding ignore index support for divergence losses

We've had implemented KL divergence and JSD loss. Thanks to the community!
This feature request is to: add an optional feature for ignoring index (need to and an extra index tensor input + an arg for ignore index) during the loss compute

### Alternatives

_No response_

### Additional context

_No response_

Tcc0403:
## Summary
Resolve #277.

## Testing Done

- Hardware Type: gpu-ci

### DIFF
--- a/src/liger_kernel/ops/jsd.py
+++ b/src/liger_kernel/ops/jsd.py
@@ -67,7 +67,7 @@ def _jsd_kernel(
 MAX_FUSED_SIZE = 65536
 
 
-def jsd_forward(_input, target, label, beta, ignore_index, has_label):
+def jsd_forward(_input, target, shift_labels, beta, ignore_index, has_label):
     BT, V = _input.shape
     n_rows = BT
     BLOCK_SIZE = min(MAX_FUSED_SIZE, triton.next_power_of_2(V))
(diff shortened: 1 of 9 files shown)


src/liger_kernel/ops/jsd.py
```diff
@@ -43,24 +57,29 @@ def _jsd_kernel(
...
-        dX = (1 - beta) * Q * (X - log_M) / n_rows
+        dX = (1 - beta) * Q * (X - log_M) / n_non_ignore
         tl.store(dX_ptr + offsets, dX, mask=mask)
 
 
 MAX_FUSED_SIZE = 65536
 
 
-def jsd_forward(_input, target, beta):
+def jsd_forward(_input, target, label, beta, ignore_index, has_label):
```
yundai424, line 70: I might be wrong -- if i'm understanding it correctly, we currently have an intrinsic assumption that the label is shifted already. It would be helpful to specify this requirement and provide an example of what kind of input we'll expect in this case 🤔

Tcc0403: Makes sense. I added some examples in transformers files.

src/liger_kernel/ops/jsd.py
```diff
@@ -15,10 +17,13 @@ def _jsd_kernel(
     loss_stride,
     dX_ptr,
     dX_stride,
+    label_ptr,
     beta,
-    n_rows,
+    n_non_ignore,
+    ignore_index,
```
yundai424, line 23: nit: this could be a constexpr

yundai424 (changes requested):
overall LGTM just a very minor suggestion!

yundai424 (approved):

Issue #227, opened 2024-09-07
Torch compiled FLCE is 2x faster than the current FLCE


ByronHsu:
### 🚀 The feature, motivation and pitch

We can leverage torch compile to fuse the things we cannot fuse now like upcasting, contiguous call, etc

![image](https://github.com/user-attachments/assets/94b377d3-6530-43f5-9f88-1c7108a4746f)

Sample code: https://gist.github.com/Chillee/22cd93e11b887db1f596ab754d60a899#file-lce_benchmark-py
Provided by the brilliant @Chillee

wizyoung:
Actually, if we align the CHUNK_SIZE of the Torch-compiled FLCE with the strategy used in Liger's FLCE, the compiled version is only slightly faster than the Liger version, but it does require a bit more memory as well. The advantage of the Torch-compiled version is its flexibility; implementing the Gemma2 softcap logits is very straightforward, whereas I struggled for some time to achieve consistent accuracy with this in Liger.

wizyoung:
By setting CHUNK_SIZE and run benchmark script:
![image](https://github.com/user-attachments/assets/2de14072-7a48-4439-92d9-d460c0af2f0b)
![image](https://github.com/user-attachments/assets/8bc4d1c3-0fb6-489b-a4b6-4706f8fb2e76)

Chillee:
@wizyoung how are you setting the chunk size? I wasn't able to get the liger kernel to perform much better even when changing the chunk size.

wizyoung:
@Chillee By referencing https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/ops/fused_linear_cross_entropy.py#L23. I mean, chaning chunk size in torch compiled FLCE. Your default chunk size is 1024, and I change to 256. Then I have:
![image](https://github.com/user-attachments/assets/06e43de9-bd5f-4d08-a00e-cef45fa1f80f)
By only keep benchmark test with liger and compiled chunkce:
![image](https://github.com/user-attachments/assets/660c9584-c1b7-4eb8-9e8b-3ed517b8a20b)
![image](https://github.com/user-attachments/assets/ce8ef2b7-442c-47c3-9c17-5e2e203939f0)

wizyoung:
I have done some quick tests with different B, T, D and V to mimic my training conditions(llama3 and gemma2) in my env, my conclusion is that torch compiled flce is indeed faster, but has worse memory management.

wizyoung:
[...]

PR #362, opened 2024-11-08
Add Chunked ORPO Loss

shivam15s:
## Summary

Adds chunked ORPO loss kernel 

## Testing Done

Benchmarks
Speed ORPO
Mem ORPO

References:
1. #227 
2. https://gist.github.com/Chillee/22cd93e11b887db1f596ab754d60a899#file-lce_benchmark-py

- Hardware Type: <BLANK>


src/liger_kernel/chunked_loss/orpo_loss.py
```diff
@@ -0,0 +1,180 @@
...
+    )
+    ratio = F.logsigmoid(log_odds)
+    return beta * ratio.sum()
+
+
+class LigerFusedLinearORPOFunction(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx, _input, weight, target, bias=None, ignore_index=-100, beta=0.1, compiled=True):
+        """
+        Fused linear forward function with ORPO (Odds-Ratio Preference Optimization).
```
pramodith, line 31: Can you please add a reference link to the paper or a blog explaining the function.

src/liger_kernel/chunked_loss/orpo_loss.py
```diff
@@ -0,0 +1,180 @@
...
+        Args:
+            _input (torch.Tensor): Input tensor. Shape: (batch_size, seq_len, hidden_size).
+            weight (torch.Tensor): Weight tensor. Shape: (vocab_size, hidden_size).
+            bias (torch.Tensor, optional): Bias tensor. Shape: (hidden_size,).
+            ignore_index (int): Index to ignore for loss computation.
+            compiled (bool): Whether to use compiled mode for chunk accumulation.
+        """
+        CHUNK_SIZE = 1
```
pramodith, line 39: Should we add a TODO to tune this param? We might be under-utilizing GPU memory if we number of chunks is always set to batch_size/2.

src/liger_kernel/chunked_loss/orpo_loss.py
```diff
@@ -0,0 +1,180 @@
...
+
+class LigerFusedLinearORPOFunction(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx, _input, weight, target, bias=None, ignore_index=-100, beta=0.1, compiled=True):
+        """
+        Fused linear forward function with ORPO (Odds-Ratio Preference Optimization).
+        Args:
+            _input (torch.Tensor): Input tensor. Shape: (batch_size, seq_len, hidden_size).
+            weight (torch.Tensor): Weight tensor. Shape: (vocab_size, hidden_size).
+            bias (torch.Tensor, optional): Bias tensor. Shape: (hidden_size,).
```
pramodith, line 35: Shouldn't shape of the bias term be = `vocab_size` ?

src/liger_kernel/chunked_loss/orpo_loss.py
```diff
@@ -0,0 +1,180 @@
...
+            unnorm_logits = input_chunk @ weight.t()  # chunk_size x V
+            if bias is not None:
+                unnorm_logits = unnorm_logits + bias
+            unnorm_logits = unnorm_logits.float()
+            norm_logits = F.log_softmax(unnorm_logits, dim=-1)
+
+            chosen_nll_loss = F.nll_loss(
+                norm_logits[:len_chosen_chunk].view(-1, norm_logits.shape[-1]),
+                target_chunk[:len_chosen_chunk].view(-1),
+                reduction="sum",
```
pramodith, line 53: wondering if this should be a customizable choice between `mean` and `sum`

src/liger_kernel/chunked_loss/orpo_loss.py
```diff
@@ -0,0 +1,180 @@
...
+                ignore_index=ignore_index,
+            )
+            chosen_nll_loss = chosen_nll_loss / (target[:target.shape[0]//2] != ignore_index).sum()
```
pramodith, line 56: If my understanding is correct, this is computing the mean nll loss for the chosen/accepted response, why can't we just use 'mean' as the reduction method above rather than computing the sum first and then dividing by the number of non-pad tokens?

src/liger_kernel/chunked_loss/orpo_loss.py
pramodith: We're also missing a doc string for `target`

src/liger_kernel/chunked_loss/orpo_loss.py
```diff
@@ -0,0 +1,180 @@
...
+        grad_chosen_inputs = []
+        grad_rejected_inputs = []
+        grad_bias = torch.zeros_like(bias) if bias is not None else None
+        loss_acc = torch.zeros((), device=_input.device)
+
+        chunks = max(1, _input.shape[0] // (2 * CHUNK_SIZE))
+
+        def accumulate_chunk(input_chunk, target_chunk):
+            if bias is not None:
+                (chunk_grad_input, chunk_grad_weight, chunk_grad_bias), chunk_loss = torch.func.grad_and_value(
```
pramodith, line 86: TIL about `torch.func.grad_and_value`! Cool stuff!

test/chunked_loss/test_orpo_loss.py
```diff
@@ -0,0 +1,191 @@
...
+        labels = torch.where(labels == self.ignore_index, 0, labels)
+
+        per_token_logps = torch.gather(logits.log_softmax(-1), dim=2, index=labels.unsqueeze(2)).squeeze(2)
+
+        if average_log_prob:
+            return (per_token_logps * loss_mask).sum(-1) / loss_mask.sum(-1)
+        else:
+            return (per_token_logps * loss_mask).sum(-1)
+
+    def odds_ratio_loss(
```
pramodith, line 46: We should probably add a link to the original source of the code: https://github.com/huggingface/trl/blob/v0.11.3/trl/trainer/orpo_trainer.py#L75

test/chunked_loss/test_orpo_loss.py
```diff
@@ -0,0 +1,191 @@
...
+    [
+        (1.0, torch.bfloat16, 5e-2, 5e-1),
+        (1.0, torch.float32, 1e-5, 5e-4),
+    ],
+)
+@pytest.mark.parametrize("bias", [True, False])
+@pytest.mark.parametrize("ignore_index, beta", [(-100, 0.1), (42, 0.2)])
+def test_correctness(B, T, H, V, scalar, dtype, atol, rtol, bias, ignore_index, beta):
+    # Define input tensors
+    _tensor = torch.randn(B, T, H, device="cuda", dtype=dtype) * scalar
```
pramodith, line 163: What's the purpose of scalar?

pramodith (changes requested):
Great stuff! Most of my comments are around documentation, wondering if we should also include a benchmark script for this one!

src/liger_kernel/chunked_loss/orpo_loss.py
shivam15s: The target variable here is the whole target while target_chunk is the target for a given chunk. Normalizing directly using nll_loss mean reduction would normalize with target_chunk length which is mathematically wrong. So this is kinda hacky way to do normalization

test/chunked_loss/test_orpo_loss.py
shivam15s: I believe the intention is to support testing inputs with variances other than one. However, the current tests do not yet include cases with non-unit variances.

lancerts:
> Great stuff! Most of my comments are around documentation, wondering if we should also include a benchmark script for this one!

@shivam15s great work! @pramodith thanks for the detailed review. 
And yes, we should add the bench script

src/liger_kernel/chunked_loss/orpo_loss.py
pramodith: Ahh gotcha, that makes sense, thanks for the explanation.

src/liger_kernel/chunked_loss/orpo_loss.py
```diff
@@ -0,0 +1,211 @@
...
+
+            per_token_logps = norm_logits.gather(-1, label_chunk.unsqueeze(-1)).squeeze(
+                -1
+            )
+            average_log_prob = (per_token_logps * loss_mask).sum(-1) / loss_mask.sum(-1)
+
+            chosen_logps = average_log_prob[:len_chosen_chunk]
+            rejected_logps = average_log_prob[len_chosen_chunk:]
+
+            or_loss = odds_ratio_loss(chosen_logps, rejected_logps, beta=beta)
```
Xiaoling1992, line 86: could we rename this to odds ratio loss?

The returned is orpo loss = sft loss + odds ratio loss, not this one.

src/liger_kernel/chunked_loss/orpo_loss.py
```diff
@@ -0,0 +1,211 @@
...
+            or_loss = or_loss / (target.shape[0] // 2)
+
+            loss = chosen_nll_loss - or_loss
```
Xiaoling1992, line 89: Could this class take a flag like enable_sft_loss? 

Since users might have their customized sft loss logic and they needs odds ratio loss only. When the flag is True, this class computes chosen_nll_loss and returns chosen_nll_loss - or_loss, otherwise, it computes and returns odds_ratio_loss only.

src/liger_kernel/chunked_loss/orpo_loss.py
shivam15s: To clarify, or_loss here is the odds ratio loss; while the loss returned from the fn is the sum of odds ratio loss (or_loss) and sft loss (chosen_nll_loss). Do you want to rename or_loss to something more verbose?

src/liger_kernel/chunked_loss/orpo_loss.py
lancerts: +1, or we can add an input parameter `chosen_nll_loss_weight` and `loss = chosen_nll_loss_weight* chosen_nll_loss - or_loss`

Xiaoling1992: There is already a beta to balance the weights between sft loss and odds_ratio loss.
loss = sft loss + beta * odds ratio loss

src/liger_kernel/chunked_loss/orpo_loss.py
Xiaoling1992: Thanks for clarification. Could we rename or_loss to odds_ratio_loss since or_loss might be misunderstood to be orpo_loss.

src/liger_kernel/chunked_loss/fused_linear_preference.py
```diff
@@ -0,0 +1,107 @@
+import torch
+
+
+class LigerFusedLinearPreferenceBase(torch.autograd.Function):
+    @staticmethod
+    def forward(
```
ByronHsu, line 6: general suggestion:

1. can we make loss_fn a member method in the base class and raise not implemented exception and the derive class needs to implement it? instead of being a separate python function

2. can we use `__super__` instead of `LigerFusedLinearPreferenceBase.forward` (backward). Or we can not do that for static method.


I will merge this PR first and we can address later! Great contribution!!

ByronHsu (approved):

Issue #369, opened 2024-11-08
LigerCrossEntropyLoss is not patched for latest transformers models


Tcc0403:
### 🐛 Describe the bug

To fix GA bug, latest transformers no longer imports CrossEntropyLoss. Instead, it's wrapped in [`self.loss_function`](https://github.com/huggingface/transformers/blob/a06a0d12636756352494b99b5b264ac9955bc735/src/transformers/models/llama/modeling_llama.py#L1224) which can be traced back to [here](https://github.com/huggingface/transformers/blob/a06a0d12636756352494b99b5b264ac9955bc735/src/transformers/loss/loss_utils.py#L26). So that [current patching method](https://github.com/linkedin/Liger-Kernel/blob/2d3eb94679cec946c8f85a15590cbef9c1327250/src/liger_kernel/transformers/monkey_patch.py#L115) doesn't work on latest transformers model.

Note that [labels and logits shifting operations are wrapped](https://github.com/huggingface/transformers/blob/a06a0d12636756352494b99b5b264ac9955bc735/src/transformers/loss/loss_utils.py#L32) in `loss_function` as well, we need to take care of it when implementing new patching method.

### Versions

none

ByronHsu:
fixed by https://github.com/linkedin/Liger-Kernel/pull/375

Issue #368, opened 2024-11-08
Keyword arguments for liger functional


Tcc0403:
### 🚀 The feature, motivation and pitch

Liger's functional doesn't support keyword arguments because it's implemented by replacing `torchautograd.Function.apply`.

kwargs support is necessary for LigerCrossEntropyLoss to patch transformers model after GA fix 
https://github.com/huggingface/transformers/blob/a06a0d12636756352494b99b5b264ac9955bc735/src/transformers/models/llama/modeling_llama.py#L1224

ByronHsu:
CE is fixed by https://github.com/linkedin/Liger-Kernel/pull/375. But need to add functional wrapper for other modules

hongpeng-guo:
I can take this one 😄

PR #374, opened 2024-11-12
Fix cross entropy patch for LLama

pramodith:
## Summary
This PR addresses issues #368 and partially addresses #369 by fixing the patch for Llama based models.

## Details
Some recent changes to HF's transformers lib meant that the training scripts for llama models no longer defined a `cross_entropy` attribute, rather they included a `loss_function` [attribute](https://github.com/huggingface/transformers/blob/a06a0d12636756352494b99b5b264ac9955bc735/src/transformers/models/llama/modeling_llama.py#L1224) that effectively mapped to torch's `nn.cross_entropy` [function](https://github.com/huggingface/transformers/blob/main/src/transformers/loss/loss_utils.py#L32).

These changes invalidated the way we patched up the models with LigerCrossEntropy.

In this PR we correct the patch in LLama by:

- Allowing `LigerCrossEntropyLoss` to accept kwargs in the constructor.
- Update the patched forward pass to also consider the implications of using `LigerCrossEntropyLoss`

_Note: Additional PR's might be needed to patch other models._
## Testing Done
Added a new test case in our convergence tests, to ensure that the patched cross entropy loss works.
- Hardware Type: A100 SXM 80GB


ByronHsu:
This PR is covered by https://github.com/linkedin/Liger-Kernel/pull/375. Feel free to close it and open a new one on the main's branch for the follow-up works (e.g. rms, rope functional)

PR #375, opened 2024-11-12
Support CE after grad acc fix

ByronHsu:
## Summary

Based on https://github.com/linkedin/Liger-Kernel/pull/374, but make it leaner
1. The use of cross entropy in model code has changed after grad fix
2. It changed from module CrossEntropy to functional cross_entropy
3. Our monkey patching needs to change accordingly
4. While also make sure backward compatibility by adding a condition for different versions

Notable Changes

1. Add a functional api for CE to take keyword args
2. Add back conv test with logits to test CE convergence
3. Add back comp test for transformers 4.44

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,3 +64,28 @@ jobs:
     - name: Run unit tests
       run: |
         modal run dev.modal.tests
+
+  tests-bwd:
+    runs-on: ubuntu-latest
+    needs: [checkstyle]
+    env:
+      MODAL_TOKEN_ID: ${{ secrets.MODAL_TOKEN_ID }}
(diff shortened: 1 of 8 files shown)


ByronHsu:
For the ops not available in torch, make whatever args we have as kwargs

For the ops present in torch, follow the following rule

1. make the argument order align with torch. For example, if torch rmsnorm has `(a=..., b=..., c=...)`, then we also follow this order
2. for args present only in torch, pass a None as a default value, and do a no-op for that argument
3. for args present only in liger, put it after torch args and pass our default value

```
def my_functional(
   torch_arg_1 = <default>,
   torch_arg_2 = <default>,
   ...
   liger_arg_1 = <default>
   ...
)
```

PR #381, opened 2024-11-14
Refactor `LigerFusedLinearPreferenceBase`

pramodith:
## Summary
This PR refactors the `LigerFusedLinearPreferenceBase` class to contain an abstractmethod corresponding to the calculation of the loss that needs to be implemented by all sub-classes.

It also adds a new function to the class called `_compute_loss` which is mostly the same as the `_compute_orpo_loss` function introduced in #362 but makes it generic to calculate the NLL/Cross Entropy Loss plus accepts a custom loss function that implements a new alignment loss function.

Most RLHF/RLAIF/Alignment algorithms state their final loss as `NLL + Beta * (Alignment_Loss) `so adding the NLL logic inside the base class reduces repeated code.

The _compute_loss function accepts

## Testing Done

On A100-80G-SXM

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/chunked_loss/orpo_loss.py
+++ b/src/liger_kernel/chunked_loss/orpo_loss.py
@@ -23,63 +23,6 @@ def odds_ratio_loss(chosen_logps, rejected_logps, beta=0.1):
     return beta * ratio.sum()
 
 
-def _compute_orpo_loss(
-    input_chunk,
-    weight,
-    target_chunk,
-    bias=None,
-    full_target=None,
-    ignore_index=-100,
-    beta=0.1,
-    compute_nll_loss=True,
-):
-    """
-    Compute ORPO loss for a chunk of input and target.
-    Args:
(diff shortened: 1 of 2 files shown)


src/liger_kernel/chunked_loss/orpo_loss.py
```diff
@@ -98,15 +41,18 @@ def forward(
...
-        orpo_loss_fn = partial(
-            _compute_orpo_loss,
-            full_target=target,
+
+        return LigerFusedLinearPreferenceBase.forward(
+            ctx,
+            _input,
+            weight,
+            target,
+            bias,
```
shivam15s, line 50: I wrongly assumed that forward just takes positional arguments. 
Could you also make input, target, bias keyword args?

shivam15s:
This was quick. Great refactor!

shivam15s (dismissed):
lgtm

src/liger_kernel/chunked_loss/orpo_loss.py
pramodith: Done!

shivam15s (approved):
lgtm

PR #405, opened 2024-11-21
add reference model logps to chunkedloss interface and fix dpo loss fn

shivam15s:
accomodate reference model logps in chunked loss interface and make dpo loss use reference model logps in its loss function
## Summary

as title

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/chunked_loss/fused_linear_preference.py
+++ b/src/liger_kernel/chunked_loss/fused_linear_preference.py
@@ -19,7 +19,9 @@ class LigerFusedLinearPreferenceBase(torch.autograd.Function):
         raise NotImplementedError("Preference loss function must be implemented.")
 
     @staticmethod
-    def get_ref_logps(input_chunk, ref_weight, target_chunk, ref_bias=None, ignore_index=-100):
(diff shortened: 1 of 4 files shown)


src/liger_kernel/chunked_loss/fused_linear_preference.py
```diff
@@ -18,6 +18,30 @@ def preference_loss_fn(chosen_logps, rejected_logps, beta=0.1):
         """
         raise NotImplementedError("Preference loss function must be implemented.")
 
+    @staticmethod
+    def get_ref_logps(
+        input_chunk, ref_weight, target_chunk, ref_bias=None, ignore_index=-100
+    ):
+        with torch.no_grad():
+            ref_logits_chunk = input_chunk @ ref_weight.t()
```
pramodith, line 26: The logic to get the log_probs is common for both the reference and active policy model, wondering if we can create a util function that can be called by both instead of repeating the code.

pramodith (commented):
Great work! Wrapping DPO up.

pramodith (approved):
Thanks for the refactor!

Issue #371, opened 2024-11-08
Liger FlexChunkLoss: Supporting various alignment (DPO, ORPO, IRPO, CPO, etc) and distillation (KL divergence, cosine_similarity, earth_mover_distance, etc)  loss functions


shivam15s:
### Approach Overview:
The core idea is to extend the methods used in chunked Fused Linear Cross Entropy (FLCE) to various alignment algorithms. Here's how the process is structured:
1. Modular Optimization Process:
    - Every alignment algorithm’s optimization can be broken into three key steps:
        - Linear layer computation
        - Loss computation
        - Gradient calculation
2. Fused Linear and Loss Computation:
    - Similar to FLCE, we aim to fuse the linear layer with the loss computation for efficiency.
3. Chunking & Forward Optimization:
    - Since this is the final step in the model’s forward pass, we can also compute gradients directly during the forward pass instead of waiting for a separate backward pass.
    - We also chunk the input within the forward pass of the model, allowing significant reduction in peak gpu memory required. 
4. Torch Compile for Kernel Optimization:
    - Instead of manually handling kernel-level optimizations, we let torch.compile automatically optimize kernel execution. This reduces the need for low-level optimizations while still achieving performance gains.

By combining these strategies, we efficiently optimize alignment algorithms while also simplifying development.

### Key Findings
By leveraging torch.compile alongside optimization techniques like chunking, online softmax, etc, we observed close to custom triton kernel performance and reduced development time. This is why we want to introduce torch.compile as a key component of Liger.
References:
1. #227 
3. https://gist.github.com/Chillee/22cd93e11b887db1f596ab754d60a899#file-lce_benchmark-py

### Interface
Have a base class `FlexChunkLoss` that handles chunking, accumulation and compiling strategies.
A custom loss class wraps the  `FlexChunkLoss` and implements the loss fn that operates on a given chunk.
```py
class Mycustomloss(FlexChunkLoss):
  def loss_fn(...):
    ..do something here
```

hongpeng-guo:
I can take fused linear kl div. BTW, really nice illustration on the chunk linear op fusion from the paper. Very clear to new contributors 😄

pramodith:
@shivam15s @ByronHsu I think we should also consider including some of the loss functions commonly used for training embedding models, especially the popular ones supported in Sentence transformers.

[...]

PR #408, opened 2024-11-27
Introduce Distillation with a Chunked, Fused Linear JS-divergence Loss

austin362667:
#### Knowledge Distillation

```math
\mathcal{L} = \mathcal{L}_{\text{distill}}(\text{softmax}(\mathbf{z_t}, T), \text{softmax}(\mathbf{z_s}, T)) + \lambda \mathcal{L}_{CE}(\mathbf{y}, \mathbf{z_s}),
``` 

Here, `lambda` is a hyperparameter that balances the distillation loss and the supervised objective.

#### Jensen-Shannon Divergence Loss

In addition to adding the base class, this PR implements Jensen-Shannon Divergence (JSD) loss as the soft learning objective in the distillation setting. This component can be replaced with other losses (e.g., KL divergence) as `distillation_loss_fn`.

JSD is defined as the average of the KL divergences between each distribution and the mean distribution:

```math
\text{JSD}(P || Q) = \frac{1}{2} \text{KL}(P || M) + \frac{1}{2} \text{KL}(Q || M), \quad \text{where } M = \frac{1}{2}(P + Q)
``` 

Here, `P`and `Q` are the two probability distributions, and `M` is their average.

 - [ ] Investigate why the speed of chunked implementation is **so slow** compared to the naive approach.
 - [ ] Integrate temperature scaling.

## Testing Done

Yes. 

```
> modal run dev.modal.tests
========== 656 passed, 215 skipped, 58 warnings in 233.39s (0:03:53) ===========
```

jsd_loss_memory

jsd_loss_speed

- Hardware Type: A100 40G


src/liger_kernel/chunked_loss/fused_linear_distillation.py
```diff
@@ -0,0 +1,278 @@
...
+            -1, label_chunk.unsqueeze(-1)
+        ).squeeze(-1)
+
+        loss_mask_sum = loss_mask.sum(-1)
+        valid_mask = loss_mask_sum > 0
+
+        if valid_mask.any():
+            student_average_log_prob[valid_mask] = (
+                student_per_token_logps * loss_mask
+            ).sum(-1)[valid_mask] / loss_mask_sum[valid_mask]
```
Tcc0403, line 62: I'm not quite understand what `loss_mask_sum` and `valid_mask` do. Is it just a way to avoid ZeroDivisionError?

src/liger_kernel/chunked_loss/fused_linear_distillation.py
```diff
@@ -0,0 +1,278 @@
...
+        student_weight,
+        teacher_input,
+        teacher_weight,
+        target,
+        student_bias=None,
+        teacher_bias=None,
+        loss_fn=None,
+        chunk_size=1,
+        ignore_index=-100,
+        beta=0.5,
```
Tcc0403, line 95: Perhaps we need another variable name for the weight between soft and hard loss, since some loss functions have 'beta' parameter, such as [generalized jsd](https://arxiv.org/abs/2306.13649) we've implemented in #278.

test/utils.py
```diff
@@ -509,3 +507,144 @@ def get_batch_loss_metrics(
...
+            loss = loss_fct(logits, labels)
+            return loss
+
+        labels = target
+        ce_loss = cross_entropy_loss(
+            student_logits.view(-1, student_logits.shape[-1]),
+            labels.view(-1),
+        )
+
+        student_logps = self.get_batch_logps(
```
Tcc0403, line 603: Do we need to calculate the probability per token for knowledge distillation? I might be wrong but don't we just pass teacher_logits and student_logits directly to divergence loss function, such as kldiv(normally with reduction="batchmean") or jsd?

src/liger_kernel/chunked_loss/fused_linear_distillation.py
```diff
@@ -0,0 +1,278 @@
...
+                target_chunk.view(-1),
+                reduction="sum",
+                ignore_index=ignore_index,
+            )
+
+        loss_mask = target_chunk != ignore_index
+        label_chunk = torch.where(loss_mask, target_chunk, 0)
+
+        student_average_log_prob = torch.zeros_like(loss_mask, dtype=torch.float)
+        student_per_token_logps = student_log_probs_chunk.gather(
```
Tcc0403, line 52: same question as above

test/utils.py
austin362667: @Tcc0403  Thank you for the review!!

Actually, you're right—I’m aware of that. I was just trying to align the interface with the preference-based design. However, if it's not necessary to maintain the same interface, I prefer your suggestion.

As shown in the preference calculation function in this PR, it essentially undoes the operations. This redundant computation could be avoided by directly passing the raw logits to the divergence function, instead of first converting them to log probabilities and then reversing them back to the original values.

Tcc0403:
> [EDIT] I found the speed in Triton-based, fused-linear JSD loss is also much slower than the naive torch JSD impl in forward pass.

There're some benchmark in #300. If comparing forward pass only, fljsd kernel is supposed to be slower since it does gradient calculations in forward pass as well, and it isn't purely written in triton so it might also suffer from kernel launching overhead. But it's true that it doesn't perform well in low BT scenario.

src/liger_kernel/chunked_loss/fused_linear_distillation.py
```diff
@@ -0,0 +1,278 @@
...
+
+class LigerFusedLinearDistillationBase(torch.autograd.Function):
+
+    @abstractmethod
+    def distillation_loss_fn(student_logps, teacher_logps):
+        """
+        Compute distillation loss.
+        Args:
+            student_logps (torch.Tensor): Avg log probabilities of student inputs. Shape: (batch_size, hidden_size,).
+            teacher_logps (torch.Tensor): Avg log probabilities of teacher inputs. Shape: (batch_size, hidden_size,).
```
hongpeng-guo, line 16: I think for the general distillation loss, the student and teacher logps should be per-token instead of being averaged in the sequence length dimension. I.e., both tensors should be of shape (bathc_size, sequence_size, vocab_size) or (flattended_batch_sequence_size, vocab_size)

src/liger_kernel/chunked_loss/fused_linear_distillation.py
```diff
@@ -0,0 +1,278 @@
...
+                compute_ce_loss=compute_ce_loss,
+            )
+        )
+
+        ce_loss = ce_loss / (full_target != ignore_index).sum()
+
+        distillation_loss = distillation_loss_fn(
+            student_logps, teacher_logps, temperature
+        )
+        distillation_loss = distillation_loss / (full_target.shape[0])
```
hongpeng-guo, line 275: After we made the distillation loss per_token, we may normalize the `distillation_loss` with `full_target != ignore_index).sum` similar to the ce_loss.

hongpeng-guo (commented):
Thanks a lot for drafting the distillation base class, left some comments on the `fused_linear_distillation.py`, mainly discussing the loss should be computed per token or averaged from the sequence level first.

Anther major question that I am having is on the `chunking` dimensions. Current implementation of this PR is just chunking from the `batch_size` dimension, which is similar to the implementation of `fused_linear_preference.py`. However, I think it would be better if we can chunk from the flattened dim[0] of (B*T, vocab_size), which is also the way of chunking described in the [paper](https://arxiv.org/abs/2410.10989) for CE_loss.

For preference_base class, I think the chunking only happens on the batch_size dimension because the sequence dimension is reduced when calculating the average logps [(link).](https://github.com/linkedin/Liger-Kernel/blob/7e0f459149d298c84f162363cc6f1347494b80f2/src/liger_kernel/chunked_loss/fused_linear_preference.py#L51) . But for distillation, we may prefer to follow the patten of CE loss to chunk on the joint dimension of B*T, so that this kernel can work for very long sequence/ context scenario. Happy to help refine this base class @austin362667 

cc @shivam15s what do you think on this?

austin362667:
@hongpeng-guo Thanks for review~

> For preference_base class, I think the chunking only happens on the batch_size dimension because the sequence dimension is reduced when calculating the average logps [(link).](https://github.com/linkedin/Liger-Kernel/blob/7e0f459149d298c84f162363cc6f1347494b80f2/src/liger_kernel/chunked_loss/fused_linear_preference.py#L51) . But for distillation, we may prefer to follow the patten of CE loss to chunk on the joint dimension of B*T, so that this kernel can work for very long sequence/ context scenario. Happy to help refine this base class @austin362667

That makes perfect sense to me; I'll proceed with this approach.

> @austin362667 nit: A side note is to split this PR into two stacked PRs: first for the distillation base class and second for the JSDloss based from it. We can prioritize to polish and merge the first PR so that other distillation losses can be based on it and it's non-blocking 😄

Absolutely! I'll split this into two separate PRs.

austin362667:
Move discussion to https://github.com/linkedin/Liger-Kernel/pull/417

PR #399, opened 2024-11-20
[1/2] Enable  keyword  arguments for liger functional

hongpeng-guo:
## Summary

This PR is the first step to enable the keyword arguments of liger functional #368. 

1. Warp the Liger Operator Functions (`torch.autograd.Function`) with an extra layer that can take key word arguments.
2. For each of the liger functions, updating its unit test function `test_{operator_name}.py::test_correctness_functional` to reflect that keyword args can be accepted.
3. **Next step** is do apply the new op functionals in `transformers/monkey_patch.py`

## Testing Done

- Hardware Type:  A10G
- [x] run `make test` to ensure correctness
- [x] run `make checkstyle` to ensure code style
- [x] run `make test-convergence` to ensure convergence

PR #400, opened 2024-11-20
Enable keyword arguments for liger functional 

hongpeng-guo:
## Summary

This PR enables the keyword arguments of liger functional #368. 

1. Warp the Liger Operator Functions (`torch.autograd.Function`) with an extra layer that can take key word arguments.
2. For each of the liger functions, updating its unit test function `test_{operator_name}.py::test_correctness_functional` to reflect that keyword args can be accepted.

## Testing Done

- Hardware Type:  A10G
- [x] run `make test` to ensure correctness
- [x] run `make checkstyle` to ensure code style
- [x] run `make test-convergence` to ensure convergence

PR #417, opened 2024-12-02
Introduce Knowledge Distillation Base

austin362667:
## Summary

Thanks to the nice suggestions from @Tcc0403  and @hongpeng-guo. This PR is the first split from https://github.com/linkedin/Liger-Kernel/pull/408, focusing solely on introducing the Knowledge Distillation base class. As a result, this PR does not include any tests at the moment.

#### Code Changes

1. Refactor `beta` into two weights: `weight_hard_loss` and `weight_soft_loss`, as coefficients between `hard_loss` and `soft_loss`. @Tcc0403 also pointed out that we could use `torch.lerp` if applicable.

2. Pass `teacher_logits` and `student_logits` directly to the divergence loss function. This avoids redundant computations of converting logits to log probabilities and then reverting them to raw logits. However note that we are not reusing the `student_log_probs` value calculated during `ce_loss` in distillation base.

    1. Remove the unnecessary `get_batch_logps` in `test/utils.py`.

3. Modify `chunking` dimensions from `B` to `B * T`. Thanks to @hongpeng-guo's great advice.
    1. Fix the loss calculation to use per-token values instead of averaging across the sequence length dimension.

4. Normalize the `distillation_loss` using `(full_target != ignore_index).sum()`.

#### TODO  

1. [X] Although a slightly slowdown is reasonable, we need to investigate why this PR's implementation is **significantly slower** compared to the naive approach. Thanks to @Tcc0403 's clarification.
    
    The issue arises because we are not properly configuring the `chunk_size` for the `B * T` dimension, which is extremely large (a few thousand). The previous default of 1 results in an excessive number of chunks.

    In contrast, this problem does not occur with the preference loss, as chunking is performed on the `B` dimension. This produces fewer than 10 chunks, which is efficient and works as expected.

    In conclusion, I set `chunk_size` to `1024` works pretty well in new benchmark results as shown in https://github.com/linkedin/Liger-Kernel/pull/425

#### Knowledge Distillation

```math
\mathcal{L}_{\text{knowledge distillation}} = \mathcal{w}_{\text{soft}} \cdot \mathcal{L}_{\text{distill}}(\mathbf{z_t}, \mathbf{z_s}, T) + \mathcal{w}_{\text{hard}} \cdot \mathcal{L}_{\text{cross entropy}}(\mathbf{y}, \mathbf{z_s}),
``` 

Here,  we directly pass in `logits` rather than `logpbs`. @Tcc0403 

## Testing Done

I'll post JSD tests and benchmarks results in next PR: https://github.com/linkedin/Liger-Kernel/pull/425

- Hardware Type: L40S


test/utils.py
```diff
@@ -509,3 +509,112 @@ def get_batch_loss_metrics(
...
+            student_bias,
+            teacher_bias,
+        )
+        (
+            student_logits,
+            teacher_logits,
+            hard_loss,
+        ) = forward_output
+
+        soft_loss = self.distillation_loss(student_logits, teacher_logits)
```
winglian, line 617: the method use logprobs : `def distillation_loss(self, student_logps, teacher_logps):` but you use logits here.

I'd actually like to see both a logit and logprob implementation since it's easy to get logprobs offline from vllm and that is a faster way to generate the dataset.

austin362667: > the method use logprobs : def distillation_loss(self, student_logps, teacher_logps): but you use logits here.

@winglian Nice catch! Thank you so much.

> I'd actually like to see both a logit and logprob implementation since it's easy to get logprobs offline from vllm and that is a faster way to generate the dataset.

I think it's doable. And, I'm not quite sure I fully understand the need for `logprobs` implementation. Could you elaborate more on the vLLM use case?

winglian: So rather than having to have the teacher model loaded during training, depending on the workload type, it can be faster and more compute efficient to pre-compute the logins/logprobs offline beforehand. However, vllm and sglang only provide the logprobs, and that's not easily back-calculated to logits.

austin362667: I see. That makes a lot sense to me. Thank you!

austin362667:
> Although a slightly slowdown is reasonable, we need to investigate why this PR's implementation is significantly slower compared to the naive approach. Thanks to @Tcc0403 's clarification.

The issue arises because we are not properly configuring the `chunk_size` for the `B * T` dimension, which is extremely large (a few thousand). The previous default of 1 results in an excessive number of chunks.

In contrast, this problem does not occur with the preference loss, as chunking is performed on the `B` dimension. This produces fewer than 10 chunks, which is efficient and works as expected.

In conclusion, I set `chunk_size` to `1024` works pretty well in new benchmark results as shown in https://github.com/linkedin/Liger-Kernel/pull/425

hongpeng-guo (commented):
@austin362667  FWIW, to run the Modal GPU CIs, this PR needs to be made from the main repo, i.e., `linkedin/Liger-Kernel`, instead of the forked repo. 
A similar example is: I closed #399 and moved to #400 to enable the CI pipeline.

shivam15s (approved):
LGTM

test/utils.py
shivam15s: I believe we can address this ask in a subsequent PR 
@ByronHsu what do you think?

shivam15s (changes requested):
can you create another PR in linkedin? Some tests fail for me locally so I'd like to confirm before merging

austin362667:
@shivam15s  Certainly, right here https://github.com/linkedin/Liger-Kernel/pull/432 Thanks a lot

austin362667:
Move discussion to https://github.com/linkedin/Liger-Kernel/pull/432

PR #420, opened 2024-12-02
 Add weight support for LigerCrossEntropy

### LINKED ISSUE #404: Weighted Cross Entropy Loss

Allow for passing a weighting tensor to weight the CEL similar to C-RLFT where some tokens or inputs in the batch may have lower weighting to the overall loss. see https://arxiv.org/abs/2309.11235

### Alternatives

_No response_

### Additional context

_No response_

Tcc0403:
## Summary

Resolve #404.
Note: current implementation doesn't weight z loss.

Reference: [PyTorch's CrossEntropyLoss](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html)
## Testing Done

It hasn't fully tested with other params.

- Hardware Type: <BLANK>


src/liger_kernel/ops/cross_entropy.py
```diff
@@ -50,18 +54,22 @@ def liger_cross_entropy_kernel(
     X_stride (int): The stride of the input tensor.
     Y_ptr: Pointer to target tensor.
     Y_stride (int): The stride of the target tensor.
+    weight_ptr: Pointer to weight tensor.
+    weight_stride (int): The stride of the weight tesnor.
```
pramodith, line 58: seems like this argument isn't being used in this function.

src/liger_kernel/ops/cross_entropy.py
```diff
@@ -162,7 +173,12 @@ def liger_cross_entropy_kernel(
         X_block = tl.where(X_offsets != y, X_block, X_block - (1 - label_smoothing))
         # reduction scale
         if reduction == "mean":
-            X_block = X_block / (n_non_ignore)
+            if HAS_WEIGHT:
+                X_block = X_block / (sum_of_non_ignore_weight)
+            else:
+                X_block = X_block / (n_non_ignore)
+        if HAS_WEIGHT:
+            X_block = X_block * weight
```
pramodith, line 181: For my understanding is the multiplication of X_block with the weight here to help with the backward gradient computation?

src/liger_kernel/ops/cross_entropy.py
```diff
@@ -254,6 +279,23 @@ def cross_entropy_forward(
...
+        assert (
+            weight.shape[0] == V
+        ), f"If given, weight has to be a Tensor of size V. Got: {weight.shape}"
+        assert torch.is_floating_point(
+            weight
+        ), f"If given, weight has to be a Tensor of floating point dtype. Got: {weight.dtype}"
+        if ignore_index >= 0 and ignore_index < V:
+            weight_mask = torch.ones_like(weight)
+            weight_mask[ignore_index] = 0
+            selected_weight = torch.gather(weight * weight_mask, dim=-1, index=target)
```
pramodith, line 293: Wondering if there's any benefit to doing a torch.gather here compared to summing over `weight` after setting:

`weight[ignore_index]` = 0.

src/liger_kernel/ops/cross_entropy.py
```diff
@@ -343,6 +390,7 @@ def forward(
         ctx : The context object.
         _input (tensor): The input tensor of shape (BT, V) where B is batch size, T is sequence length, V is vocab size.
         target (tensor): The target tensor of shape (BT) where each value is in [0, V-1].
+        weight(Tensor, optional): a manual rescaling weight given to each class. If given, has to be a Tensor of size C and floating point dtype
```
pramodith, line 393: nit: Shouldn't the doc say a `Tensor of size V`?

test/transformers/test_cross_entropy.py
```diff
@@ -687,6 +709,41 @@ def test_correctness_with_z_loss_with_other_params_once(
...
 
+@pytest.mark.parametrize(
+    "B, T, V",
+    [
+        (2, 4096, 32000),  # llama2, mistral
+        # # weird shapes
+        (3, 423, 32000),
+    ],
+)
+@pytest.mark.parametrize("weight", [0.5, 0.1])
```
pramodith, line 720: I think this param can be removed, since the weight tensor is being generated from rand.

pramodith (changes requested):
Thanks for taking care of this! Had a few minor suggestions.

Another TODO is based on the original paper linked in the original issue for this feature. We also need to support a sample level weight. i.e. a weight that can be applied to each element of the batch if we have logits in the shape (B, S, V). We'd have sample level weights of shape (B, ). This is what's proposed in the C-RLFT paper. https://arxiv.org/abs/2309.11235

src/liger_kernel/ops/cross_entropy.py
Tcc0403: yes, it's part of the derivative of the unreduced loss [in the first equation](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html)

src/liger_kernel/ops/cross_entropy.py
pramodith: Hmmm, but we don't compute the gradient on the weight tensor, so I'm guessing that's ok.

Tcc0403: This part of code needs to be fixed. Becuase if there's a out of range index in `target` like `-100`, it will omit error.

src/liger_kernel/ops/cross_entropy.py
Tcc0403: removed!

src/liger_kernel/ops/cross_entropy.py
```diff
@@ -50,18 +54,22 @@ def liger_cross_entropy_kernel(
...
     n_non_ignore (int): The number of non-ignored elements in the batch.
+    sum_of_non_ignore_weight (float): The denominator when `reduction="mean"` if `weight` is given.
     ignore_index (int): The index to ignore in the target.
     label_smoothing (float): The amount of smoothing when computing the loss, where 0.0 means no smoothing.
     lse_square_scale (float): The scaler of (logsumexp(_input)) ^ 2 adding to the loss for the stability of training.
     RETURN_Z_LOSS (int): The boolean value to decide whether storing z loss to z_loss_ptr or not. It must be 0 or 1.
     reduction (str): The string for the reduction to apply
     softcap (float): The upper threshold for scaling logits to the range (-softcap, +softcap).
     BLOCK_SIZE (int): The block size for Triton operations.
+    HAS_WEIGHT (bool): The boolean value to dteremine whether assigning weight to each of the classes.
```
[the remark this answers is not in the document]
Tcc0403: nice catch

Tcc0403:
Feel free to push to this branch or even take over it and open a new PR, I won't be able to update that often in the next few months. Just trying to make the first step when I got time.

pramodith:
> Feel free to push to this branch or even take over it and open a new PR, I won't be able to update that often in the next few months. Just trying to make the first step when I got time.

Gotcha! I'll try wrapping it up, you've done most of the heavy lifting already.

test/transformers/test_cross_entropy.py
```diff
@@ -687,6 +788,110 @@ def test_correctness_with_z_loss_with_other_params_once(
...
+            1e-8,
+            5e-2,
+            marks=pytest.mark.skipif(
+                not supports_bfloat16(), reason="bfloat16 not supported on this GPU"
+            ),
+        ),
+        (1.0, torch.float32, 1e-8, 1e-6),
+    ],
+)
+def test_correctness_with_weight_with_other_params_once(
```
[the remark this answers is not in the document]
pramodith: So, the issue seems to be with combining `label_smoothing` with `weighted` loss. I've been staring at the code and equations for a while now but I can't pinpoint anything that's wrong. Simply multiplying the final loss with the weight of the label token seems like the right thing to do to me.

If not there can only be an issue with the:

`scaled_x_sum` term since all the other terms in smoothed loss are also a part of the plain ce loss which we know works correctly.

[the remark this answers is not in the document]
Tcc0403: Figuring out where it doesn't work is a big! I'll take a look on Saturday.

Tcc0403:
I took a look at torch's impl, and here's how they compute `smooth_loss`
https://github.com/pytorch/pytorch/blob/2682e5e0d48a8200c1672b6a42250d3c8de44190/aten/src/ATen/native/LossNLL.cpp#L558
```python
    if (weight.defined()) {
      // Expand weight to the correct number of dims for broadcasting with input / target
      auto weight_broadcast_shape = SmallBuffer<int64_t, 5>(input.dim());
      std::fill(weight_broadcast_shape.begin(), weight_broadcast_shape.end(), 1);
      weight_broadcast_shape[class_dim] = weight.size(0);
      Tensor weight_ = weight.view(weight_broadcast_shape);

      smooth_loss = -(input * weight_).sum(class_dim);
```
related code blocks in liger:
https://github.com/linkedin/Liger-Kernel/blob/bd65c47999cebc2ac3dce39447ecec051b8b6159/src/liger_kernel/ops/cross_entropy.py#L119
https://github.com/linkedin/Liger-Kernel/blob/bd65c47999cebc2ac3dce39447ecec051b8b6159/src/liger_kernel/ops/cross_entropy.py#L194-L196

src/liger_kernel/ops/cross_entropy.py
```diff
@@ -253,7 +277,22 @@ def cross_entropy_forward(
...
+        assert (
+            weight.shape[0] == V
+        ), f"If given, weight has to be a Tensor of size V. Got: {weight.shape}"
+        assert torch.is_floating_point(
+            weight
+        ), f"If given, weight has to be a Tensor of floating point dtype. Got: {weight.dtype}"
+        selected_weight = torch.where(
+            target_mask, torch.gather(weight, dim=0, index=target * target_mask), 0.0
+        )
+        sum_of_non_ignore_weight = selected_weight.sum().item()
```
Tcc0403, line 293: we can rewrite it with torch.masked_select
```python 
sum_of_non_ignore_weight = (torch.gather(weight, dim=0, index=target.masked_select(target_mask))
            .sum()
            .item()
        )
```

winglian:
@pramodith anything I can do to help with this PR?

pramodith:
> @pramodith anything I can do to help with this PR?

Hey @winglian I won't be able to look into this any further, feel free to take over and see if you can figure out the source of discrepancy. The tests fail when combining smoothing loss with weighted ce.

Tcc0403:
I'll make an another PR for sample level weight.

src/liger_kernel/ops/fused_linear_cross_entropy.py
```diff
@@ -54,7 +55,25 @@ def fused_linear_cross_entropy_forward(
     loss_1d = torch.zeros(BT, dtype=torch.float32, device=device)
 
     # NOTE: skip .item() here to avoid CUDA synchronization
-    total_n_non_ignore = (target != ignore_index).sum()
+    target_mask = target != ignore_index
+    total_n_non_ignore = target_mask.sum().item()
```
bboyleonp666, line 59: I have noticed the comment above to avoid using multiple times. Will this change align the synchronization?

src/liger_kernel/ops/fused_linear_cross_entropy.py
```diff
@@ -54,7 +55,25 @@ def fused_linear_cross_entropy_forward(
...
+            ce_weight
+        ), f"If given, weight has to be a Tensor of floating point dtype. Got: {ce_weight.dtype}"
+        total_sum_non_ignore_ce_weight = (
+            torch.gather(ce_weight, dim=0, index=target.masked_select(target_mask))
+            .sum()
+            .item()
+        )
+        ce_weight_sum = ce_weight.sum().item()
+        if ce_weight.stride(-1) != 1:
+            ce_weight = ce_weight.contiguous()
```
bboyleonp666, line 76: I have seen the **shape checking** and **torch.is_floating_point** for multiple times. I wonder if we can add another method to deal with these duplications.

src/liger_kernel/ops/fused_linear_cross_entropy.py
```diff
@@ -214,21 +226,23 @@ def forward(
...
+            _input=_input,
+            weight=weight,
+            target=target,
+            bias=bias,
+            ce_weight=ce_weight,
+            ignore_index=ignore_index,
+            lse_square_scale=lse_square_scale,
+            label_smoothing=label_smoothing,
+            reduction=reduction,
+            softcap=softcap,
```
bboyleonp666, line 245: I like the idea to explicitly declare the parameters to be used. Great work.

bboyleonp666 (commented):
Hi @Tcc0403, thanks for your wonderful work. I left some of my thoughts for this PR, PTAL.

src/liger_kernel/ops/cross_entropy.py
```diff
@@ -152,18 +171,42 @@ def liger_cross_entropy_kernel(
...
+            dloss_smooth = eps * (-weight_block + softmax_X * weight_sum)
+            # derivative of z-loss
+            dz_loss = 2 * lse_square_scale * lse * softmax_X
+            # reduction scale
+            if reduction == "mean":
+                dloss_ori = dloss_ori / sum_non_ignore_weight
+                dloss_smooth = dloss_smooth / sum_non_ignore_weight
+                dz_loss = dz_loss / n_non_ignore
+            # derivative of total_loss
+            X_block = dloss_ori + dloss_smooth + dz_loss
```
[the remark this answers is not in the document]
Tcc0403: I thought putting the base case (none weight) first would be better, but I'll consider it, thank you.

src/liger_kernel/ops/cross_entropy.py
```diff
@@ -192,17 +237,23 @@ def liger_cross_entropy_kernel(
...
-    loss += z_loss
     # Normalize the loss by the number of non-ignored elements if reduction is "mean"
     if reduction == "mean":
+        if HAS_WEIGHT:
+            loss = loss / sum_non_ignore_weight
+        else:
+            loss = loss / n_non_ignore
         z_loss = z_loss / n_non_ignore
-        loss = loss / n_non_ignore
+    loss += z_loss
```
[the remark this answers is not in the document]
Tcc0403: zloss wasn't scaled by weight right now, so it is divided by number of non ignore token, unlike the rest part of loss divided by the sum of weight when weight exists.
That's why I have to do divisions first before summing them.

src/liger_kernel/ops/fused_linear_cross_entropy.py
Tcc0403: Forgot to remove the comment, it doesn't affect the result.

Tcc0403:
I'll update it on Saturday. Thanks for your review.

src/liger_kernel/ops/cross_entropy.py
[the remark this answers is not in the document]
bboyleonp666: Oh, I see your point. Never mind. I was a little too nitpicking.

src/liger_kernel/ops/cross_entropy.py
```diff
@@ -197,6 +197,7 @@ def liger_cross_entropy_kernel(
             if reduction == "mean":
                 dloss_ori = dloss_ori / sum_non_ignore_weight
                 dloss_smooth = dloss_smooth / sum_non_ignore_weight
+                # z_loss isn't scaled by weight
```
bboyleonp666, line 200: Correct me if I were wrong. I guess the comment is to point out the z loss is not scaled by weight, thus it's not divided it by `sum_non_ignore_weight` but `n_non_ignore`. In this case, I think adding a `TODO` for it will be a better idea.

src/liger_kernel/ops/cross_entropy.py
```diff
@@ -247,6 +248,7 @@ def liger_cross_entropy_kernel(
             loss = loss / sum_non_ignore_weight
         else:
             loss = loss / n_non_ignore
+        # z_loss isn't scaled by weight
```
bboyleonp666, line 251: ditto.

bboyleonp666 (commented):
It's great to see that all the tests are passed right now. I need some time for further checking the tests. But for algorithm, it already LGTM.

src/liger_kernel/ops/cross_entropy.py
Tcc0403: Makes sense to me. Thanks for your fast response.

austin362667 (approved):
LGTM :rocket:

PR #425, opened 2024-12-04
Add JSD Loss for Distillation

austin362667:
## Summary

> [!CAUTION]
> This PR depends on https://github.com/linkedin/Liger-Kernel/pull/417. Do not merge until https://github.com/linkedin/Liger-Kernel/pull/417 is merged.

This is a pure torch compiled, chunked fused linear JSD Loss, aiming for knowledge distillation.

#### Jensen-Shannon Divergence Loss

This PR implements Jensen-Shannon Divergence (JSD) loss as the soft learning objective in a distillation setting (teacher & student). This component can be replaced with other losses (e.g., KL divergence) as `distillation_loss_fn`.

JSD is defined as the average of the KL divergences between each distribution and the mean distribution:

```math
\text{JSD}(P || Q) = \frac{1}{2} \text{KL}(P || M) + \frac{1}{2} \text{KL}(Q || M), \quad \text{where } M = \frac{1}{2}(P + Q)
``` 

Here, `P`and `Q` are the two probability distributions, and `M` is their average.

## Testing Done

Benchmark results with different `chunk_size`, which significantly affect performance.

#### Hint: 
User can tune their `chunk_size` as suggested by the liger [paper](https://arxiv.org/pdf/2306.13649) for the moment: 
```math
2^{\lceil \log_2 \lceil \frac{BT}{V/H} \rceil \rceil}
``` 
 
#### Memory

1. `chunk_size` = 1

    distill_jsd_loss_memory_chunk_size_1

2. `chunk_size` = 1024

    distill_jsd_loss_memory_chunk_size_1024

#### Speed (Elapsed Time)

1. `chunk_size` = 1

    distill_jsd_loss_speed_chunk_size_1

2. `chunk_size` = 1024

    distill_jsd_loss_speed_chunk_size_1024

- Hardware Type: NVIDIA H100 80GB HBM3 (SXM5)


austin362667:
cc @mecoli1219

src/liger_kernel/chunked_loss/jsd_loss.py
```diff
@@ -0,0 +1,155 @@
...
+            student_logits (torch.Tensor): Logits of student tokens. Shape: (batch_size * seq_len,).
+            teacher_logits (torch.Tensor): Logits of teacher tokens. Shape: (batch_size * seq_len,).
+        Returns:
+            torch.Tensor: Jensen-Shannon Divergence loss
+        """
+        # Convert to probabilities
+        student_probs = F.softmax(student_logits, dim=-1)
+        teacher_probs = F.softmax(teacher_logits, dim=-1)
+
+        log_mean_probs = torch.log((student_probs + teacher_probs) / 2)
```
Tcc0403, line 22: Perhaps we can add beta support for general jsd as #278 did. And call logsoftmax directly for numerical stability.

Tcc0403 (changes requested):

austin362667:
We can refactor distillation base so that it aligns with the code structure of preference base in follow-up PRs.

Mecoli1219:
@austin362667 there is no place to set `beta` for the JSD loss function in this version, right?

test/chunked_loss/test_jsd_loss.py
```diff
@@ -0,0 +1,316 @@
...
+
+        # Compute probabilities (only required for mean calculation)
+        mean_probs = beta * student_log_probs.exp() + (1 - beta) * teacher_log_probs.exp()
+        log_mean_probs = mean_probs.log()
+
+        student_kl = F.kl_div(log_mean_probs, student_log_probs, reduction="batchmean", log_target=True)
+        teacher_kl = F.kl_div(log_mean_probs, teacher_log_probs, reduction="batchmean", log_target=True)
+
+        # JSD is the average of the KL divergences
+        jsd_loss = beta * student_kl + (1 - beta) * teacher_kl
```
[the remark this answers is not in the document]
austin362667: sure! that makes sense, thank you so much.

src/liger_kernel/chunked_loss/jsd_loss.py
```diff
@@ -0,0 +1,154 @@
...
+            torch.Tensor: Jensen-Shannon Divergence loss
+        """
+        student_log_probs = F.log_softmax(student_logits, dim=-1)
+        teacher_log_probs = F.log_softmax(teacher_logits, dim=-1)
+
+        # Compute probabilities (only required for mean calculation)
+        mean_probs = beta * student_log_probs.exp() + (1 - beta) * teacher_log_probs.exp()
+        log_mean_probs = mean_probs.log()
+
+        student_kl = F.kl_div(log_mean_probs, student_log_probs, reduction="sum", log_target=True)
```
Tcc0403, line 26: Should we just pass `reduction="batchmean"` and remove manually applying reduction in  `LigerFusedLinearDistillationBase`'s `_compute_loss` ?

https://github.com/linkedin/Liger-Kernel/blob/1441a310b241449b5626a6435aa88b44b58ebbc9/src/liger_kernel/chunked_loss/fused_linear_distillation.py#L113

So we can match `HFJSDLoss`'s `distillation_loss` implementation in unit tests. 

The reason I give the above suggestion is because I got confused when I saw two different reduction methods in liger and hf give the same result. It turns out that the reduction is hiding under `LigerFusedLinearDistillationBase`.

austin362667: Hey @Tcc0403, thanks for the advice. From what I can recall, this issue was discussed in [this thread](https://github.com/linkedin/Liger-Kernel/pull/442/files). The conclusion then was to just let it be as it is.

Tcc0403: Never mind! I was wrong!

If using `reduction=batchmean` instead, it will result in a wrong value after accumulating losses from all chunks. This is similar to the gradient accumulation bug in transformers before grad acc fix.




TL;DR
Sum of average losses from each chunk is not equal to average loss from the whole sequence
```math
averageLoss_{chunk1} + averageLoss_{chunk2} + ... + averageLoss_{chunkN} = N \times averageLoss_{total}\neq averageLoss_{total}
```

Tcc0403: nit: add a reminder in `distillation_loss_fn`'s docstring: This function should compute loss with no reduction (sum), and it will be normalized (mean)  in `_compute_loss` after accumulating from each chunk.

Tcc0403 (approved):
LGTM

PR #429, opened 2024-12-06
Add ORPO Trainer + support HF metrics directly from chunked loss functions + fixes to avoid torch compile recompilations

shivam15s:
## Summary

This PR adds support for the following:
1. LigerORPOTrainer: a wrapper on top of [HuggingFace ORPO Trainer](https://github.com/huggingface/trl/blob/main/trl/trainer/orpo_trainer.py) to use LigerORPOLoss module. 
2. We also provide an example for using LigerORPOTrainer in `examples/alignment/run_orpo.py`
3. Change FusedLinearPreference base class' forward function to return additional metrics to align our implementation with HF ORPO Trainer
4. Additional refactor to avoid torch compile recompilations -- accumulate_chunk function now calls accumulate_helper which is torch compiled solely and input_chunk/target_chunk/target dimension 1 (seq len) is explicitly marked as dynamic to avoid recompilations

## Testing Done

- Hardware Type: <BLANK>


ByronHsu (approved):
awesome work!!

PR #432, opened 2024-12-07
Introduce Knowledge Distillation Base

austin362667:
## Summary

Made https://github.com/linkedin/Liger-Kernel/pull/417 from the main repo.

Thanks to the nice suggestions from @Tcc0403  and @hongpeng-guo. This PR is the s first split from https://github.com/linkedin/Liger-Kernel/pull/408, focusing solely on introducing the Knowledge Distillation base class. As a result, this PR does not include any tests at the moment.

#### Code Changes

1. Refactor `beta` into two weights: `weight_hard_loss` and `weight_soft_loss`, as coefficients between `hard_loss` and `soft_loss`. @Tcc0403 also pointed out that we could use `torch.lerp` if applicable.

2. Pass `teacher_logits` and `student_logits` directly to the divergence loss function. This avoids redundant computations of converting logits to log probabilities and then reverting them to raw logits. However note that we are not reusing the `student_log_probs` value calculated during `ce_loss` in distillation base.

    1. Remove the unnecessary `get_batch_logps` in `test/utils.py`.

3. Modify `chunking` dimensions from `B` to `B * T`. Thanks to @hongpeng-guo's great advice.
    1. Fix the loss calculation to use per-token values instead of averaging across the sequence length dimension.

4. Normalize the `distillation_loss` using `(full_target != ignore_index).sum()`.

#### TODO  

1. [X] Although a slightly slowdown is reasonable, we need to investigate why this PR's implementation is **significantly slower** compared to the naive approach. Thanks to @Tcc0403 's clarification.
    
    The issue arises because we are not properly configuring the `chunk_size` for the `B * T` dimension, which is extremely large (a few thousand). The previous default of 1 results in an excessive number of chunks.

    In contrast, this problem does not occur with the preference loss, as chunking is performed on the `B` dimension. This produces fewer than 10 chunks, which is efficient and works as expected.

    In conclusion, I set `chunk_size` to `1024` works pretty well in new benchmark results as shown in https://github.com/linkedin/Liger-Kernel/pull/425

2. [ ] https://github.com/linkedin/Liger-Kernel/pull/417#discussion_r1874231427

#### Knowledge Distillation

```math
\mathcal{L}_{\text{knowledge distillation}} = \mathcal{w}_{\text{soft}} \cdot \mathcal{L}_{\text{distill}}(\mathbf{z_t}, \mathbf{z_s}, T) + \mathcal{w}_{\text{hard}} \cdot \mathcal{L}_{\text{cross entropy}}(\mathbf{y}, \mathbf{z_s}),
``` 

Here,  we directly pass in `logits` rather than `logpbs`. @Tcc0403 

## Testing Done

I'll post JSD tests and benchmarks results in next PR: https://github.com/linkedin/Liger-Kernel/pull/425

- Hardware Type: L40S


Tcc0403:
> Here, we directly pass in logits rather than logpbs.
Sorry for the misleading question and late response. Passing `logpbs` is totally fine, it's actually better that it can avoid underflow issues in the log-space. Torch's [KLDivLoss](https://pytorch.org/docs/stable/generated/torch.nn.KLDivLoss.html) also expect inputs in the log-space, and the extra amount of calculation from `softmax` to `logsoftmax` shouldn't be an issue anyway. So if most APIs are expecting input as logpbs, then I think it's the way to go.

What I questioned in the last comment was about the `xxx_per_token_xxx` things, but there's no problems with this PR now.

hongpeng-guo (approved):
LGTM.

austin362667:
Thanks @Tcc0403  and @hongpeng-guo .

I think we can merge this PR first? With some follow-ups in my mind:

1. Refactor to use `logprobs` for aligning `sglang`/`vllm` interfaces afterwards. [src](https://github.com/linkedin/Liger-Kernel/pull/432#issuecomment-2525024926)
2. Figure out  temperature scaling in  `logprobs`. [src](https://github.com/linkedin/Liger-Kernel/pull/417#discussion_r1874199456)
3. Pre-compute the `logits`/`logprobs` offline beforehand , rather than having to have the teacher model loaded during training. [src](https://github.com/linkedin/Liger-Kernel/pull/417#discussion_r1872484660)

@ByronHsu WDYT?

Issue #421, opened 2024-12-03
CrossEntropyLoss return single value when reduction is "none"


leng-yue:
### 🐛 Describe the bug

When using `LigerCrossEntropyLoss` with `reduction="none"`, it returns a single value, instead of an array.

### Reproduce

```python
import torch
from liger_kernel.transformers.cross_entropy import LigerCrossEntropyLoss


def test_cross_entropy_matches_torch():
    # Test cases with various shapes and values
    batch_size = 3
    num_classes = 5
    seq_length = 4

    # Generate random logits and labels using torch
    torch.manual_seed(42)
    logits = torch.randn(batch_size, seq_length, num_classes, device='cuda')
    labels = torch.randint(0, num_classes, (batch_size, seq_length), device='cuda')

    # Set some labels to -100 to ignore them
    labels[0, -1] = -100  # Ignore last token of first sequence
    labels[1, -2:] = -100  # Ignore last two tokens of second sequence

    # Liger cross entropy
    liger_ce = LigerCrossEntropyLoss(reduction="none")
    liger_loss = liger_ce(logits.view(-1, num_classes), labels.view(-1))
    print(liger_loss)

    # PyTorch cross entropy
    torch_ce = torch.nn.CrossEntropyLoss(reduction="none")
    torch_loss = torch_ce(logits.view(-1, num_classes), labels.view(-1))

    # Check if losses match
    torch.testing.assert_close(
        liger_loss,
        torch_loss,
        rtol=1e-5,
        atol=1e-5,
        msg="Liger and PyTorch losses don't match",
    )

    print("All tests passed! LigerCrossEntropyLoss matches PyTorch's implementation")


if __name__ == "__main__":
    test_cross_entropy_matches_torch()
```

### Versions

Environment Report:
-------------------
Operating System: Linux-5.15.0-88-generic-x86_64-with-glibc2.35
Python version: 3.10.15
PyTorch version: 2.5.1+cu124
CUDA version: 12.4
Triton version: 3.1.0
Transformers version: 4.46.2

ByronHsu:
Sorry. We only support "mean" and "sum" for now. Are you willing to contribute by adding an assertion or implementing "none" reduction? Thanks!

leng-yue:
I think I can make a PR to add an assert to block "none" first, then let's see how to implement the "none" reduction one.

Tcc0403:
[...]

PR #435, opened 2024-12-08
Fix LigerCrossEntropyLoss Reduction Behavior for "None" Mode

hebiao064:
## Summary

Closes https://github.com/linkedin/Liger-Kernel/issues/421 

This pull request addresses an issue in the `cross_entropy_forward` function where the `reduction="none"` mode did not behave as expected. 

Previously, the function always returned a single scalar value, even when reduction="none" was specified. This update ensures that when reduction="none" is used, the function directly outputs the unreduced loss array (loss_1d) instead of summing it.

### Changes Made:

- Added a condition to handle `reduction="none"`, ensuring the function outputs loss_1d directly.
- Updated the computation of z_loss to respect the reduction="none" mode.
- Add test for cases when `reduction="none"`

#### Background on Gradients in PyTorch

- **Scalar Outputs**: When a tensor is a scalar (a single number), PyTorch can compute gradients automatically by assuming the scalar has an implicit gradient of 1.0.
- **Non-Scalar Outputs**: For tensors that are not scalars, gradients must be provided explicitly because PyTorch cannot infer the shape or distribution of gradients. Without this, it raises the error: "grad can be implicitly created only for scalar outputs."

#### Why reduction="none" Needs Explicit Gradients

When `reduction="none"`, the loss function does not reduce the per-example loss values into a single scalar. Instead, it outputs a vector of losses, with one value per example in the batch. This means that the loss tensor has multiple values, and PyTorch cannot assume what the gradient for each of these values should be unless explicitly provided.

#### The Fix
By passing `gradient=torch.ones_like(loss)` to `backward()`:

- **Gradient Tensor**: The `torch.ones_like(loss)` serves as the gradient tensor. It specifies that each element in the loss tensor contributes equally to the gradients during backpropagation.
- **Shape Match**: The gradient tensor's shape matches the loss tensor's shape, fulfilling PyTorch's requirements for non-scalar outputs during backward().

## Testing Done

make test

`pytest /home/jobuser/Liger-Kernel/test/transformers/test_cross_entropy.py` shows:

```
=================================== 93 passed, 1 warning in 13.18s ===================================
```

### DIFF
--- a/src/liger_kernel/ops/cross_entropy.py
+++ b/src/liger_kernel/ops/cross_entropy.py
@@ -285,7 +285,11 @@ def cross_entropy_forward(
         num_warps=32 if not is_hip() else 16,
     )
 
-    loss = torch.sum(loss_1d)
+    if reduction == "none":
+        loss = loss_1d
+    else:
+        loss = torch.sum(loss_1d)
+
     if return_z_loss == _TRUE.value:
         z_loss = torch.sum(z_loss_1d)
     else:
--- a/src/liger_kernel/ops/cross_entropy.py
+++ b/src/liger_kernel/ops/cross_entropy.py
@@ -287,13 +287,10 @@ def cross_entropy_forward(
 
     if reduction == "none":
         loss = loss_1d
+        z_loss = z_loss_1d if return_z_loss == _TRUE.value else None
     else:
         loss = torch.sum(loss_1d)
-
-    if return_z_loss == _TRUE.value:
-        z_loss = torch.sum(z_loss_1d)
-    else:
-        z_loss = None
+        z_loss = torch.sum(z_loss_1d) if return_z_loss == _TRUE.value else None
 
     return loss, z_loss, _input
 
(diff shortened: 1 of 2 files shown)


Tcc0403:
LGTM!

src/liger_kernel/ops/cross_entropy.py
```diff
@@ -285,7 +285,11 @@ def cross_entropy_forward(
...
     )
 
-    loss = torch.sum(loss_1d)
+    if reduction == "none":
+        loss = loss_1d
+    else:
+        loss = torch.sum(loss_1d)
+
     if return_z_loss == _TRUE.value:
         z_loss = torch.sum(z_loss_1d)
```
Tcc0403, line 294: Let's add none reduction option for z_loss as well

hebiao064: Thanks for pointing it out, just updated, pls review.

Tcc0403 (approved):

ByronHsu:
Awesome work! Please push branch to the main repo next time, so you can run CI directly. We have disabled CI from the outside fork

PR #438, opened 2024-12-08
Fix All `chunked_loss` Benchmark Scripts

austin362667:
## Summary

With recent updates, we need to align the APIs used in the benchmark script to ensure consistency.

Code changes:

- Fix `.backward()` pass in `backward` and `full` mode. 
- Reuse `TorchLMHead` from `chunked_loss` tests.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/benchmark/scripts/benchmark_cpo_loss.py
+++ b/benchmark/scripts/benchmark_cpo_loss.py
@@ -19,16 +19,6 @@ device = infer_device()
 sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "../..")))
 
 
-class LigerLMHeadCPO(torch.nn.Module):
-    def __init__(self, H: int, V: int, dtype: torch.dtype, ignore_index: int = -100):
-        super().__init__()
-        self.lin = torch.nn.Linear(in_features=H, out_features=V, bias=False, dtype=dtype)
-        self.cpo_loss = LigerFusedLinearCPOFunction.apply
-
-    def forward(self, x, y):
-        return self.cpo_loss(x, self.lin.weight, y)
-
-
 #############################################################################
 # Test the memory consumption of the linear fused cross entropy loss
 #############################################################################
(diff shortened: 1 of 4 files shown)


benchmark/scripts/benchmark_dpo_loss.py
```diff
@@ -1,4 +1,5 @@
-from test.chunked_loss.test_dpo_loss import HF_DPO_Loss
```
shivam15s, line 1: I believe this fn wouldn't work because right now return is a tuple
https://github.com/linkedin/Liger-Kernel/blob/aeb8ce38b637eeecd92058ad6bbc0fac88fb257f/benchmark/scripts/benchmark_dpo_loss.py#L120C1-L122C21

shivam15s (changes requested):

benchmark/scripts/benchmark_dpo_loss.py
shivam15s: also I believe other chunked loss benchmarking scripts would face the same issue

austin362667: Got it! Thanks for pointing that out~

shivam15s (changes requested):

benchmark/scripts/benchmark_dpo_loss.py
```diff
@@ -65,13 +43,19 @@ def forward(self, x, target):
             self.lin.weight,
```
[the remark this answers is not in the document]
austin362667: Sure. @shivam15s  Thanks for review! 

I'm using `lambda` to create a function that will access the first element only when it's actually called.
Reason is that I want to reuse `TorchLMHead[CUSTOM]` in tests without modify them as well as for `TorchLMHead[CUSTOM]`.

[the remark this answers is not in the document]
shivam15s: Great catch, I forget about the tests dependency! Thanks for this fix!

shivam15s (approved):
lgtm

benchmark/scripts/benchmark_cpo_loss.py
```diff
@@ -62,15 +40,21 @@ def forward(self, x, y):
 def bench_memory_fused_linear_cpo_loss(
     input: SingleBenchmarkRunInput,
 ) -> SingleBenchmarkRunOutput:
+    from test.chunked_loss.test_cpo_loss import TorchLMHeadCPO
```
shivam15s, line 43: Curious why not also use LigerLMHeadCPO from tests if we're using TorchLMHeadCPO here

shivam15s (changes requested):

benchmark/scripts/benchmark_cpo_loss.py
austin362667: Ah certainly! Thanks. Updated~

Tcc0403 (approved):

PR #442, opened 2024-12-09
Fix Normalization Term in Distillation Loss 

austin362667:
## Summary

Based on this modification https://github.com/linkedin/Liger-Kernel/pull/432/commits/e381569d6c88283490887d76de70a824bb463b0c#r1875638713 introduced by @shivam15s.

I believe it's more accurate to compute chunked loss by normalizing it based on the "number of `chunks`". Please correct me if I'm mistaken—thanks!

## Testing Done

- Hardware Type: <BLANK>


shivam15s:
Hey @austin362667 that's a valid way to do normalization. However, I want to be consistent with the preference base and also avoid ops involving chunk sizes/shapes as this can change in the future.

shivam15s:
@austin362667 I believe that the chunk

PR #481, opened 2024-12-16
preference loss sign is inverted and leads to negative loss

winglian:
## Summary
In testing cases where there is no alpha/NLL loss used, the loss becomes negative, which is probably not the intended behavior. see https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L1234

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/chunked_loss/fused_linear_preference.py
+++ b/src/liger_kernel/chunked_loss/fused_linear_preference.py
@@ -408,7 +408,7 @@ class LigerFusedLinearPreferenceBase(torch.autograd.Function):
         else:
             preference_loss, aux_outputs = preference_loss_outputs, []
 
-        loss = alpha * chosen_nll_loss - preference_loss
+        loss = alpha * chosen_nll_loss + preference_loss
         return_vars = (
             chosen_logps,
             rejected_logps,
--- a/test/utils.py
+++ b/test/utils.py
@@ -511,7 +511,7 @@ class HFAlignmentLoss:
         else:
             losses, aggregated_aux_outputs = alignment_loss_outputs, []
         # full loss
-        loss = policy_nll_loss * self.alpha - losses.mean()
+        loss = policy_nll_loss * self.alpha + losses.mean()
         return_vars = (
             policy_chosen_logps,
             policy_rejected_logps,


austin362667 (approved):
Great catch. Thank you

austin362667 (changes requested):
Would you mind updating the corresponding tests? They failed in my CI as expected.

austin362667:
I believe the issue lies here: https://github.com/linkedin/Liger-Kernel/blob/0bb6c72d03f0fc570b9e954c42b723ab7e0c315f/test/utils.py#L514.

austin362667 (approved):
Thank you!

src/liger_kernel/chunked_loss/fused_linear_preference.py
```diff
@@ -408,7 +408,7 @@ def _compute_loss(
         else:
             preference_loss, aux_outputs = preference_loss_outputs, []
 
-        loss = alpha * chosen_nll_loss - preference_loss
```
shivam15s, line 411: Seems like this is correct for all losses except dpo_loss. I think the change the sign change should be here instead: https://github.com/linkedin/Liger-Kernel/blob/21bacccd107b40146e599e2c27af46d7d157f174/src/liger_kernel/chunked_loss/dpo_loss.py#L52

shivam15s (changes requested):

src/liger_kernel/chunked_loss/fused_linear_preference.py
austin362667: Hi @shivam15s  Thanks for noticing this. What are your thoughts on negating each preference loss term to align with the formulas in the docstrings? This would allow us to maintain the base preference structure as `nll_loss + preference_loss` while making both terms represent losses to be minimized.

shivam15s: Sure @austin362667 ! I think that might help with readability.
If you have some time, could you add the fix to this PR?

austin362667: Ya sure! I might need to open a new PR (base on this one) cc @winglian

PR #486, opened 2024-12-17
fix chosen_nll_loss in chunked losses

kashif:
## Summary
Fix the nll loss in the the chunked loses when the model is a decoder only model, by shifting the logits and targets

- Hardware Type: <BLANK>


shivam15s (approved):
Great catch, and thanks for the quick PR! With the TRL fixes we discussed today, we should be able to closely match the loss curve obtained from TRL Trainer

PR #491, opened 2024-12-20
Refactor accumulate logic

shivam15s:
## Summary

Remove redundant code by refactoring

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/chunked_loss/fused_linear_preference.py
+++ b/src/liger_kernel/chunked_loss/fused_linear_preference.py
@@ -101,47 +101,33 @@ class LigerFusedLinearPreferenceBase(torch.autograd.Function):
             """
             Fused forward and backward pass for a chunk of input and target.
             """
-            if bias is not None:
-                return torch.func.grad_and_value(
-                    compute_loss, argnums=(0, 1, 3), has_aux=True
-                )(
-                    input_chunk,
-                    weight,
-                    target_chunk,
-                    bias,
-                    ref_input_chunk=ref_input_chunk,
-                )
-            else:
-                return torch.func.grad_and_value(
-                    compute_loss, argnums=(0, 1), has_aux=True
-                )(input_chunk, weight, target_chunk, ref_input_chunk=ref_input_chunk)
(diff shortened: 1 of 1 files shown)


src/liger_kernel/chunked_loss/cpo_loss.py
```diff
@@ -26,14 +28,14 @@ def preference_loss_fn(chosen_logps, rejected_logps, full_target, beta=0.1):
         - D: Dataset of preferences
 
         Args:
-            chosen_logps (torch.Tensor): Avg log probabilities of chosen tokens. Shape: (batch_size,).
-            rejected_logps (torch.Tensor): Avg log probabilities of rejected tokens. Shape: (batch_size,).
-            full_target (torch.Tensor): Non chunked full target tensor
-            beta (float): Weight for the CPO loss
+            chosen_logps_chunk (torch.Tensor): Avg log probabilities of chosen tokens in the chunk. Shape: (batch_size,).
+            rejected_logps_chunk (torch.Tensor): Avg log probabilities of rejected tokens in the chunk. Shape: (batch_size,).
+            full_target (torch.Tensor): Non chunked full target tensor.
```
hebiao064, line 33: I wonder is it `full_target` or actually `target_chunk`?

From the fused function, we are feeding into `target_chunk`

```python
        def fused_fwd_bwd(
            input_chunk, target_chunk, ref_input_chunk, preference_labels_chunk
        ):
            """
            Fused forward and backward pass for a chunk of input and target.
            """
            if bias is not None:
                return torch.func.grad_and_value(
                    compute_loss, argnums=(0, 1, 3), has_aux=True
                )(
                    input_chunk,
                    weight,
                    target_chunk,
                    bias,
                    ref_input_chunk=ref_input_chunk,
                    preference_labels=preference_labels_chunk,
                )
```

qingquansong: Feel like it should be the full target as we use it to normalize and then sum up for all chunks but seems we're feeding the target_chunk instead?

Issue #507, opened 2024-12-31
result of LigerCrossEntropyLoss is always 0


wa008:
### 🐛 Describe the bug

As title, code as below
```
from liger_kernel.transformers.cross_entropy import LigerCrossEntropyLoss

torch.manual_seed(2024)
M = 4
N = 3
input = torch.randn(M, N, requires_grad = False, device = DEVICE)
target = torch.empty(M, dtype = torch.long, device = DEVICE).random_(N)

liger_loss = LigerCrossEntropyLoss()
output_liger = liger_loss(input, target)

torch_loss_lib = torch.nn.CrossEntropyLoss()
output_torch_lib = torch_loss_lib(input, target)

print(f'input: {input}')
print(f'target: {target}')
print(f'output_liger: {output_liger}')
print(f'output_torch_lib: {output_torch_lib}')
```
output: 
```
input: tensor([[-0.1118,  0.0162,  0.0956],
        [-0.1134,  0.0088,  0.1046],
        [-0.1737,  0.1315,  0.0422],
        [-0.0921,  0.0728,  0.0194]], device='cuda:0')
target: tensor([0, 0, 0, 0], device='cuda:0')
output_liger: 0.0
output_torch_lib: 1.2258632183074951
```

### Reproduce

I run it on colab and install it by pip
```
!pip install liger-kernel
```

### Versions

Environment Report:
-------------------
Operating System: Linux-6.1.85+-x86_64-with-glibc2.35
Python version: 3.10.12
Liger Kernel version: 0.5.2
PyTorch version: 2.5.1+cu121
CUDA version: 12.1
HIP(ROCm) version: Not available
Triton version: 3.1.0
Transformers version: 4.47.1
XPU version: XPU Not Available

DandinPower:
Hi, I think that because the Liger kernel is designed for memory optimization, the input tensor's storage space is used in-place to store the gradients during the forward computation. As a result, if you print the input before and after the `liger_loss` in your code, you'll see that the input tensor becomes the gradient values for the input. (You can verify this by setting `requires_grad=True` and printing `input.grad` after calling `liger_loss`.)

Given your test code, the output loss cannot be the same because the inputs to Liger CE and Torch CE are different.

The correct way to test this might look like the following:
```python
from liger_kernel.transformers.cross_entropy import LigerCrossEntropyLoss

torch.manual_seed(2024)
M = 4
N = 3
input = torch.randn(M, N, requires_grad=False, device=DEVICE)
[...]

PR #508, opened 2025-01-02
Set z_loss_1d=None when return_z_loss=False in cross_entropy_loss to avoid tl.store fail when triton_interpret=1(for tl.device_print etc.)

wa008:
For [issue-507](https://github.com/linkedin/Liger-Kernel/issues/507)

## Summary
In cross_entropy_loss kernel, `tl.store(loss_ptr, loss)` doesn't work when `return_z_loss=False` and `triton_interpret=1`, because loss_1d is assigned to tensor z_loss_1d, So I set `z_loss_1d = None` in this situation and it works well.

## Testing Done
I test it on my code and [this most simplified example](https://github.com/linkedin/Liger-Kernel/issues/507#issuecomment-2566829506), both work well.

Hardware Type: T4 GPU

### DIFF
--- a/src/liger_kernel/ops/cross_entropy.py
+++ b/src/liger_kernel/ops/cross_entropy.py
@@ -95,7 +95,8 @@ def liger_cross_entropy_kernel(
         return
 
     loss_ptr += program_id * loss_stride
-    z_loss_ptr += program_id * loss_stride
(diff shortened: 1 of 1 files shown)


Tcc0403 (approved):
LGTM! All tests passed.

PR #510, opened 2025-01-03
Add `average_log_prob` args for cpo

Mecoli1219:
## Summary

`trl` CPO implementation didn't average the log_probs, while the liger kernel averages it when computing the loss. This will cause a mismatch when integrating them.

## Testing Done

- [x] Updating unit test (still investigating why unit test fail locally)

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/chunked_loss/fused_linear_preference.py
+++ b/src/liger_kernel/chunked_loss/fused_linear_preference.py
@@ -32,6 +32,7 @@ class LigerFusedLinearPreferenceBase(torch.autograd.Function):
         ref_input=None,
         ref_weight=None,
         ref_bias=None,
+        average_log_prob=True,
         **loss_kwargs,
     ):
         """
@@ -61,6 +62,7 @@ class LigerFusedLinearPreferenceBase(torch.autograd.Function):
             use_ref_model (bool): Whether to use a reference model for the alignment loss.
             ref_weight (torch.Tensor): Reference weight tensor. Shape: (vocab_size, hidden_size).
             ref_bias (torch.Tensor, optional): Reference bias tensor. Shape: (vocab_size,).
+            average_log_prob (bool): Whether to average log probabilities or the sum.
             loss_kwargs (dict): Other possible arguments that a loss function might need
         """
         # TODO: Tune CHUNK_SIZE to fully utilize the GPU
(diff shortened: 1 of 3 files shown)


src/liger_kernel/chunked_loss/fused_linear_preference.py
```diff
@@ -61,6 +62,7 @@ def forward(
             use_ref_model (bool): Whether to use a reference model for the alignment loss.
             ref_weight (torch.Tensor): Reference weight tensor. Shape: (vocab_size, hidden_size).
             ref_bias (torch.Tensor, optional): Reference bias tensor. Shape: (vocab_size,).
+            average_log_prob (bool): Whether to average log probabilities or the sum.
```
kashif, line 65:
```suggestion
            average_log_prob (bool): Whether to average log probabilities or to sum them over the completion.
```

kashif:
TRL is using the default in the official repo for CPO: https://github.com/fe1ixxu/CPO_SIMPO/blob/main/scripts/cpo_trainer.py#L626

test/chunked_loss/test_cpo_loss.py
```diff
@@ -139,7 +140,7 @@ def forward(self, x, y):
 @pytest.mark.parametrize(
     "scalar, dtype, atol, rtol",
     [
-        (1.0, torch.bfloat16, 5e-3, 5e-3),
+        (1.0, torch.bfloat16, 5e-2, 5e-2),
```
austin362667, line 143: Why we need to adjust the tol?

Mecoli1219: @kashif and I find that after disabling `average_log_prob` for CPO, it will have a higher deviation from HF implementation when the model is large and the data type is `bf16`. Since the result is still close within both methods, we increase `atol` and `rtol` to make this test pass.

kashif: as bfloat16 is less accurate for larger numbers, this is needed to make the test pass and is the same as in the other bfloat16 tests

austin362667: Then adjusting tol makes sense. :heart:

kashif:
awesome thank you! we would still need a release of liger-kernel for the CI to pass but yes it will hopefully unblock!

PR #521, opened 2025-01-14
[DPO] add reference log-prob outputs in DPO

kashif:
## Summary
Since the DPO uses a reference model we also need to return the reference logprobs in DPO

## Testing Done

- Hardware Type: <BLANK>


austin362667 (dismissed):
LGTM! Thanks

Tcc0403 (approved):
LGTM

PR #523, opened 2025-01-15
Fix Unit Test error brought by Transformer Breaking Changes

hebiao064:
## Summary

There is a breaking changes from HF Transformers that will break out unit test:

https://github.com/huggingface/transformers/blame/137965ca7d0b453b22806c0a5fffc51fde821c33/src/transformers/models/llama/modeling_llama.py#L85

For example, it broke my [PR about KTO Loss](https://github.com/linkedin/Liger-Kernel/pull/475) which has nothing to do with `FAILED test/transformers/test_rope.py::test_correctness`

https://github.com/linkedin/Liger-Kernel/actions/runs/12779518445/job/35624270834?pr=475

## Testing Done

Unit Test is done

- Hardware Type: <BLANK>

### DIFF
--- a/benchmark/scripts/benchmark_rope.py
+++ b/benchmark/scripts/benchmark_rope.py
@@ -32,7 +32,6 @@ def bench_speed_rope(input: SingleBenchmarkRunInput) -> SingleBenchmarkRunOutput
     seq_len = extra_benchmark_config["seq_len"] if "seq_len" in extra_benchmark_config else input.x
 
     head_dim = hidden_size // num_q_heads
-    llama_config = LlamaConfig(head_dim=head_dim)
 
     if transformers_version < "4.48.0":
         # LlamaRotaryEmbedding constructor signature changed in transformers 4.48.0
--- a/benchmark/scripts/benchmark_rope.py
+++ b/benchmark/scripts/benchmark_rope.py
@@ -35,7 +35,7 @@ def bench_speed_rope(input: SingleBenchmarkRunInput) -> SingleBenchmarkRunOutput
 
     if transformers_version < "4.48.0":
         # LlamaRotaryEmbedding constructor signature changed in transformers 4.48.0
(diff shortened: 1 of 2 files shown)


ByronHsu (dismissed):

ByronHsu:
@hebiao064 can you make a condition for tests-bwd to pass with older transformers version? Some users are on older transformers version

Something like

```
if transformer_version < ...:
  ...
else:
  ...
```

hebiao064:
> * test-convergence

Make sense, added

hebiao064:
> @hebiao064 can you make a condition for tests-bwd to pass with older transformers version? Some users are on older transformers version
> 
> Something like
> 
> ```
> if transformer_version < ...:
>   ...
> else:
>   ...
> ```



Make sense, added

benchmark/scripts/benchmark_rope.py
```diff
@@ -105,7 +113,13 @@ def bench_memory_rope(input: SingleBenchmarkRunInput) -> SingleBenchmarkRunOutpu
...
 
     head_dim = hidden_size // num_q_heads
-    rotary_emb = LlamaRotaryEmbedding(head_dim, device=device)
+
+    if transformers_version < "4.48.0":
+        # LlamaRotaryEmbedding constructor signature changed in transformers 4.48.0
+        rotary_emb = LlamaRotaryEmbedding(head_dim, device=device)
+    else:
+        llama_config = LlamaConfig(head_dim=head_dim)
+        rotary_emb = LlamaRotaryEmbedding(llama_config, device=device)
```
kvignesh1420, line 122: seems like we are repeating same code in all places. Shall we prepare a common function to get the rotary embeddings?

austin362667: hey @kvignesh1420  thanks for the great advice, I added a more general helper function here: https://github.com/linkedin/Liger-Kernel/pull/526.

PR #535, opened 2025-01-22
[tests] use a valid hexadecimal string instead of a placeholder

faaany:
## Summary
In the latest code of `get_cache_manager` of the triton library (https://github.com/triton-lang/triton/blob/1bb8b8055c81f6bb85055645a20e0dbd27d5295f/python/triton/runtime/cache.py#L261), the key needs to be a valid hexadecimal string, otherwise the test will fail. 

This PR fixes the issue by using a valid hex string instead of a placeholder.

## Testing Done
The test passes both on CUDA and XPU.
```bash
========================================================== PASSES ===========================================================
================================================== short test summary info ==================================================
PASSED test/triton/test_triton_monkey_patch.py::test_import_custom_cache_manager
=============================================== 1 passed, 1 warning in 0.12s ================================================
```

On  Intel(R) Data Center GPU Max 1550 and NV GPU A100

- Hardware Type: <BLANK>


faaany:
@sywangyi @yao-matrix

ByronHsu (approved):
Thanks!

PR #549, opened 2025-01-30
[Fix] Fix the type hint of `test_utils::concatenated_forward`

hongpeng-guo:
## Summary

Fix the type hint of two func args from `Tensor | None=None` to `Optional[Tensor]=None`, the former typing hint is only working for python 3.10+, switching it back to `Optional` to fit for 3.9 and below python versions.

## Testing Done

- Hardware Type: <BLANK>


austin362667 (approved):
Thank you!

PR #552, opened 2025-01-31
Monkeypatch for Qwen2.5-VL

BenasdTW:
## Summary

[Qwen2.5-VL](https://huggingface.co/Qwen/Qwen2.5-VL-7B-Instruct) was released a few days ago. This PR aims to add support for it.
There are some small changes in its architecture compared to Qwen2-VL.

## Details
As of February 1, the latest git version of `transformers` ([62db3e6](https://github.com/huggingface/transformers/commit/62db3e6ed67a74cc1ed1436acd9973915c0a4475)) supports it, but it's not yet in the latest release of `transformers`.

Needs help with the tests, as I got OOMs on my laptop GPU. I have tried LoRA with PR, which works great.

## Testing Done

Can't run the tests on my hardware, got OOMs.

- Hardware Type: RTX 4070 laptop
- [ ] run `make test` to ensure correctness
- [ ] run `make checkstyle` to ensure code style
- [ ] run `make test-convergence` to ensure convergence

PR #563, opened 2025-02-12
adjust atol value for `mini_qwen2_vl` FP32 in `test_mini_model_multimodal`

faaany:
## Summary
`test_mini_model_multimodal` has 2 test models: mini_mllama and mini_qwen2_vl. On XPU, mini_mllama cases pass but mini_qwen2_vl cases fail. In the 2 failed cases: one fails at 1e-2 decimal point, which we need to fix; but the other fails at 1e-6 decimal point, which we suspect that the atol is set too strict for non-cuda hardware accelerators as can be seen below:

```bash
FAILED test_mini_models_multimodal.py::test_mini_model_multimodal[mini_qwen2_vl-32-0.0001-dtype0-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: Number of mismatched elements: 3
Mismatch at index (0, 21): tensor1[(0, 21)] = 0.03207247704267502, tensor2[(0, 21)] = 0.0320717953145504
Mismatch at index (0, 24): tensor1[(0, 24)] = 0.03104054555296898, tensor2[(0, 24)] = 0.03104141354560852
Mismatch at index (0, 27): tensor1[(0, 27)] = 0.04820457845926285, tensor2[(0, 27)] = 0.04820366948843002
```

This PR tries to fix this. What do you guys think? 

- Hardware Type: <BLANK>

### DIFF
--- a/test/convergence/test_mini_models_multimodal.py
+++ b/test/convergence/test_mini_models_multimodal.py
@@ -340,7 +340,7 @@ def run_mini_model_multimodal(
             32,
             1e-4,
             torch.float32,
-            1e-8,
+            1e-6,
             1e-5,
             5e-3,
             1e-5,


faaany:
cc @yao-matrix and @sywangyi

tyler-romero:
This LGTM but I'm going to pass to someone at LI to provide a final review: @shivam15s

faaany:
close this PR, as it got fixed in  #552

PR #571, opened 2025-02-20
fix qwen2vl and mllama test to pass failing tests

shivam15s:
## Summary

1. Fix Qwen2VL mrope tests. Transformers removed BC with Qwen2VLRotaryEmbedding which is why we need this change
2. Fix mllama tests to not use cross entropy. Using Cross Entropy causes logits mismatch which is expected as we do inplace operation in cross entropy kernel.
3. Runs make checkstyle --- there seems to be a bug in checkstyle workflow where unformatted checkstyle seems to pass tests. Changed the makefile to throw error in such cases.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,10 @@ test:
 
 # Command to run ruff for linting and formatting code
 checkstyle:
-	ruff check . --fix; ruff_check_status=$$?; \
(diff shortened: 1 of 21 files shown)


momochen (approved):
🚢

PR #587, opened 2025-02-25
Infer correct device for AMD HIP device

helloworld1:
## Summary

In AMD HIP, it is using CUDA as device name instead of "hip"

## Testing Done

build and test on AMD device

- Hardware Type: <BLANK>


qgallouedec:
Do you plan making a patch release?

PR #591, opened 2025-03-01
[KTO] Only return summed metrics

vaibhavjindal:
## Summary

This change modifies the code to only return summed metrics for chosen_logps, chosen_rewards etc. This will help in reducing unnecessary overhead is returning entire vectors.

## Testing Done

- Hardware Type: <BLANK>


shivam15s (approved):

hebiao064:
wonder what's the benchmark after this change? is it on faster then HF now?

vaibhavjindal:
> wonder what's the benchmark after this change? is it on faster then HF now?

No, it is still not faster than HF. The speed issue is still to be solved.

Here are the plots:
kto_loss_memory
kto_loss_speed

PR #608, opened 2025-03-14
Paligemma support

eljandoubi:
## Summary
This PR adds Paligemma support to the Liger-Kernel.

## Testing Done

- Hardware Type: RTX 3070
- [x] run `make test` to ensure correctness
- [x] run `make checkstyle` to ensure code style
- [x] run `make test-convergence` to ensure convergence

PR #613, opened 2025-03-16
language model of paligemma 1 is gemma 1.

eljandoubi:
## Summary
@lancerts we need to match the patch of gemma version to paligemma version

### DIFF
--- a/src/liger_kernel/transformers/model/paligemma.py
+++ b/src/liger_kernel/transformers/model/paligemma.py
@@ -21,6 +21,190 @@ from liger_kernel.transformers.fused_linear_cross_entropy import LigerFusedLinea
 logger = logging.get_logger(__name__)
 
 
+@add_start_docstrings_to_model_forward(PALIGEMMA_INPUTS_DOCSTRING)
+@replace_return_docstrings(output_type=PaliGemmaCausalLMOutputWithPast, config_class=_CONFIG_FOR_DOC)
+def lce_forward_deprecated(
+    self,
+    input_ids: torch.LongTensor = None,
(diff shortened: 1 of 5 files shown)


shivam15s:
Does this fix the failing bwd compatibility test on the main branch?

eljandoubi:
@shivam15s can you provide more information about the bwd issue?

shivam15s:
This test failed after PR #608 : https://github.com/linkedin/Liger-Kernel/actions/runs/13875264364/job/38826758036
More information about the bwd test is here https://github.com/linkedin/Liger-Kernel/blob/main/dev/modal/tests_bwd.py (it tests backward compatibility with transformers==4.44.2)

eljandoubi:
@shivam15s Even the deprecated `forward` function from `transformers==4.44.2` is causing the same CUDA error, so I have skipped that test.

shivam15s:
Could you explain the error reason and why is it expected?

RichhLi:
This seems to be the problem? https://github.com/linkedin/Liger-Kernel/actions/runs/13935968640/job/39003807039#step:5:1623

eljandoubi:
@shivam15s @RichhLi 
image
image
In `transformers==4.44.2`, the issue originates from the `masked_scatter` operation on `final_embeddings`. A more explicit error message appears in `transformers==4.45.2`, stating that the number of images does not match the number of special image tokens in the input text. In other words, the way the model is initialized from the configuration has changed.

shivam15s:
@eljandoubi Thanks for the explanation. Skipping the test is fine for now

shivam15s (approved):
lgtm

Issue #626, opened 2025-03-24
Generalized PPO loss


qingquansong:
### 🚀 The feature, motivation and pitch

- [ ] 1. Current GRPO assume KL term is added and advantage is computed in side the loss, we wanna open this to become configurable from user
- [ ] 2. Clipping is not used in GRPO and need to be added (also as an option for generalized PPO case)
- [ ] 3. Reference model logits/prob can be provided directly without providing LM head + hidden states as an option
- [ ] 4. Old policy prob or LM head/hidden states is not provided for importance sampling purpose in GRPO.

qingquansong:
Assigned to me now and if anyone is interested in anyone of the features, feel free to take.

kashif:
yes the implementation in liger is older so would need updating to the newer version in TRL

shivam15s:
worked on #628 ... didn't see this issue before so was working on this independently 
 1. KL is only calculated if beta is non-zero. Assume user directly provides the advantage instead of reward (otherwise we need all gather to compute reward mean/std)
 2. Added clipping
 3. Not supported yet 
 4. added an option for old policy prob
Also added fix for loss correctness, metrics calculation, testing, etc.

PR #628, opened 2025-03-25
Fix GRPO to conform with TRL: Fix loss, make tests accurate, correct metrics computation

shivam15s:
## Summary

Tries to address #626 and other correctness improvements

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/chunked_loss/fused_linear_rlhf.py
+++ b/src/liger_kernel/chunked_loss/fused_linear_rlhf.py
@@ -80,7 +80,14 @@ class LigerFusedLinearRLHFBase(torch.autograd.Function):
             rlhf_loss_fn=cls.rlhf_loss_fn,
         )
 
-        def fused_fwd_bwd(input_chunk, selected_token_ids_chunk, attention_mask_chunk, advantages_chunk, ref_input_chunk, old_per_token_logps_chunk):
+        def fused_fwd_bwd(
+            input_chunk,
+            selected_token_ids_chunk,
+            attention_mask_chunk,
+            advantages_chunk,
+            ref_input_chunk,
+            old_per_token_logps_chunk,
+        ):
             """Fused forward and backward for a chunk."""
             argnums = (0, 1, 5) if bias is not None else (0, 1)
             return torch.func.grad_and_value(compute_loss, argnums=argnums, has_aux=True)(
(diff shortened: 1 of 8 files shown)


qingquansong:
One more thing I forget to mention in the issue. The aggregation way is also implemented differently as the original GRPO paper equation. It should be a per sequence/response avg and then per group avg. Here we directly aggregate the entire batch loss and average based on sum of tokens. Ideally it should be a three level thing (or maybe two level depending on the input) `batch avg <- group avg <- seq/response avg`

src/liger_kernel/chunked_loss/grpo_loss.py
```diff
@@ -7,60 +7,60 @@ class LigerFusedLinearGRPOFunction(LigerFusedLinearRLHFBase):
...
-
-        # Combine losses
-        per_token_loss = policy_loss + beta * kl_div
-
-        # Apply masking and normalize
-        masked_loss = per_token_loss * attention_mask
-        seq_lengths = attention_mask.sum()
-        seq_lengths = torch.clamp(seq_lengths, min=1.0)
-        loss = masked_loss.sum() / seq_lengths
+        old_per_token_logps = old_per_token_logps if old_per_token_logps is not None else per_token_logps.detach()
```
qingquansong, line 38: It should be great if we can create two small utility functions for the kl compute part and also the advantage clip part and can be reused and switched by different losses. For example, GRPO use k3 approximation here for the kl penalty (and you can use k3 to represent the function name) and we can switch to others if needed. Thank you1

qingquansong (dismissed):
LGTM! Just two minor small comments!

src/liger_kernel/chunked_loss/fused_linear_rlhf.py
```diff
@@ -158,60 +229,74 @@ def accumulate_chunk(input_chunk, attention_mask_chunk, rewards_chunk, ref_input
...
+            logits = logits / temperature
 
         # Compute log probabilities using softmax over the last dimension
         log_probs = F.log_softmax(logits.float(), dim=-1)
 
-        # Monitoring: compute mean of logits
-        batch_size, seq_len, _ = input_chunk.shape
-        logits_mean = logits.sum() / (batch_size * seq_len * weight.shape[0])
-        return log_probs, logits, logits_mean
+        return log_probs, logits
```
lancerts, line 299: So this is a backward-breaking change? What was the motivation?

shivam15s: logits_mean isn't needed for any metric logging as far as trl is concerned

src/liger_kernel/chunked_loss/grpo_loss.py
```diff
@@ -7,60 +7,60 @@ class LigerFusedLinearGRPOFunction(LigerFusedLinearRLHFBase):
...
         # Calculate metrics
-        metrics = (
-            chosen_token_logprobs.mean(),  # mean log prob
-            chosen_token_logprobs.std(),  # std log prob
-            log_probs.mean(),  # mean all log probs
-            ((kl_div * attention_mask).sum(dim=1) / attention_mask.sum(dim=1)).mean(),  # mean KL div
+        metrics = []
+        if beta != 0.0:
+            metrics.append(((kl_div * attention_mask).sum() / torch.clamp(full_attention_mask.sum(), min=1.0)))
+        is_clipped = ((coef_1 < 1 - epsilon_low) & (advantages.unsqueeze(1) < 0)) | (
```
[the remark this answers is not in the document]
shivam15s: I feel this is because of how we compute per_token_loss:
`per_token_loss = -torch.min(per_token_loss1, per_token_loss2)`
Assume, `coef_1 = 1-2*eps; and coef_2 = 1 - eps;` 
if advantage is `+ve`, `torch.min()` would select `per_token_loss_1` rather than `per_token_loss2` as `per_token_loss1` < `per_token_loss2`.
So, even though `coef_2` is clipped, we still took `per_token_loss_1` as loss.

tldr: sign of advantage is important to figure out is_clipped

qingquansong (approved):
LGTM! :shipit:

qingquansong (approved):

PR #634, opened 2025-03-27
[chunked loss] align teacher and student logit shape

yundai424:
## Summary

In rare cases where the teacher and student models don't have the same vocab size (but their vocabs are actually the same), for example qwen models, we pad students to match the teacher's logit.

## Testing Done

make test

- Hardware Type: <BLANK>


shivam15s (approved):
lgtm

PR #635, opened 2025-03-28
Zhasun/grpo add ref logits

mRSun15:
## Summary

Add the ref_log_probs as input. and if the log probs is not none, we don't need to re-calculate it using ref_input.

Add it in corresponding unit test as well.

## Testing Done

- Hardware Type: <BLANK>


mRSun15:
@shivam15s  please take a look.

PR #645, opened 2025-04-02
add xpu tuning to CE

mgrabban:
## Summary
Tuning on XPU: In cross-entropy, if device is `xpu`, set `MAX_FUSED_SIZE` to `4096` instead of default `65536 // 2`. This gives slightly better performance on xpu.

## Testing Done

- Hardware Type: Intel(R) Data Center GPU Max 1550


mgrabban:
Very similar to #647

Issue #569, opened 2025-02-19
`transformers` is meant to be an optional dep but import fails without it


tyler-romero:
### 🐛 Describe the bug

Attempting to run `from liger_kernel.transformers.fused_linear_cross_entropy import LigerFusedLinearCrossEntropyLoss` fails when transformers is not installed. However transformers is an optional dependency.

Many modules in Liger-Kernel are useful outside of the context of transformers (such as `LigerFusedLinearCrossEntropyLoss`) and so it would be nice to avoid the dependency when it is not needed (even in `liger_kernel.transformers` package).

```
Traceback (most recent call last):
  File "...", line 20, in <module>
    from liger_kernel.transformers.fused_linear_cross_entropy import LigerFusedLinearCrossEntropyLoss
  File "/home/tromero/workspace/nanogpt-speedrun/.venv/lib/python3.12/site-packages/liger_kernel/transformers/__init__.py", line 1, in <module>
    from liger_kernel.transformers.auto_model import AutoLigerKernelForCausalLM  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tromero/workspace/.../.venv/lib/python3.12/site-packages/liger_kernel/transformers/auto_model.py", line 3, in <module>
    from transformers import AutoConfig
ModuleNotFoundError: No module named 'transformers'
Traceback (most recent call last):
  File "/home/tromero/workspace/...", line 20, in <module>
    from liger_kernel.transformers.fused_linear_cross_entropy import LigerFusedLinearCrossEntropyLoss
  File "/home/tromero/workspace/.../.venv/lib/python3.12/site-packages/liger_kernel/transformers/__init__.py", line 1, in <module>
    from liger_kernel.transformers.auto_model import AutoLigerKernelForCausalLM  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tromero/workspace/.../.venv/lib/python3.12/site-packages/liger_kernel/transformers/auto_model.py", line 3, in <module>
    from transformers import AutoConfig
ModuleNotFoundError: No module named 'transformers'
```

### Versions

Environment Report:
-------------------
[...]

PR #659, opened 2025-04-08
Fix imports without transformers

vaibhavjindal:
## Summary

Fixes https://github.com/linkedin/Liger-Kernel/issues/569.

## Details
This pull request addresses import errors that occur when the transformers library is not installed. By implementing conditional imports and handling missing dependencies gracefully, the codebase now functions correctly even without transformers. These changes don't modify the existing API at all.

Now, the following commands will run gracefully even when transformers is not installed:
`from liger_kernel.transformers.fused_linear_cross_entropy import LigerFusedLinearCrossEntropyLoss`
and
`import liger_kernel.transformers`

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/transformers/__init__.py
+++ b/src/liger_kernel/transformers/__init__.py
@@ -1,4 +1,22 @@
-from liger_kernel.transformers.auto_model import AutoLigerKernelForCausalLM  # noqa: F401
+import importlib
+
+# Check if 'transformers' is installed
+try:
+    import transformers  # noqa: F401
+    _TRANSFORMERS_AVAILABLE = True
+except ImportError:
+    _TRANSFORMERS_AVAILABLE = False
+
+
+def is_transformers_available() -> bool:
+    """
+    Returns True if the 'transformers' package is available.
+    Useful for conditional logic in downstream code.
+    """
+    return _TRANSFORMERS_AVAILABLE
+
+
+# Always-safe imports (independent of 'transformers')
 from liger_kernel.transformers.cross_entropy import LigerCrossEntropyLoss  # noqa: F401
 from liger_kernel.transformers.dyt import LigerDyT  # noqa: F401
 from liger_kernel.transformers.fused_linear_cross_entropy import LigerFusedLinearCrossEntropyLoss  # noqa: F401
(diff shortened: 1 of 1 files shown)


vaibhavjindal:
https://github.com/linkedin/Liger-Kernel/issues/569

src/liger_kernel/transformers/__init__.py
```diff
@@ -1,32 +1,120 @@
...
+        "apply_liger_kernel_to_paligemma",
+        "apply_liger_kernel_to_phi3",
+        "apply_liger_kernel_to_qwen2",
+        "apply_liger_kernel_to_qwen2_5_vl",
+        "apply_liger_kernel_to_qwen2_vl",
+    }
+
+    if name in monkey_patch_symbols:
+        module = importlib.import_module("liger_kernel.transformers.monkey_patch")
+        return getattr(module, name)
```
shivam15s, line 73: Curious, won't line 97 already handle these imports?

yundai424 (approved):
LGTM. just out of curiosity, will indexing still work in this case? 🤔

src/liger_kernel/transformers/__init__.py
vaibhavjindal: I think just having line 97 would be insufficient. For example, before this change, when transformers is installed, something like `from liger_kernel.transformers import AutoLigerKernelForCausalLM` will work. But, if we remove the __getattr__ function and just keep line 97, this import will fail. Is this what you meant?

vaibhavjindal:
> LGTM. just out of curiosity, will indexing still work in this case? 🤔

What exactly do you mean by indexing here? Do you mean that text editors etc might not be able to see all the imports?

src/liger_kernel/transformers/__init__.py
shivam15s: My bad I misunderstood what the code is doing

shivam15s:
Yes, I think @yundai424 meant will IDE features like autocomplete/linting/"Go to Definition" work with this?

vaibhavjindal:
> Yes, I think @yundai424 meant will IDE features like autocomplete/linting/"Go to Definition" work with this?

Ah okay, such features might not work. My VSCode is able to show suggestions for non transformer dependent imports, but not for the ones that depend on transformers. However, copilot is able to give correct suggestions even for transformer dependent imports. 

I think we can do something like https://github.com/huggingface/transformers/blob/08e3217bafddc5d11ce0e7369bcfaaabe5501ba5/src/transformers/__init__.py#L5290 to help us deal with this. Let me do it in a separate PR!

PR #661, opened 2025-04-10
 Modified block and warp sizes for improved performance on XPU for both layernnorm and rmsnorm

Tarakarevu1:
## Summary

This change is related to performance tuning on the Intel Max 1550 GPUs. By keeping the block and warp sizes the same in the forward and backward Triton kernels.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/ops/layer_norm.py
+++ b/src/liger_kernel/ops/layer_norm.py
@@ -197,11 +197,22 @@ def layer_norm_backward(dY, X, W, B, Mean, RSTD):
     _DW = torch.empty((sm_count, n_cols), dtype=W.dtype, device=W.device)
     _DB = torch.empty((sm_count, n_cols), dtype=W.dtype, device=W.device)
 
-    BLOCK_SIZE, num_warps = calculate_settings(n_cols)
-    if n_cols > BLOCK_SIZE:
-        raise RuntimeError(
-            f"Feature dimension {n_cols} exceeds maximum supported size of {BLOCK_SIZE}. Consider using a smaller feature dimension."
-        )
+    if X.device.type == "xpu":  # XPU-specific optimization
+        BLOCK_SIZE = torch.xpu.get_device_properties(X.device).max_work_group_size
+        num_warps = 4
+    else:
+        BLOCK_SIZE, num_warps = calculate_settings(n_cols)
+
+    if X.device.type == "xpu":  # XPU-specific optimization
+        if n_cols > 65536:
+            raise RuntimeError(
+                f"Feature dimension {n_cols} exceeds maximum supported size of {BLOCK_SIZE}. Consider using a smaller feature dimension."
+            )
+    else:
+        if n_cols > BLOCK_SIZE:
+            raise RuntimeError(
+                f"Feature dimension {n_cols} exceeds maximum supported size of {BLOCK_SIZE}. Consider using a smaller feature dimension."
+            )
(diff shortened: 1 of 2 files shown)


src/liger_kernel/ops/layer_norm.py
```diff
@@ -197,11 +197,21 @@ def layer_norm_backward(dY, X, W, B, Mean, RSTD):
     _DW = torch.empty((sm_count, n_cols), dtype=W.dtype, device=W.device)
     _DB = torch.empty((sm_count, n_cols), dtype=W.dtype, device=W.device)
 
-    BLOCK_SIZE, num_warps = calculate_settings(n_cols)
-    if n_cols > BLOCK_SIZE:
-        raise RuntimeError(
-            f"Feature dimension {n_cols} exceeds maximum supported size of {BLOCK_SIZE}. Consider using a smaller feature dimension."
-        )
+    if X.device.type == "xpu":  # XPU-specific optimization
+        BLOCK_SIZE = torch.xpu.get_device_properties(X.device).max_work_group_size
```
shivam15s, line 201: afaik we want the block size here to be `triton.next_power_of_2(n)`. `torch.xpu.get_device_properties(X.device).max_work_group_size` is supposedly giving the max possible block size on the xpu device. Better way to handle this is to change the calculate_settings function directly where you set the `MAX_FUSED_SIZE` according to the device.

src/liger_kernel/ops/rms_norm.py
```diff
@@ -212,7 +212,11 @@ def rms_norm_forward(X, W, eps, offset, casting_mode):
     dim = shape[-1]
     X = X.view(-1, dim)
     n_rows, n_cols = X.shape
-    BLOCK_SIZE, num_warps = calculate_settings(n_cols)
+    if X.device.type == "xpu":  # XPU-specific optimization
+        BLOCK_SIZE = torch.xpu.get_device_properties(X.device).max_work_group_size
```
shivam15s, line 216: same as above

src/liger_kernel/ops/layer_norm.py
Tarakarevu1: @shivam15s These changes are specific for the layernorm and rmsnorm.That's why i made the changes locally and this will impact other kernels.

src/liger_kernel/ops/rms_norm.py
Tarakarevu1: same as above

src/liger_kernel/ops/layer_norm.py
lancerts: @Tarakarevu1 how much is the perf improvment of layernorm and rmsnorm after this change?

src/liger_kernel/ops/layer_norm.py
```diff
@@ -197,11 +197,21 @@ def layer_norm_backward(dY, X, W, B, Mean, RSTD):
...
+    else:
+        BLOCK_SIZE, num_warps = calculate_settings(n_cols)
+
+    if X.device.type == "xpu":  # XPU-specific optimization
```
abhilash1910, line 205: @Tarakarevu1  I think this is not an optimization but rather a guardrail to avoid spilling beyond 64 KB ?
LGTM on the code changes although the runtime behaviour error can be tweaked to show more xpu usage ?

PR #668, opened 2025-04-15
update setup.py for installation on xpu

faaany:
## Summary
We want to remove `pytorch-triton-xpu` from the setup.py file, because it gets installed automatically with torch>=2.6.0 when using the below installation:
```bash
pip install -e .[dev] --extra-index-url https://download.pytorch.org/whl/test/xpu
```
So no need to explicitly mention it there.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,6 @@ def get_default_dependencies():
     elif platform == "xpu":
         return [
             "torch>=2.6.0",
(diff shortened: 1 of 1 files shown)


faaany:
cc @hebiao064 @ByronHsu

shivam15s (approved):

PR #669, opened 2025-04-16
update XPU CI yaml file to use docker container

faaany:
## Summary
This PR does 2 things
- skip 2 tests that fails on XPU
- update XPU yaml file to use docker container

- Hardware Type: Intel GPU


.github/workflows/intel-ci.yml
```diff
@@ -45,18 +45,20 @@ jobs:
       run: make checkstyle
 
   tests:
-    runs-on: linux-max1550-gpu-8
+    runs-on: linux-max1550-pvc-8
     needs: [checkstyle]
-
+    container:
+      image: intel/oneapi-basekit:2025.0.1-0-devel-ubuntu22.04
```
yao-matrix, line 51: should we use 24.04 image, so we have python 3.12 here, to increase inter-operability w/ other ecosystem libraries like vllm (which requires > 3.11)

faaany: okay, I updated the docker image to use ubuntu24.04.

shivam15s (approved):
lgtm

PR #680, opened 2025-04-24
Fix LigerCrossEntropy reduction='none'

### LINKED ISSUE #678: LigerCrossEntropyLoss(reduction='none') cause grad_norm=0
### 🐛 Describe the bug

`LigerCrossEntropyLoss(reduction='none')` or `LigerFusedLinearCrossEntropyLoss(reduction='none')` makes grad_norm to zero.

I'm not sure but I think `labels` contains lots of `IGNORE_INDEX`, makes grad_norm to zero.
when `reduction='mean'`, It works fine

### Reproduce

reproduction code
```python
    # override compute_loss from huggingface.Trainer
    # model = LlamaForCausalLM
    def __init__(self, args, **kwargs):
        ....
        self.loss_fct = LigerCrossEntropyLoss(reduction='none')
        ....
    
    @override
    def compute_loss(self, model, inputs, return_outputs=False, **kwargs):       
        loss = None
        labels = inputs.pop("labels") if "labels" in inputs else None
        inputs.update({"return_dict": True})
        outputs = model(**inputs)
        logits = outputs.logits

        if labels is not None:
            # Upcast to float if we need to compute the loss to avoid potential precision issues
            logits = logits.float()
            # Shift so that tokens < n predict n
            shift_logits = logits[..., :-1, :].contiguous()
            shift_labels = labels[..., 1:].contiguous()
            # Flatten the tokens
            shift_logits = shift_logits.view(-1, self.model.config.vocab_size)
            shift_labels = shift_labels.view(-1)
            # Enable model parallelism
            shift_labels = shift_labels.to(shift_logits.device)
            ce_loss = self.loss_fct(shift_logits, shift_labels)
            mask = (shift_labels != -100)
            loss = (ce_loss * mask).sum() / mask.sum()

        return (loss, outputs) if return_outputs else loss
```

### Versions

0.5.5

Tcc0403:
Resolves #678.

Current FLCE doesn't calcualte the gradients correctly when reduction='none'. In this scenario, `grad_output` is not just a scalar and it should be multiplied to each row respectively isntead.

## Testing Done

- Hardware Type: <BLANK>


shivam15s (approved):
lgtm

PR #689, opened 2025-04-28
Multi-token Attention

AndreSlavescu:
## Summary

Paper: https://arxiv.org/pdf/2504.00927

## Testing Done

tested fp32 + bf16

- Hardware Type: 3090


test/transformers/test_softmax.py
```diff
@@ -0,0 +1,82 @@
...
+device = infer_device()
+
+
+@pytest.mark.parametrize(
+    "shape",
+    [
+        (2, 8),
+        (4, 16),
+        (1, 1023),  # Large single row
+        (3, 7, 256),  # 3D input
```
Tcc0403, line 17: Let's add a large dimension case to test the multiblock version of softmax

test/transformers/test_multi_token_attention.py
```diff
@@ -0,0 +1,323 @@
...
+    "dtype, atol, rtol",
+    [
+        (torch.float32, 1e-4, 1e-4),
+        pytest.param(
+            torch.bfloat16,
+            2e-2,
+            2e-2,
+            marks=pytest.mark.skipif(
+                not (torch.cuda.is_available() and torch.cuda.is_bf16_supported()),
+                reason="bfloat16 not supported on this device",
```
Tcc0403, line 59: There's a `support_bfloat16()` in test/utils that covers xpu bf16 support.

test/transformers/test_softmax.py
```diff
@@ -0,0 +1,82 @@
...
+    ],
+)
+@pytest.mark.parametrize(
+    "dtype, atol, rtol",
+    [
+        (torch.float32, 1e-5, 1e-5),
+        (torch.bfloat16, 5e-2, 5e-2)
+        if torch.cuda.is_available() and torch.cuda.get_device_capability(0)[0] >= 8
+        else pytest.param(torch.bfloat16, 0, 0, marks=pytest.mark.skip(reason="bfloat16 not supported")),
```
Tcc0403, line 26: ditto. Also, is there any specific reason not using pytest.mark.skipif as other tests do?

src/liger_kernel/ops/multi_token_attention.py
```diff
@@ -0,0 +1,251 @@
...
+        scores_inf = _mask_inf_forward(scores)
+
+        out_flat_sparse = None
+        activation_output = None
+
+        ctx.sparse = sparse
+
+        if sparse:
+            if scores_inf.dtype != torch.float32:
+                raise RuntimeError("Liger sparse multi-token attention currently only supports fp32 input scores")
```
Tcc0403, line 175: Would you mind explaining why this contraint exists?

src/liger_kernel/ops/multi_token_attention.py
```diff
@@ -0,0 +1,251 @@
...
+
+from liger_kernel.ops.softmax import _softmax_forward
+from liger_kernel.ops.sparsemax import _sparsemax_backward
+from liger_kernel.ops.sparsemax import _sparsemax_forward
+from liger_kernel.ops.utils import calculate_settings
+from liger_kernel.ops.utils import ensure_contiguous
+
+
+@triton.jit
+def _mask_inf_fwd_kernel(
```
Tcc0403, line 16: Just to confirm, the difference between _mask_inf_fwd_kernel and _mask_zero_fwd_kernel is just the filled value, and the bwd_kernel of them is identical. Is my undersetanding correct? If so, is it possible to just make a mask kernel that accecpt an additional argument to decide what value to fill?

src/liger_kernel/ops/multi_token_attention.py
```diff
@@ -0,0 +1,251 @@
...
+
+    base = scores_ptr + batch_id * stride_b
+    offs = row_idx[:, None] * stride_m + col_idx[None, :] * stride_n
+
+    vals = tl.load(base + offs, mask=in_bounds, other=0.0, cache_modifier=".ca")
+
+    future = col_idx[None, :] > row_idx[:, None]
+
+    neg_inf = tl.full(vals.shape, -1e9, dtype=vals.dtype)
+    out = tl.where(future, neg_inf, vals)
```
Tcc0403, line 36: Just a thought, I didn't try it out myself. Is it equivalent to passing `mask=in_bound & future` and `other=-1e9` to `tl.load()`

benchmark/benchmarks_visualizer.py
```diff
@@ -103,22 +103,37 @@ def plot_data(df: pd.DataFrame, config: VisualizationsConfig):
...
+    except Exception:
+        ax = sns.lineplot(
+            data=df,
+            x="x_value",
+            y="y_value_50",
+            hue="kernel_provider",
+            marker="o",
+            palette="tab10",
+            errorbar=None,
+        )
```
Tcc0403, line 136: See #715. We need to add a filtering logic to extra benchmark config.

benchmark/scripts/benchmark_multi_token_attention.py
```diff
@@ -0,0 +1,218 @@
...
+from utils import parse_benchmark_script_args
+from utils import run_benchmarks
+
+from liger_kernel.transformers.multi_token_attention import LigerMultiTokenAttention
+from liger_kernel.utils import infer_device
+
+device = infer_device()
+
+
+class TorchMultiTokenAttention(torch.nn.Module):
```
Tcc0403, line 17: Let's import torch implementation from test to keep it easier to maintain.

src/liger_kernel/ops/multi_token_attention.py
AndreSlavescu: torch.sort does not support CUDA backend currently. So I think it makes sense to constrain this first, and then cast to bf16 after calling the sparsemax API if one were to want bf16 representation.

src/liger_kernel/ops/multi_token_attention.py
AndreSlavescu: Yes, I should be able to fold the future vals into the mask. Although, I think the triton compiler will do this anyway.

benchmark/scripts/benchmark_multi_token_attention.py
AndreSlavescu: When running the benchmark script from benchmark dir, it's not able to find the test dir.

benchmark/benchmarks_visualizer.py
AndreSlavescu: added some more filtering logic

lancerts:
There are dozen of test failing..

AndreSlavescu:
> There are dozen of test failing..

Seems like #714 is experiencing same errors for AMD GPU tests. Maybe a faulty commit earlier in master is causing these errors

Tcc0403:
It's just transformers change #723

benchmark/benchmarks_visualizer.py
Tcc0403: Thanks!

src/liger_kernel/ops/multi_token_attention.py
Tcc0403: > Although, I think the triton compiler will do this anyway.

Yeah, the compiler should do it itself. Even if it doesn't, I can expect the improvement would be little to none. I'm just trying to avoid `tl.where`

AndreSlavescu: Made those changes with simplification 👍

benchmark/scripts/benchmark_multi_token_attention.py
Tcc0403: I see. Just keep it this way for now and I'll see how to handle it in the future.

AndreSlavescu: My original idea was to have some sort of global utils, which implements a bunch of pytorch layers as baseline units, but that will just add another layer of complexity. Although it should keep everything in one place.

Tcc0403:
All relevant unit tests (multi_token_attention, softmax and sprasemax) passed on my device.

PR #693, opened 2025-05-02
[dpo] set default average_log_prob to False

cyr0930:
## Summary
Match default with this (https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/chunked_loss/dpo_loss.py#L71)
Not sure but set it to False looks right according to paper?

## Testing Done

- Hardware Type: <BLANK>


vaibhavjindal:
@cyr0930 Thanks for this change. Yes, we should keep the default setting as False. 

Added a commit to the PR to make sure that the tests keep passing with this change.

vaibhavjindal (approved):

PR #696, opened 2025-05-02
Use logits_to_keep logic for training runs

### LINKED ISSUE #694: `logits_to_keep` is not working in training mode
### 🐛 Describe the bug

`logits_to_keep` should work in training mode as HF transformers [does](https://github.com/huggingface/transformers/blob/v4.51.3/src/transformers/models/llama/modeling_llama.py#L836).

Most models in Liger only trim logits in inference mode:
https://github.com/linkedin/Liger-Kernel/blob/05b43a14913ced3776aa3fc50020089b8c0d63c1/src/liger_kernel/transformers/model/gemma.py#L217-L219

We should trim hidden states instead to ensure that Liger and HF have same behaviors when `logits_to_keep` is given. See how Liger's gemma3 does:
https://github.com/linkedin/Liger-Kernel/blob/05b43a14913ced3776aa3fc50020089b8c0d63c1/src/liger_kernel/transformers/model/gemma3.py#L106-L110

### Reproduce

_No response_

### Versions

liger_kernel==0.5.8

vaibhavjindal:
## Summary

This PR fixes #694. 

It adds a change to slice the outputs according to `logits_to_keep` before calculating the loss during the training.

## Testing Done

- Hardware Type: <BLANK>


shivam15s (approved):
lgtm

Issue #537, opened 2025-01-22
Support the new Solar architecture


arnavgarg1:
### 🚀 The feature, motivation and pitch

This model from Upstage is extremely strong for models that fit on a single GPU for training and inference! https://huggingface.co/upstage/solar-pro-preview-instruct. However, it does use a custom architecture `solar` which is based on Llama/Mistral but modifies the forward pass to add long range residual connections. It would be awesome to support this architecture natively out of the box! 

### Additional context

Thank you so much for this awesome project :)

vulkomilev:
I will try to work on it

arnavgarg1:
That would be awesome @vulkomilev

PR #692, opened 2025-05-02
Qwen3 model support

vaibhavjindal:
## Summary

This pull request introduces support for the Qwen3 model within the Liger-Kernel framework.
Fixes: https://github.com/linkedin/Liger-Kernel/issues/690

## Testing Done

A simple check:
```
from transformers import Qwen3ForCausalLM
from liger_kernel.transformers.monkey_patch import apply_liger_kernel_to_qwen3

apply_liger_kernel_to_qwen3()
model = Qwen3ForCausalLM.from_pretrained("Qwen/Qwen3-8B")

print(model)
```
gives the following output:
Screenshot 2025-05-02 at 12 15 14 AM

- Hardware Type: <BLANK>
- [x] run `make test` to ensure correctness
- [x] run `make checkstyle` to ensure code style
- [x] run `make test-convergence` to ensure convergence

PR #699, opened 2025-05-04
Solar pro implementation

KareemMusleh:
## Summary
implements #537 

## Details
I don't know if `modeling_solar.py` and `configuration_solar.py` are in the right place.

I also changed `labels is not None` to `self.training and (labels is not None)` in `solar.py`. To make it similar to how `llama.py` is written.

## Testing Done

- Hardware Type: NVIDIA GeForce MX230
- [ ] write new tests


KareemMusleh:
what other tests should I add?

vaibhavjindal:
> what other tests should I add?

Please add fp16 and fp32 convergence tests as well. 
See https://github.com/linkedin/Liger-Kernel/pull/685 and https://github.com/linkedin/Liger-Kernel/pull/692

KareemMusleh:
@vaibhavjindal done! I don't have the hardware to test it but I think it should work.

Issue #703, opened 2025-05-05
Allow option to use fused linear cross entropy during validation


connermanuel:
### 🚀 The feature, motivation and pitch

Currently, the `forward` methods of various models are hardcoded so that the fused linear cross entropy loss is only used if (a) `model.training` is True and (b) labels are provided. This means that in any other scenario, fused linear cross entropy is not used - instead, the logits are materialized and, if labels are provided, regular cross entropy is used. 

This can lead to memory spikes in validation, which can be annoying especially if the user doesn't require logits, e.g.:

![Image](https://github.com/user-attachments/assets/6acc6f0f-4b3d-46ec-b0fe-8d2f2d82f337)

My proposal is to add a kwarg to the model forward methods that allows users to explicitly choose whether they want to use the fused kernel or if they want to materialize logits and use regular CE loss. This kwarg should be optional: if not provided, default to the original behavior.

Basically, changing this:
```
if self.training and labels is not None:
    loss = LigerForCausalLMLoss(...)
else:
    logits = forward(...)
    loss = CrossEntropy(logits, ...)
```
into this:
```
# kwarg: skip_logits: Optional[bool]

if skip_logits is None:
    skip_logits = self.training and labels is not None

if skip_logits:
    loss = LigerForCausalLMLoss(...)
else:
    logits = forward(...)
    loss = CrossEntropy(logits, ...)
```

If this sounds good I'll make the PR for it.

shivam15s:
Sounds good to me. Feel free to create a PR

shimizust:
Thanks for adding this!

PR #704, opened 2025-05-06
[WIP] Allow option to choose between fused LCE and CE loss

connermanuel:
## Summary
More context in #703
This adds a kwarg to the forward method of models that allows users to choose if they want to use fused LCE or if they want to materialize logits and use regular CE loss. Currently this decision is made implicitly using model state.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/transformers/model/gemma.py
+++ b/src/liger_kernel/transformers/model/gemma.py
@@ -145,6 +145,7 @@ def lce_forward(
     return_dict: Optional[bool] = None,
     cache_position: Optional[torch.LongTensor] = None,
     logits_to_keep: Union[int, torch.Tensor] = 0,
+    skip_logits: Optional[bool] = None,
     **loss_kwargs,
 ) -> Union[Tuple, CausalLMOutputWithPast]:
     r"""
@@ -207,8 +208,15 @@ def lce_forward(
     shift_labels = loss_kwargs.pop("shift_labels", None)
     logits = None
     loss = None
-    # if in training mode, don't materialize logits
-    if self.training and (labels is not None or shift_labels is not None):
+
+    if skip_logits and labels is None and shift_labels is None:
+        raise ValueError("skip_logits is True, but labels and shift_labels are None")
+
+    if skip_logits is None:
+        # By default, if in training mode, don't materialize logits
+        skip_logits = self.training and (labels is not None or shift_labels is not None)
+
+    if skip_logits:
         loss = LigerForCausalLMLoss(
             hidden_states=kept_hidden_states,
             lm_head_weight=self.lm_head.weight,
@@ -217,7 +225,7 @@ def lce_forward(
             hidden_size=self.config.hidden_size,
             **loss_kwargs,
         )
-    else:  # if in inference mode materialize logits
+    else:
         logits = self.lm_head(kept_hidden_states)
         if labels is not None:
             loss = self.loss_function(
--- a/src/liger_kernel/transformers/model/gemma2.py
+++ b/src/liger_kernel/transformers/model/gemma2.py
@@ -152,6 +152,7 @@ def lce_forward(
     return_dict: Optional[bool] = None,
     cache_position: Optional[torch.LongTensor] = None,
     logits_to_keep: Union[int, torch.Tensor] = 0,
+    skip_logits: Optional[bool] = None,
     **loss_kwargs,
 ) -> Union[Tuple, CausalLMOutputWithPast]:
     r"""
@@ -219,8 +220,15 @@ def lce_forward(
     shift_labels = loss_kwargs.pop("shift_labels", None)
     logits = None
     loss = None
-    # if in training mode, don't materialize logits
-    if self.training and (labels is not None or shift_labels is not None):
+
+    if skip_logits and labels is None and shift_labels is None:
+        raise ValueError("skip_logits is True, but labels and shift_labels are None")
+
+    if skip_logits is None:
+        # By default, if in training mode, don't materialize logits
+        skip_logits = self.training and (labels is not None or shift_labels is not None)
+
+    if skip_logits:
         loss = LigerForCausalLMLoss(
             hidden_states=kept_hidden_states,
             lm_head_weight=self.lm_head.weight,
@@ -231,7 +239,7 @@ def lce_forward(
             **loss_kwargs,
         )
 
-    else:  # if in inference mode materialize logits
+    else:
         logits = self.lm_head(kept_hidden_states)
         if self.config.final_logit_softcapping is not None:
             logits = logits / self.config.final_logit_softcapping
--- a/src/liger_kernel/transformers/model/gemma3.py
+++ b/src/liger_kernel/transformers/model/gemma3.py
@@ -41,6 +41,7 @@ def causal_forward(
     return_dict: Optional[bool] = None,
     cache_position: Optional[torch.LongTensor] = None,
     logits_to_keep: Union[int, torch.Tensor] = 0,
+    skip_logits: Optional[bool] = None,
     **loss_kwargs,
 ) -> Union[Tuple, CausalLMOutputWithPast]:
     r"""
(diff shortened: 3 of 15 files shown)


connermanuel:
Would appreciate feedback on the interface here:
* `skip_logits` was the natural choice for developing this because it emulates the existing logic flow. Could consider naming it something like `return_logits` or `use_fused_kernel` instead.
* Added an error here when `skip_logits` is True but there are no labels. If you don't have labels, you're not returning loss, and it doesn't make sense to return neither loss nor logits

Tcc0403:
Thanks for the contribution! But I'm a little bit confused the goal of this PR. 

My understanding of the proposal is that it enables users to access logits while the model is in training mode. Am I understanding this correctly? How is it different from patching CE without patching FLCE? 

If a user only want to retrieve logits for validation, simply setting the model to evaluation mode can materialize full logits and the model is still able to calculate CE loss if `labels` or `shift_labels` is provided.

Or is it for a flexible option to swtich between CE and FLCE in training mode? Could you elaborate on the scenarios where users can benefit from this option?

chiwanpark:
@Tcc0403 I think this PR is valuable when the validation metric is just loss in SFT scenario. There is no need to materialize logits if the validation metric is just validation loss. However, Liger-Kernel currently materializes the logits even if we just compute validation loss.

#703 (as posted by the PR author) shows the efficient usage of this option.

chiwanpark:
@connermanuel BTW, you need to implement this feature for qwen3_moe models that is recently merged in #706.

Tcc0403:
@chiwanpark Thanks for the explanation. I misunderstood the intent of this PR. So it's more about boardening the use of flce in eval mode, not about falling back to CE in training mode.

connermanuel:
@Tcc0403 @chiwanpark Thanks both. The primary intent is to enable FLCE for validation, but I suppose a secondary benefit is enabling accessing logits during training.
Added qwen3-moe support.

Tcc0403:
> I also didn't add it to llava because llava never returns logits, let me know if I should put it in anyway as something of a no-op to keep the interface the same.

Yes, put it in llava as well. It seems we missing validation mode support for llava, keep the interface so later PR can handle it.

> Added qwen3-moe support. Is convergence testing required here?

I think we can adjust the workflow of convergence test for it. Adding an additional validation step (forward pass only) after for loop to check the following metrics: 
1. losses in eval mode and train mode 
2. losses when `skip_logits=False` in eval mode
3. logits when  `skip_logits=False` in eval mode

connermanuel:
@Tcc0403 I modified the tests, can you see if this is what you meant?

test/convergence/bf16/test_mini_models.py
```diff
@@ -826,19 +826,33 @@ def run_mini_model(
...
     optimizer = torch.optim.AdamW(model.parameters(), lr=lr)
 
-    loss_list = []
+    train_loss_list = []
 
     for i in range(num_steps):
         batch = next(loader_iter).to(model.device)
         optimizer.zero_grad()
+        if with_liger:
+            batch["skip_logits"] = True
```
Tcc0403, line 835: We don't need to pass `skip_logits` in training loop because default is True for liger patched model.

connermanuel: Sounds good, by this reasoning I should remove `skip_logits=False` from validation as well

test/convergence/bf16/test_mini_models.py
```diff
@@ -826,19 +826,33 @@ def run_mini_model(
...
+        eval_output = model(**eval_batch)
 
     MINI_MODEL_SETUPS[model_name].liger_kernel_patch_revert_func(**revert_kwargs)
-    return {"loss": loss_list, "logits": output.logits, "model": model}
+    return {
+        "train_loss": train_loss_list,
+        "eval_loss": eval_output.loss.item(),
+        "eval_logits": eval_output.logits,
+        "model": model,
+    }
```
Tcc0403, line 855: I prefer keeping existing names, `loss` and `logits` to minimize the difference between files.

test/convergence/bf16/test_mini_models.py
```diff
@@ -1165,23 +1179,23 @@ def test_mini_model(
...
-
-    # # Compare the logits from the last step
-    # assert_verbose_allclose(
-    #     expected_output["logits"],
-    #     actual_output["logits"],
-    #     atol=logits_atol,
-    #     rtol=logits_rtol,
-    # )
+    # Compare the logits from the last step
+    if expected_output["eval_logits"] is not None and actual_output["eval_logits"] is not None:
```
Tcc0403, line 1192: Is this line only for llava since it's the only model that doesn't generate logits?

test/convergence/bf16/test_mini_models.py
Tcc0403: We can keep `skip_logits=False` in validation since we want to make sure the logits can be computed correctly.

test/convergence/bf16/test_mini_models.py
connermanuel: Yes

test/convergence/bf16/test_mini_models.py
connermanuel: Sounds good, I can add eval loss to the loss list then

connermanuel:
On an H100, three of the convergence tests are failing:
```
FAILED test/convergence/fp32/test_mini_models.py::test_mini_model[mini_qwen2_vl-32-0.0001-dtype7-1e-05-0.1-0.005-1e-05-0.005-1e-05]
FAILED test/convergence/fp32/test_mini_models.py::test_mini_model[mini_qwen2_5_vl-32-0.0001-dtype8-1e-05-0.1-0.005-1e-05-0.005-1e-05]
FAILED test/convergence/fp32/test_mini_models.py::test_mini_model[mini_granite3-32-0.0001-dtype16-1e-08-0.0001-0.005-1e-05-0.005-1e-05]
```
They all have similar trajectories as the reference
```
>           raise AssertionError("\n".join(mismatch_details))
E           AssertionError: Number of mismatched elements: 29133
E           Mismatch at index (1, 124, 6158): tensor1[(1, 124, 6158)] = -0.3988599479198456, tensor2[(1, 124, 6158)] = -0.3936949074268341
E           Mismatch at index (1, 124, 8486): tensor1[(1, 124, 8486)] = -1.195212483406067, tensor2[(1, 124, 8486)] = -1.1899968385696411
E           Mismatch at index (1, 124, 14097): tensor1[(1, 124, 14097)] = -0.02671273611485958, tensor2[(1, 124, 14097)] = -0.0215948186814785
E           Mismatch at index (1, 124, 15443): tensor1[(1, 124, 15443)] = -0.045016493648290634, tensor2[(1, 124, 15443)] = -0.050152797251939774
E           Mismatch at index (1, 124, 23038): tensor1[(1, 124, 23038)] = -1.4875845909118652, tensor2[(1, 124, 23038)] = -1.492673397064209
E           ... and 29128 more mismatched elements.
```
and these failures happen during training, which, AFAICT doesn't have any behavior changes since it would follow the same code path anyway.

Tcc0403:
@connermanuel What you showed seems like a numerical error to me. Feel free to relax logits tolerances of those failing models.

connermanuel:
@Tcc0403 I have updated the logits tolerances for select models. I tried to keep them reasonable but `qwen_2_5_vl` in particular proved to have a good amount of divergence, though judging by some comments left on those test params that seems expected.

Tcc0403:
> @Tcc0403 I have updated the logits tolerances for select models. I tried to keep them reasonable but `qwen_2_5_vl` in particular proved to have a good amount of divergence, though judging by some comments left on those test params that seems expected.

Mark a TODO next to qwen_25_vl logits tolerances. I'll further investigate the issue.

connermanuel:
There is already a TODO beside the params for those models because the loss was already relaxed. I'll adjust that todo

Tcc0403 (approved):

connermanuel:
@Tcc0403 @yundai424 Can I merge this, or do you need to run the workflows first?

Tcc0403:
> @Tcc0403 @yundai424 ~Can I merge this, or do you need to run the workflows first?~ Updated one more set of tolerances which dismissed the approval, after approval am I clear to merge?

Yes, merging it now. Thanks for the contribution!

Tcc0403 (approved):

Issue #709, opened 2025-05-12
Import Error: no _CONFIG_FOR_DOC in liger_kernel/transformers/model/gemma.py


Andy-Cheng:
### 🐛 Describe the bug

Hi,
I found that there's no 
1. _CONFIG_FOR_DOC from transformers.models.gemma.modeling_gemma  
2. transformers.models.gemma.modeling_gemma  from transformers.models.gemma.modeling_gemma

when I tried to use the latest liger_kernel and transformers.

[Environment]
transformers 4.52.0.dev0
liger_kernel   0.5.9

### Versions

Environment Report:
-------------------
Operating System: Linux-5.15.0-136-generic-x86_64-with-glibc2.35
Python version: 3.12.9
Liger Kernel version: 0.5.9
PyTorch version: 2.6.0+cu124
CUDA version: 12.4
HIP(ROCm) version: Not available
Triton version: 3.2.0
Transformers version: 4.52.0.dev0
XPU version: XPU Not Available

Andy-Cheng:
I think there's no _CONFIG_FOR_DOC for

Tcc0403:
Transformers got a new decorator `@auto_docstring` in https://github.com/huggingface/transformers/pull/33771

We need to update our modelings.

ringohoffman:
Here is [the documentation Hugging Face wrote for the `auto_docstring` decorator](https://github.com/huggingface/transformers/blob/1c7139580d94c04abe6f7876dd6a1cce60103813/docs/source/en/auto_docstring.md). Is there an actual docsite for `liger_kernel` where we can see how the docstrings get displayed?

Tcc0403:
No, we don't have a docsite. I think we can just simply remove alll docstring decorators before we make one.

PR #712, opened 2025-05-14
fix: remove docstring imports in transformer patches

NanoCode012:
## Summary

Transformers did a refactor on docstrings to use autodocstring instead on main. This would break any patches that import `_CONFIG_FOR_DOC` or `*INPUTS_DOCSTRING`

https://github.com/huggingface/transformers/pull/33771

An error example:
```
ImportError: cannot import name '_CONFIG_FOR_DOC' from 'transformers.models.gemma.modeling_gemma' (/opt/hostedtoolcache/Python/3.11.12/x64/lib/python3.11/site-packages/transformers/models/gemma/modeling_gemma.py)
```

This PR removes any imports that would break it.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/transformers/model/gemma.py
+++ b/src/liger_kernel/transformers/model/gemma.py
@@ -8,18 +8,12 @@ import torch
 from torch.nn import CrossEntropyLoss
 from transformers.cache_utils import Cache
 from transformers.modeling_outputs import CausalLMOutputWithPast
-from transformers.models.gemma.modeling_gemma import _CONFIG_FOR_DOC
-from transformers.models.gemma.modeling_gemma import GEMMA_INPUTS_DOCSTRING
-from transformers.utils import add_start_docstrings_to_model_forward
-from transformers.utils import replace_return_docstrings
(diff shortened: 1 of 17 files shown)


NanoCode012:
tagging @vaibhavjindal for review as I saw you made a lot of patch changes (sorry if you're the wrong person for this)

vaibhavjindal:
Thanks for the change @NanoCode012 . I see that @Tcc0403 was following up on a similar issue: https://github.com/linkedin/Liger-Kernel/issues/709, so tagging him here to take a look.

Tcc0403:
I'm not familiar with docstring. But since we don't have a library documentation, I think it's okay to remove all docstring related decorators until we make one. WDYT @vaibhavjindal ?

vaibhavjindal:
@Tcc0403 yes, let's merge this change. This will ensure that things won't break for older and newer versions. We can later add support for `auto_docstring` decorator for newer version of transformers.

vaibhavjindal (approved):
LGTM! Thanks for the change.

Issue #742, opened 2025-06-02
[RFC] Logits numerical issues in convergence test


Tcc0403:
### 🐛 Describe the bug

This issue is to discuss how we should modify our convergence test to handle numerical issues of logits.

### Context
In #704, we make `FusedLinearCrossEntropy`(flce) patched models have the freedom to switch between materializing logits and skipping logits. It allows us to compare logits from the last step again in flce convergence test, `test/convergence/(bf16|fp32)/test_mini_models.py`. 

However, this change also exposed the hidden numerical issue that we didn't take care of. 

The reasons why tests failed can be categorized into two types:
1. numerical instability of floating points due to the limitation of bf16 representation, rounding errors, nonassociativity of floating point calculation, etc.
2. actual bugs inside our implementation. 

Ideally, we want to ignore type1 and handle type2 errors. But in reality, best we can do to ignore type1 errors is just increasing tolerences, which also increases the difficulty to capture type2 errors, the actual bugs in our kernels.

### Existed methods

Before discussing how we can modify our convergence tests to address the issue, we can take a look of how other mainstream libraries handle it:
#### sglang
[`check_close_model_outputs()`](https://github.com/sgl-project/sglang/blob/a2cb5913a00965a8bb33f3986710adaaf1828a8e/python/sglang/test/runners.py#L758)

techniques:
[calculate the ROUGE-L score](https://github.com/sgl-project/sglang/blob/a2cb5913a00965a8bb33f3986710adaaf1828a8e/python/sglang/test/runners.py#L770)
[compare top logprobs](https://github.com/sgl-project/sglang/blob/a2cb5913a00965a8bb33f3986710adaaf1828a8e/python/sglang/test/runners.py#L776)
#### vllm
[`check_logprobs_close()`](https://github.com/vllm-project/vllm/blob/ca2f6b9c301df6dbe2e5c83c705051f478140695/tests/models/utils.py#L80)

techniques:
[compare logprob only when token id mismatch](https://github.com/vllm-project/vllm/blob/ca2f6b9c301df6dbe2e5c83c705051f478140695/tests/models/utils.py#L97)
#### verl
[`test_hf_causal_models()`](https://github.com/volcengine/verl/blob/0e127b208b7ec89bbd3647668973a35fe5d81390/tests/models/test_transformer.py#L39)

techniques:
[...]

PR #745, opened 2025-06-05
Trying out logprobs and top logprobs for testing rather than logits.

Manan17:
## Summary

Just testing out logprobs as mentioned in #742 

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/test/convergence/bf16/test_mini_models.py
+++ b/test/convergence/bf16/test_mini_models.py
@@ -896,7 +896,7 @@ def run_mini_model(
             torch.bfloat16,
             1e-3,
             1e-2,
-            1,  # 1e-1
+            1e-1,  # 1e-1
             1e-1,  # 1e-2
             1e-2,
             1e-2,
@@ -966,7 +966,7 @@ def run_mini_model(
             torch.bfloat16,
             1e-3,
             1e-2,
-            1,  # 1e-1
+            1e-1,  # 1e-1
             1e-1,  # 1e-2
             1e-2,
             1e-2,
@@ -1203,6 +1203,17 @@ def test_mini_model(
             actual_logprobs,
             atol=logits_atol,
             rtol=logits_rtol
+        )
+        k = 5
+        exp_topk_vals, _ = torch.topk(expected_logprobs, k, dim=-1)
+        act_topk_vals, _ = torch.topk(actual_logprobs, k, dim=-1)
+
+        # Compare top-k logprobs with tolerance (ignoring order)
+        max_diff = torch.max(torch.abs(exp_topk_vals - act_topk_vals)).item()
+        print(f"Top-{k} logprobs max diff: {max_diff:.6f}")
+        assert torch.all(torch.abs(exp_topk_vals - act_topk_vals) < (logits_atol + logits_rtol * torch.abs(exp_topk_vals))), (
+            f"Top-{k} logprobs are not all close (atol={logits_atol}). "
+            f"Max diff: {max_diff:.6f}"
         )
     # Compare the params from the last step
     # Iterate over the model's parameters and compare them
--- a/test/convergence/bf16/test_mini_models.py
+++ b/test/convergence/bf16/test_mini_models.py
@@ -1106,7 +1106,7 @@ def run_mini_model(
             1e-3,
             1e-2,
             1e-1,
-            1e-2,
+            1e-1,
             1e-2,
             1e-2,
             marks=pytest.mark.skipif(not supports_bfloat16(), reason="bfloat16 not supported on this GPU"),
@@ -1119,7 +1119,7 @@ def run_mini_model(
             1e-3,
             1e-2,
             1e-1,
-            1e-2,
+            1e-1,
             1e-2,
             1e-2,
             marks=pytest.mark.skipif(not supports_bfloat16(), reason="bfloat16 not supported on this GPU"),
@@ -1148,7 +1148,7 @@ def run_mini_model(
             1e-3,
             1e-2,
             1e-1,
-            1e-2,
+            1e-1,
             1e-2,
             1e-2,
             marks=[
@@ -1189,13 +1189,21 @@ def test_mini_model(
 
     # Compare the logits from evaluation step
     if expected_output["logits"] is not None and actual_output["logits"] is not None:
+        # assert_verbose_allclose(
+        #     expected_output["logits"],
+        #     actual_output["logits"],
+        #     atol=logits_atol,
+        #     rtol=logits_rtol,
+        # )
+        expected_logprobs = torch.nn.functional.log_softmax(expected_output["logits"], dim=-1)
+        actual_logprobs = torch.nn.functional.log_softmax(actual_output["logits"], dim=-1)
+        
         assert_verbose_allclose(
-            expected_output["logits"],
-            actual_output["logits"],
+            expected_logprobs,
+            actual_logprobs,
             atol=logits_atol,
-            rtol=logits_rtol,
+            rtol=logits_rtol
         )
-
     # Compare the params from the last step
     # Iterate over the model's parameters and compare them
     for expected_param, actual_param in zip(


test/convergence/bf16/test_mini_models.py
```diff
@@ -1189,13 +1189,32 @@ def test_mini_model(
...
+        
         assert_verbose_allclose(
-            expected_output["logits"],
-            actual_output["logits"],
+            expected_logprobs,
+            actual_logprobs,
             atol=logits_atol,
-            rtol=logits_rtol,
+            rtol=logits_rtol
         )
```
Tcc0403: We don't have to compare all logprobs

test/convergence/bf16/test_mini_models.py
```diff
@@ -1189,13 +1189,32 @@ def test_mini_model(
...
+        exp_topk_vals, _ = torch.topk(expected_logprobs, k, dim=-1)
+        act_topk_vals, _ = torch.topk(actual_logprobs, k, dim=-1)
 
+        # Compare top-k logprobs with tolerance (ignoring order)
+        max_diff = torch.max(torch.abs(exp_topk_vals - act_topk_vals)).item()
+        print(f"Top-{k} logprobs max diff: {max_diff:.6f}")
+        assert torch.all(torch.abs(exp_topk_vals - act_topk_vals) < (logits_atol + logits_rtol * torch.abs(exp_topk_vals))), (
+            f"Top-{k} logprobs are not all close (atol={logits_atol}). "
+            f"Max diff: {max_diff:.6f}"
+        )
```
Tcc0403: Make it a test util

test/convergence/bf16/test_mini_models.py
```diff
@@ -896,7 +896,7 @@ def run_mini_model(
             torch.bfloat16,
             1e-3,
             1e-2,
-            1,  # 1e-1
+            1e-1,  # 1e-1
             1e-1,  # 1e-2
```
Tcc0403, line 908: After removing all logprobs comparison, we can try setting it lower.
sglang only has atol and sets it to 5e-2 ([decode_tolerance](https://github.com/search?q=repo%3Asgl-project/sglang%20decode_tolerance&type=code))
verl sets [`(atol, rtol) = (1e-2, 1e-5)`](https://github.com/volcengine/verl/blob/0e127b208b7ec89bbd3647668973a35fe5d81390/tests/models/test_transformer.py#L134), but it's mean of all logprobs not topk

Manan17: Does not work with lower tolerance.
For gemma3, it passes when atol=1e-1 and rtol=1

Manan17: I tested this out with fp32, it fails for most of the models where old logic for checking the logits is passing.

test/utils.py
```diff
@@ -57,6 +57,19 @@ def set_seed(seed=42):
...
 
+def check_logprobs(tensor1, tensor2, rtol=1e-05, atol=1e-08, k=5, max_print=5):
+    act_topk_vals, _ = torch.topk(tensor1, k, dim=-1)
+    exp_topk_vals, _ = torch.topk(tensor2, k, dim=-1)
+
+    # Compare top-k logprobs with tolerance (ignoring order)
+    max_diff = torch.max(torch.abs(exp_topk_vals - act_topk_vals)).item()
+    print(f"Top-{k} logprobs max diff: {max_diff:.6f}")
+    assert torch.all(torch.abs(exp_topk_vals - act_topk_vals) < (rtol + rtol * torch.abs(exp_topk_vals))), (
+        f"Top-{k} logprobs are not all close (tolerance={rtol + rtol * torch.abs(exp_topk_vals)}). "
```
Tcc0403, line 68:
```suggestion
    assert torch.all(torch.abs(exp_topk_vals - act_topk_vals) < (atol + rtol * torch.abs(exp_topk_vals))), (
        f"Top-{k} logprobs are not all close (tolerance={atol + rtol * torch.abs(exp_topk_vals)}). "
```

test/convergence/bf16/test_mini_models.py
Tcc0403: Since we are comparing values in log-space, the total tolerance here is actually relative tolerance.

test/convergence/bf16/test_mini_models.py
```diff
@@ -1189,13 +1190,10 @@ def test_mini_model(
...
     if expected_output["logits"] is not None and actual_output["logits"] is not None:
-        assert_verbose_allclose(
-            expected_output["logits"],
-            actual_output["logits"],
-            atol=logits_atol,
-            rtol=logits_rtol,
-        )
-
+        actual_logprobs = torch.nn.functional.log_softmax(actual_output["logits"], dim=-1)
+        expected_logprobs = torch.nn.functional.log_softmax(expected_output["logits"], dim=-1)
```
Tcc0403: Make `log_softmax()` and `topk()` a util function, so we can call it in `run_mini_model()` to avoid storing all logits

test/convergence/bf16/test_mini_models.py
```diff
@@ -1189,13 +1190,10 @@ def test_mini_model(
...
+        check_logprobs(actual_logprobs,expected_logprobs, atol=logits_atol,rtol=logits_rtol)
```
Tcc0403: Assuming we have top logprobs calculated and stored, we only need to call 
`assert_verbose_allclose(actual_top_logprobs, expected_top_logprobs, atol=logprob_atol, rtol=logprob_rtol)`.

test/convergence/bf16/test_mini_models.py
Manan17: Can we just check the rtol?
like: tolerance = rtol * torch.abs(tensor2)

Manan17:
@Tcc0403 Can you have a look at the changes, I have tested it. 
Let me know what you think, I will update it for the multimodal tests as well. 
What should be done for test_mini_models_with_logits?

test/convergence/bf16/test_mini_models.py
Tcc0403: absolute diff for two logprobs (logA - logB) = relative diff for two probs (A / B), which means the whole tolerance (`atol + rtol * torch.abs(expected)`) should be the maximum relative diff we can accept. 

I think that's also why sglang only has a single tolerance in their test.

Tcc0403:
> What should be done for test_mini_models_with_logits

check logprobs as well for consistency

I'm planning to rewrite convergence tests so just ignore namings for now.

test/utils.py
```diff
@@ -56,6 +56,13 @@ def set_seed(seed=42):
     # Python hash seed
     os.environ["PYTHONHASHSEED"] = str(seed)
 
+def get_logprobs(tensor):
+    return torch.nn.functional.log_softmax(tensor, dim=-1)
+
+def get_topk(tensor, k=5):
+    topk = torch.topk(tensor, k, dim=-1)
+    return topk
```
Tcc0403: add decorator [`@torch.no_grad`](https://docs.pytorch.org/docs/stable/generated/torch.no_grad.html)

Manan17:
> > What should be done for test_mini_models_with_logits
> 
> check logprobs as well for consistency
> 
> I'm planning to rewrite convergence tests so just ignore namings for now.

Gotcha!
I tried testing with mean logprobs as well. 
The tests pass with lower tolerance values. Verl has set atol=1e-2 and rtol=1e-5, which works for us as well in bf16.

Tcc0403:
> I tried testing with mean logprobs as well.
> The tests pass with lower tolerance values. Verl has set atol=1e-2 and rtol=1e-5, which works for us as well in bf16.

What mean logprobs do you pick? I checked verl impl, they pick [per-token logprobs for the given labels](https://github.com/volcengine/verl/blob/16662ceff4c90c0893dd2d249965bed94f2a5e1a/verl/utils/torch_functional.py#L56)

test/utils.py
```diff
@@ -56,6 +56,15 @@ def set_seed(seed=42):
     # Python hash seed
     os.environ["PYTHONHASHSEED"] = str(seed)
 
+@torch.no_grad
+def get_logprobs(tensor):
+    return torch.nn.functional.log_softmax(tensor, dim=-1)
```
Tcc0403: For numerical stability if tensor is bf16
```suggestion
@torch.no_grad
def get_logprobs(tensor):
    return torch.nn.functional.log_softmax(tensor, dim=-1, dtype=torch.float32) # fp32 for stability
```

Manan17:
I tried top 20 logprobs and it was able to pass tests for all the models! @Tcc0403

Manan17:
The tolerance for gemma3 multimodal model had to be set high as it does not pass the tests for loss and topk_logprobs. 
The atol and rtol set it 1e-1.

Tcc0403:
> The tolerance for gemma3 multimodal model had to be set high as it does not pass the tests for loss and topk_logprobs.
> The atol and rtol set it 1e-1.

Yeah, I think we can compromise with 1e-1 before further investigation in numerical issue. Just make them all green first unless there's an obvious mismatch.

shimizust (approved):
Thanks for making these changes!

PR #752, opened 2025-06-10
Automate benchmarking - disabling the docs workflow

Manan17:
## Summary

The docs workflow currently replaces the gh-pages branch with new files due to which the benchmark data is lost. Currently I am disabling this so that we can get the data appended in the gh-pages branch.
Also, modified the script to add commit hashes to a json file, so that we do not call github api to fetch it due to API limitations

## Testing Done

- Hardware Type: <BLANK>


vaibhavjindal:
@Manan17 looks like there are some issues with the yaml files that we need to fix. We need to fix both the benchmark.yml and docs.yml

PR #759, opened 2025-06-12
Automate benchmarking - changed the scripts for backward pass.

Manan17:
## Summary

The scripts are updated so that we get benchmarking results on backward pass as well. 
All benchmark scripts were run on a H100, it took around 55 minutes to run. 

There are total of 26 scripts, so each generates a plot for speed - forward, backward and full pass
for memory - full pass. Total of 104.
And the output (csv file) was of 460kB.

## Testing Done

- Hardware Type: <BLANK>


vaibhavjindal (approved):
Amazing, thanks!

PR #762, opened 2025-06-17
fix: import tanh

jue-jue-zi:
## Summary
For triton version less than `3.0.0`, `tanh` function must be imported from `triton.language.math`. There is a duplicated import in the file `dyt.py` before checking the triton version and thus needs to be removed.

## Testing Done

- Hardware Type: <BLANK>


vaibhavjindal (approved):
Thanks for the change!

PR #771, opened 2025-06-23
Revamping the UI and flow of the webpage.

Manan17:
## Summary

This pr adds a main dashboard page where you can see any major difference between various commits. 
The threshold for change has been set to 10% (Running benchmarks four times on the same commit gave a difference of about 10%)

## Testing Done

- Hardware Type: <BLANK>


benchmarks/script.js
```diff
@@ -0,0 +1,260 @@
+const referenceCommit = 'c1bdc24';
```
shimizust, line 2: We can do this in a subsequent PR, but given the importance of the reference commit, we should keep it in a separate constants file for clarity and have each kernel mapped to a "golden" reference commit

benchmarks/script.js
```diff
@@ -0,0 +1,260 @@
...
+
+async function fetchCSV(commit) {
+  const res = await fetch(`${benchmarkBase}/${commit}/benchmark.csv`);
+  if (!res.ok) return null;
+  const text = await res.text();
+  const rows = text.trim().split('\n');
+  const headers = rows[0].split(',');
+  const data = {};
+
+  console.log('Headers:', headers); // Debug headers
```
shimizust: nit: remove this

benchmarks/script.js
```diff
@@ -0,0 +1,260 @@
...
+  }));
+
+  renderTables();
+}
+
+function compareMetric(current, reference) {
+  if (current == null || reference == null) return '';
+  const delta = (current - reference) / reference;
+  if (delta > 0.10) return 'red';
+  if (delta < 0) return 'green';
```
shimizust: Can you add a more detailed comment on this (why 10%) and move the 0.10 into a constant to clarify?

benchmarks/script.js
```diff
@@ -0,0 +1,260 @@
...
+        kernelMeta[kernelKey] = {
+          x_label: row.x_label,
+          x_value: row.x_value
+        };
+      }
+      // Store configuration data for tooltips
+      if (!configData[kernelKey]) {
+        console.log('Row data for config:', {
+          extra_benchmark_config_str: row.extra_benchmark_config_str,
+          gpu_name: row.gpu_name,
```
shimizust: We should only be comparing if the GPU type is the same as the reference commit, or else the comparison isn't apples-to-apples. Since the modal benchmark script is now enforced for a certain gpu type, perhaps we can just put the GPU type used in the reference above the table as a note.

And we can add a TODO for subsequent PR to ensure the benchmark results we use are using same gpu type as reference.

shimizust (commented):
Thanks for adding this, had some comments which can be addressed in subsequent PR

benchmarks/script.js
Manan17: Modal automatically might upgrade an H100 request to H200 (keeping the pricing same when upgraded), there is no way to strictly define H100 GPU.

Issue #723, opened 2025-05-22
CI failure due to transformers VLM change


Tcc0403:
### 🐛 Describe the bug

CI details: [Qwen2VLConfig](https://github.com/linkedin/Liger-Kernel/actions/runs/15175089532/job/42680679210?pr=689#step:6:6549), [monkey patch impl related](https://github.com/linkedin/Liger-Kernel/actions/runs/15214784023/job/42797584570#step:5:1903)

Text config is seperated out from the general config in transformers>=4.52.0. 
[Qwen2VLRotaryEmbedding](https://github.com/huggingface/transformers/pull/37268/files#diff-09bc594f9680f1d042fd485106c68022d77b59831697a00b3b38f12a3e40f395L103-R104) takes `Qwen2VLTextConfig` intead of `Qwen2VLConfig` now.

https://github.com/huggingface/transformers/pull/37268

We also need to check all VLMs monkey patch since most models are refactored into TextModel and VisionModel

~blocked on: https://github.com/huggingface/transformers/issues/38331 (can walkaround by passing hidden_size and vocab_size to config dict directly though)~ fixed by https://github.com/huggingface/transformers/pull/38372

### Versions

liger_kernel==0.5.9
transformers>=4.52.0

Tcc0403:
`video_processor`(`Qwen2VLVideoProcessor`) is now [a required input](https://github.com/huggingface/transformers/blob/d03a3ca69226a776d9ed69cd90cbe8f6ebe8fa34/src/transformers/models/qwen2_vl/processing_qwen2_vl.py#L64) for `Qwen2VLProcessor`.

We might need to add torchvision to our dev dependency since it is required for videoprocessor.

Related PR: https://github.com/huggingface/transformers/pull/35206

Tcc0403:
fixed by #735 #738 #743 #755

PR #773, opened 2025-06-25
fix: AttributeError in gemma3.

ELanning:
## Summary

AttributeErrors in the gemma3 module - likely caused by a code change in `transformers`.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/transformers/model/gemma3.py
+++ b/src/liger_kernel/transformers/model/gemma3.py
@@ -208,7 +208,7 @@ def multimodal_forward(
     is_training = token_type_ids is not None and labels is not None
 
     # Replace image id woth PAD if the image token if OOV, to avoid index-errors
-    if input_ids is not None and self.config.image_token_index >= self.vocab_size:
+    if input_ids is not None and self.config.image_token_index >= self.model.vocab_size:
         special_image_mask = input_ids == self.config.image_token_index
         llm_input_ids = input_ids.clone()
         llm_input_ids[special_image_mask] = 0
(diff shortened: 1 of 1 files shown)


ELanning:
I don't think this is mergeable, I'm just opening this PR in-case anyone else has run into this / to start a discussion.
Feel free to take these changes and get it to a mergeable state.

src/liger_kernel/transformers/model/gemma3.py
```diff
@@ -229,7 +229,7 @@ def multimodal_forward(
 
     # Merge text and images
     if pixel_values is not None:
-        image_features = self.get_image_features(pixel_values)
+        image_features = self.model.get_image_features(pixel_values)
 
         if input_ids is None:
             special_image_mask = inputs_embeds == self.get_input_embeddings()(
```
ELanning, line 235: Strangely, `self.get_input_embeddings()` works- no need for `self.model` here 🤷

Tcc0403:
Thanks for reporting! I think this is also related to transformers VLM refactorings. 

Related issue/PR: #723 #735 #774

src/liger_kernel/transformers/model/gemma3.py
Tcc0403: https://github.com/huggingface/transformers/blob/1d45d90e5d1552eccb6d8cc9b7bba283ccefb808/src/transformers/models/gemma3/modeling_gemma3.py#L979

I guess transformers had it definfed on outer model as well for backward compatibility

PR #774, opened 2025-06-25
fix vocab_size path for gemma3

winglian:
## Summary
using `self.vocab_size` for the multimodal forward likely never worked or was deprecated in a transformers change.


ELanning:
Needs `image_features = self.model.get_image_features(pixel_values)` on line 232 too, otherwise worked for me.

.idea/workspace.xml
shimizust, line 1: can you remove and add .idea to the .gitignore

PR #783, opened 2025-07-02
fix(geglu, swiglu): match the casting in MLP layer for bfloat16

Tcc0403:
## Summary

Most MLPs are defined as:
```python
def forward(self, x):
    down_proj = self.down_proj(self.act_fn(self.gate_proj(x)) * self.up_proj(x))
    return down_proj
```
, where `self.act_fn` is [`torch.nn.GELU(approximate='tanh')`](https://github.com/huggingface/transformers/blob/25cd65ac43ee1a96cef4692bda0b110d1e3c6903/src/transformers/activations.py#L27)

However, in [LigerGELUMLP kernel](https://github.com/linkedin/Liger-Kernel/blob/08906d3d9cad77fcced2b9f6db0a385eff154f03/src/liger_kernel/ops/geglu.py#L22-L44), after upcasting `x` to float32, all computation are performed in float32. It can result in a huge discrepancy between original and patched models in bfloat16 scenario.

Here's the casting comparison:

Most Huggingface models' MLP
```python
def forward(self, x):
	ori_dtype = x.dtype
	# casting back before multiplying up projection
    down_proj = self.down_proj(gelu(self.gate_proj(x).float()).to(ori_dtype) * self.up_proj(x))
    return down_proj
```
LigerGEGLUMLP before this PR
```python
def forward(self, x):
    ori_dtype = x.dtype
	# casting back after multiplying up projection
    down_proj = self.down_proj((gelu(self.gate_proj(x).float()) * self.up_proj(x)).to(ori_dtype))
    return down_proj
```

This PR fixes the casting timing to match the former behavior.

Edit: Same fix can apply to `SwiGLUMLP` as well.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/ops/geglu.py
+++ b/src/liger_kernel/ops/geglu.py
@@ -40,7 +40,7 @@ def _geglu_tanh_forward_kernel(a, b, c, stride, n_cols: tl.constexpr, BLOCK_SIZE
     tanh_arg = sqrt_2_over_pi * (a_row + 0.044715 * a_cubed)
     tanh_result = tanh(tanh_arg)
     geglu_a = 0.5 * a_row * (1 + tanh_result)
-    c_row = geglu_a * b_row
+    c_row = geglu_a.cast(b_row.dtype) * b_row
     tl.store(c + col_offsets, c_row, mask=mask)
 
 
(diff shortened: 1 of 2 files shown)


shimizust:
Fixed by #778

PR #795, opened 2025-07-09
Fix gemma3 forward with skip_logits

BitPhinix:
Seems like this one was overlooked in https://github.com/linkedin/Liger-Kernel/pull/787


ELanning:
FWIW, also ran into this. I thought something was possibly wrong on my end since CI was presumably passing.

Tcc0403:
great catch!

PR #804, opened 2025-07-12
Fix illegal memory access in Triton RMSNorm kernel by casting program_id to int64fix(rms_norm): cast row_idx to int64 in _rms_norm_forward_kernel(# 803)

vvvdwbvvv:
## Testing Done

- Hardware Type: <BLANK>
- [ ] run `make test` to ensure correctness
- [ ] run `make checkstyle` to ensure code style
- [ ] run `make test-convergence` to ensure convergence

Issue #803, opened 2025-07-11
Liger-Kernel crashes on very long seqlen 2^31 / int64 indexing issue


sfc-gh-sbekman:
### 🐛 Describe the bug

Hi, in the context of developing support for training with very long seqlen https://www.arxiv.org/abs/2506.13996 - we are trying to break past 500K seqlen on a single gpu - that limit was fine on H100 with 80GB, but with H200 and B200 we have a lot more memory available and should be able to do more than 1M seqlen easily - but I'm running into int64 indexing issue with Liger-Kernel (and FA2 as well, but FA3 solves it).

Liger's rms norm crashes with:

```
[rank0]: RuntimeError: Triton Error [CUDA]: an illegal memory access was encountered
```

Please note that already at 500K and 8B llama we are dealing with `[1, 500_000, 4096]` a tensor of about 2^31 - which is the cause of the problem.

### Reproduce

```
import torch
import gc
from liger_kernel.transformers import AutoLigerKernelForCausalLM

checkpoint = "meta-llama/Llama-3.1-8B-Instruct"

torch.manual_seed(42)
device = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
dtype=torch.bfloat16

model = AutoLigerKernelForCausalLM.from_pretrained(checkpoint, torch_dtype=dtype).to(device)

seqlen = 1_000_000
hidden_size = model.config.hidden_size
hidden_states = torch.rand((1, seqlen, hidden_size), dtype=torch.bfloat16, requires_grad=True, device=device)

for idx in range(len(model.model.layers)):
    hidden_states = model.model.layers[idx].input_layernorm(hidden_states)
    print(f"{idx} {hidden_states.shape=}")
    gc.collect()
```

gives us:
```
0 hidden_states.shape=torch.Size([1, 1000000, 4096])
1 hidden_states.shape=torch.Size([1, 1000000, 4096])
Traceback (most recent call last):
  File "/code/users/stas/github/sf/arctictraining/long-seqlen-fa-liger.py", line 24, in <module>
    hidden_states = model.model.layers[idx].input_layernorm(hidden_states)
  File "/home/yak/.local/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1751, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/yak/.local/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1762, in _call_impl
[...]

PR #812, opened 2025-07-16
FusedAddRMSNorm: Fused residual addition and RMS Norm

vaibhavjindal:
## Summary

This PR adds the FusedAddRMSNorm kernel. This kernel is meant to replace a commonly observed pattern in decoder layers in transformers:
```
hidden_states = residual + hidden_states
residual = hidden_states
hidden_states = rmsnorm(hidden_states)
```
Some examples of this pattern:
- https://github.com/huggingface/transformers/blob/0dc2df5ddafe3cb5824ad24e85beba13e0aa6726/src/transformers/models/qwen3/modeling_qwen3.py#L271
- https://github.com/huggingface/transformers/blob/0dc2df5ddafe3cb5824ad24e85beba13e0aa6726/src/transformers/models/llama4/modeling_llama4.py#L393

Fusing the residual addition leads to both memory and speed savings. This is because we are able to reduce one memory read and multiple kernel launches for this pattern. This leads to ~15-20% memory savings and 2-4x speed up for (fwd+bwd) mode!

With this kernel, the above 3 operations can be written as a single efficient line of code:
```
hidden_states, residual = liger_fused_add_rms_norm(hidden_states, residual)
```

## Comparison Plots
- huggingface: Naive pytorch implementation
- liger_rms_norm: Liger RMS norm used for RMS norm, while keeping residual addition separate
- liger_fused_add_rms_norm: Fused residual addition and RMS norm

### Memory
fused_add_rms_norm_memory_full

### Speed
1. Forward
fused_add_rms_norm_speed_forward

2. Backward
fused_add_rms_norm_speed_backward

3. Full (Forward+Backward)
fused_add_rms_norm_speed_full

## Testing Done

- Hardware Type: H100


src/liger_kernel/ops/fused_add_rms_norm.py
```diff
@@ -0,0 +1,411 @@
...
+
+
+_str_to_casting_mode = {
+    "llama": _CASTING_MODE_LLAMA.value,
+    "gemma": _CASTING_MODE_GEMMA.value,
+    "none": _CASTING_MODE_NONE.value,
+}
+
+
+def fused_add_rms_norm_forward(X, R, W, eps, offset, casting_mode, row_mode):
```
Tcc0403: `row_mode` is not implemented right?

vaibhavjindal: yes, it is not implemented. Will remove it for now. Thanks!

vaibhavjindal:
@Tcc0403 removed the `row_mode` option for now.

src/liger_kernel/ops/fused_add_rms_norm.py
```diff
@@ -0,0 +1,409 @@
...
+    This is a commonly used pattern in the decoder layers of LLMs.
+    Some examples:
+    1. https://github.com/huggingface/transformers/blob/0dc2df5ddafe3cb5824ad24e85beba13e0aa6726/src/transformers/models/qwen3/modeling_qwen3.py#L271
+    2. https://github.com/huggingface/transformers/blob/0dc2df5ddafe3cb5824ad24e85beba13e0aa6726/src/transformers/models/llama4/modeling_llama4.py#L393
+
+    This kernel is inspired by the rms_norm forward kernel, and is adapted to support the residual addition in the forward pass.
+    The backward pass is also adapted to support the residual addition in the backward pass.
+    """
+
+    row_idx = tl.program_id(0)
```
Tcc0403: We should address the overflow issue in very long seqlen scenario by casting it to tl.int64 as well.

Related issue/PR: #803, #804

src/liger_kernel/ops/fused_add_rms_norm.py
```diff
@@ -0,0 +1,409 @@
...
+    3. hidden_states = rmsnorm(hidden_states)
+
+    The gradient of hidden_states and residual comes out be exactly same. The value of this gradient is
+    the sum of the gradient of the hidden_states in step 3 and the gradient of the residual in step 2.
+
+    The backward pass computation logic is same as the rms_norm backward kernel, except that the gradient
+    of the hidden_states in step 3 and the gradient of the residual in step 2 are summed up.
+    """
+
+    row_block_id = tl.program_id(0)
```
Tcc0403: ditto

sfc-gh-sbekman:
Awesome work, @vaibhavjindal !

Once landed this will get enabled automatically over the non-fused version, correct?

src/liger_kernel/ops/fused_add_rms_norm.py
vaibhavjindal: Thanks, done!

src/liger_kernel/ops/fused_add_rms_norm.py
vaibhavjindal: Done!

vaibhavjindal:
> Awesome work, @vaibhavjindal !
> 
> Once landed this will get enabled automatically over the non-fused version, correct?

Thanks @sfc-gh-sbekman! No, it won't get enabled automatically as we will have to change the patching logic in monkey_patch.py. It's not trivial to patch the existing models with this kernel as it replaces multiple rows, and not just a single class as is done by RMS norm. We still need to decide on the best way to use this kernel in the existing models :)

Tcc0403:
Thanks! Let's merge it first, then we can figure out how to patch it after.

Tcc0403 (dismissed):
lgtm

src/liger_kernel/ops/fused_add_rms_norm.py
```diff
@@ -0,0 +1,409 @@
...
+    - 'llama': matches the Llama implementation, where only the inverse RMS is computed on fp32.
+    - 'gemma': matches the Gemma implementation, where everything is cast to fp32, then computed, then cast back to the original dtype.
+    - 'none': no casting is done. The computation is done in the original dtype. This saves memory and is slightly faster, but has more error w.r.t. the original implementation.
+
+    The `in_place` option determines whether to modify dY in-place to store dX. This defaults to `True` to save memory.
+    """
+
+    @staticmethod
+    @ensure_contiguous
+    def forward(ctx, X, R, W, eps, offset=0.0, casting_mode="llama", in_place=False):
```
shimizust, line 377: Could you add a todo on adding the row_mode and _block_fused_rms_norm_forward_kernel for parity with rms_norm just so it's not forgotten if we make this the default?

shimizust (commented):
great find! Updating/maintaining the patching logic will be tricky

src/liger_kernel/ops/fused_add_rms_norm.py
vaibhavjindal: Sure, added TODOs for this.

winglian:
was there any benchmarking in real use cases outside of isolated residual addition and rms norm? 
I'm fine-tuning qwen3 and using the fused-add-rms-norm for the `pre_feedforward_layernorm` and the liger rms norm implementation for the rest of the rms norm modules and it is worse in both speed and memory

Screenshot 2025-07-30 at 10 00 56 AM

PR #582, opened 2025-02-24
Fix layer norm kernels

lancerts:
## Summary
Fix layer norm kernels, include the changes from https://github.com/linkedin/Liger-Kernel/pull/519/files

## Testing Done

- Hardware Type: <BLANK>
- [ ] run `make test` to ensure correctness
- [x ] run `make checkstyle` to ensure code style
- [ ] run `make test-convergence` to ensure convergence

PR #519, opened 2025-01-09
Fix mean subtraction in layer norm kernels

nhamanasu:
## Summary
This PR is the possible proposal to the Issue: https://github.com/linkedin/Liger-Kernel/issues/518

I've confirmed neither that the issue is really a issue nor that this PR can pass tests, so made this as a draft PR.

I'd like someone interested in this to check them please... 🙇

Issue #518, opened 2025-01-09
For better numerical accuracy in LayerNorm


nhamanasu:
### 🚀 The feature, motivation and pitch

In the implementation of LayerNorm, I wonder the `var` computation may result in a wrong value when `BLOCK_SIZE` is not the same as the feature dimension (in other words, `mask` has some False elements).

This intuition is because, the elements where `mask == True` will be `-mean` instead of `0.0`, and `var` just takes the summation including them too.

I think we should properly mask those positions with 0.0 by using `tl.where` or something.

https://github.com/linkedin/Liger-Kernel/blob/134a13e8f0ef59a98e046a61a85197902b958cea/src/liger_kernel/ops/layer_norm.py#L60

ByronHsu:
in what case BLOCK_SIZE != feature dim?

nhamanasu:
Thank you for the comment. Please see the discussion in https://github.com/linkedin/Liger-Kernel/pull/519

To put it simply, BLOCK_SIZE must be the power of 2 in triton.
So we can make arbitrary situations where BLOCK_SIZE != feature dim e.g., feat_dim == 768, 1536, whatever.

PR #815, opened 2025-07-17
Layernorm enhancement

Manan17:
## Summary

This is an enhancement to the layer_norm kernel. Majorly the backward kernel.
Changes in forward:
1. The new version performs all computations in float32 regardless of input dtype, then converts back only once at the end. This prevents precision loss that can accumulate with float16/bfloat16.
2. The new version pre-calculates row pointers instead of modifying the base pointers. This is cleaner and potentially more efficient for the compiler.
3. Eliminates the tl.where masking operation (relies on loading with other=0.0)

Changes in backward:
1. The original processes multiple rows per thread block in a loop, while the optimized version processes exactly one row per thread block.
2. The original uses local accumulation then stores to separate locations, while the optimized uses atomic `tl.atomic_add` operations to accumulate directly to the final gradient tensors.
3. The original requires a separate reduction step in Python, while the optimized kernel produces final gradients directly.

Here are the benchmark results for speed:
Before:
Backward pass- 
layer_norm_speed_backward

After:
Backward pass - 
layer_norm_speed_backward_new

Before: 
Full - 
layer_norm_speed_full

After:
Full-
layer_norm_speed_full_new

This layer norm is 27% faster than Huggingface's implementation.
The benchmarks for forwards pass (speed) and full pass (memory) are the same.

## Testing Done

Ran:
python -m pytest test/transformers/test_layer_norm.py

- Hardware Type: NVIDIA H100 80GB HBM3


src/liger_kernel/ops/layer_norm.py
```diff
@@ -263,3 +297,17 @@ def backward(ctx, dY):
         X, W, B, Mean, RSTD = ctx.saved_tensors
         DX, DW, DB = layer_norm_backward(dY, X, W, B, Mean, RSTD)
         return DX, DW, DB, None
+
+
+def liger_layer_norm(X, W, B, eps=1e-5):
```
vaibhavjindal: Looks like this is already defined in `Liger-Kernel/src/liger_kernel/transformers/functional.py`, we don't need it here.

vaibhavjindal:
Since you are changing the core logic of backward pass, you should also check the impact on speed and memory when you increase the number of rows for different values of hidden sizes.

Manan17:
> Since you are changing the core logic of backward pass, you should also check the impact on speed and memory when you increase the number of rows for different values of hidden sizes.

I tried increasing the number of rows from 4096 to 8192 and 16384. Got the same results. 
Attaching benchmarks for these:
Row Size: 8192 -
layer_norm_speed_backward_8192
Row Size: 16384 - 
layer_norm_speed_backward_16384

src/liger_kernel/ops/layer_norm.py
```diff
@@ -47,26 +46,39 @@ def _layer_norm_forward_kernel(
...
+
+    # Load input data and convert to fp32 for numerical stability
+    X_row = tl.load(row_X_ptr + col_offsets, mask=mask, other=0.0)
+    X_f32 = X_row.to(tl.float32)
+
+    # Compute statistics in fp32 for numerical stability
+    n_cols_f32 = n_cols.to(tl.float32)
+    mean = tl.sum(X_f32, axis=0) / n_cols_f32
+    X_centered = X_f32 - mean
+    var = tl.sum(X_centered * X_centered, axis=0) / n_cols_f32
```
Tcc0403: Wouldn't it cause the same issue as #582 trying to fix? Related issue/PR: #518, #519

`X_centered` would contain `-mean` where the elements are out of bound, resulting in the incorrect `var`. 

Adding (1, 1, 1500) to the unit test can reproduce the issue.

Tcc0403: But the issue only occurs when `hidden_size` isn't power of 2, which is no the common case.

Should we address this kind of edge cases? @vaibhavjindal @qingquansong

Manan17: It has been now solved, the efficiency remains the same (optimized).

src/liger_kernel/ops/layer_norm.py
```diff
@@ -47,26 +46,41 @@ def _layer_norm_forward_kernel(
     col_offsets = tl.arange(0, BLOCK_SIZE)
```
vaibhavjindal, line 46: Use int64 conversion as mentioned here: https://github.com/linkedin/Liger-Kernel/issues/803, https://github.com/linkedin/Liger-Kernel/pull/804

Manan17: Done

src/liger_kernel/ops/layer_norm.py
```diff
@@ -43,30 +42,45 @@ def _layer_norm_forward_kernel(
...
+    X_row = tl.load(row_X_ptr + col_offsets, mask=mask, other=0.0)
+    X_f32 = X_row.to(tl.float32)
+
+    # Compute statistics in fp32 for numerical stability
+    n_cols_f32 = n_cols.to(tl.float32)
+    mean = tl.sum(X_f32, axis=0) / n_cols_f32
+    X_centered = X_f32 - mean
+    # Apply mask to variance calculation to exclude contributions from masked elements
+    X_centered_masked = tl.where(mask, X_centered, 0.0)
+    var = tl.sum(X_centered_masked * X_centered_masked, axis=0) / n_cols_f32
```
shimizust, line 71: Should there be a test case added for bf16 to test this logic?

shimizust (dismissed):
lgtm, thanks for figuring out this optimization! Just had a comment on the test

src/liger_kernel/ops/layer_norm.py
Manan17: Thanks for pointing that out. tl.atomic_add does not support bfloat16. Have to convert gradients and weights to fp32 for it, have added the test and no changes in the efficiency.

shimizust (approved):

Issue #827, opened 2025-07-21
fused_linear_cross_entropy caused bad performance


waynechu1021:
### 🐛 Describe the bug

Hello, i'm finetuning Qwen-2.5-VL-3B with trl. When i turn on the liger kernel, it could lead to poor performance while the performance is normal without liger kernel.  I later discovered that it was `fused_linear_cross_entropy` that caused the performance difference. However, their loss looks almost the same, and the problem is that this option is crucial for saving GPU memory.
<img width="1755" height="470" alt="Image" src="https://github.com/user-attachments/assets/2607a954-8b4c-47ef-b291-64bd49c0e4bf" />

Here is my training args
deepspeed --master_port 25410  src/open_r1/sft.py \
    --model_name_or_path .cache/Qwen2.5-VL-3B-Instruct \
    --dataset_name data/instructions_dynamic_combine_action.jsonl \
    --deepspeed scripts/zero3.json \
    --learning_rate 2.0e-5 \
    --num_train_epochs 1 \
    --packing \
    --max_seq_length 64800 \
    --per_device_train_batch_size 4 \
    --gradient_accumulation_steps 4 \
    --dataloader_num_workers 4 \
    --gradient_checkpointing True \
    --torch_dtype bfloat16 \
    --bf16 \
    --logging_steps 5 \
    --eval_strategy no \
    --eval_steps 100 \
    --save_strategy no \
    --save_steps 6000 \
    --attn_implementation flash_attention_2 \
    --output_dir result/Qwen2.5-VL_3B_sft_r2r_dynamic_test_fix_liger \
    --report_to tensorboard

### Versions

Environment Report:
-------------------
Operating System: Linux-6.8.0-54-generic-x86_64-with-glibc2.39
Python version: 3.10.16
Liger Kernel version: 0.5.10
PyTorch version: 2.7.1+cu126
CUDA version: 12.6
HIP(ROCm) version: Not available
Triton version: 3.3.1
Transformers version: 4.50.3
Trl version: 0.16.0
XPU version: XPU Not Available

Issue #512, opened 2025-01-04
`LigerFusedLinearCrossEntropyLoss` Causes Training Loss to Diverge After Reaching ~8


penghui-yang:
#### **Description**  
When using `LigerFusedLinearCrossEntropyLoss` (Liger FLCE) from the Liger kernel to replace `torch.nn.CrossEntropyLoss`, the training loss becomes unstable and diverges after reaching a certain value (~8). In contrast, the loss computed using `torch.nn.CrossEntropyLoss` continues to decrease smoothly.

#### **Expected Behavior**  
The loss computed with `LigerFusedLinearCrossEntropyLoss` should decrease similarly to `torch.nn.CrossEntropyLoss` without significant oscillations or divergence.

#### **Observed Behavior**  
- During the initial training phase, both loss functions exhibit similar behavior, and the loss decreases as expected.
- When the loss computed with `LigerFusedLinearCrossEntropyLoss` reaches ~8, it becomes unstable, oscillates, and diverges, as shown in the attached graph.

#### **Screenshots/Logs**  
Loss curve comparison (attached):  
- The orange curve shows the behavior with `torch.nn.CrossEntropyLoss` (stable).  
- The purple curve shows the behavior with `LigerFusedLinearCrossEntropyLoss` (unstable and divergent).  

![image](https://github.com/user-attachments/assets/9133ce97-3553-4336-8b9c-60a45fb7e9e9)

#### **Additional Context**  
- This issue appears to be related to gradient computation or numerical stability with `LigerFusedLinearCrossEntropyLoss`.
- No hyperparameter changes were made between the two implementations.

#### **Request for Assistance**  
- Please investigate whether there are implementation issues with `LigerFusedLinearCrossEntropyLoss`.  
- Are there additional configurations or training parameters required to avoid instability?  

Thank you for your assistance!

#### **Code to Reproduce**  

**Original `compute_loss` implementation (works as expected):**
```python
def compute_loss(self, hidden_states, labels):
    logits = self.lm_head(hidden_states).float()
    # Using torch.nn.CrossEntropyLoss for loss computation
    loss_fn = torch.nn.CrossEntropyLoss()
    loss = loss_fn(logits[:, :-1].reshape(-1, logits.size(-1)), labels[:, 1:].reshape(-1))
    return loss
```

**New `compute_fused_loss` implementation (causes instability):**
```python
def compute_fused_loss(self, hidden_states, labels):
[...]

PR #830, opened 2025-07-22
Enforce gradients are accumulated in fp32

Tcc0403:
## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/ops/fused_linear_cross_entropy.py
+++ b/src/liger_kernel/ops/fused_linear_cross_entropy.py
@@ -44,10 +44,11 @@ def fused_linear_cross_entropy_forward(
     chunk_size = triton.next_power_of_2(triton.cdiv(BT, inc_factor))  # (BT + inc_factor - 1) // inc_factor
     num_chunks = triton.cdiv(BT, chunk_size)  # (BT + chunk_size - 1) // chunk_size
 
-    grad_weight = torch.zeros_like(weight, device=device) if weight.requires_grad else None
     grad_input = torch.zeros_like(_input, device=device)
-    grad_bias = torch.zeros_like(bias, device=device) if bias is not None else None
-    # we use fp32 for loss accumulator
+
+    # we use fp32 for loss and gradients accumulator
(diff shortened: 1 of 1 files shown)


Tcc0403:
@qingquansong @vaibhavjindal @shimizust @lancerts 
 
#512, #827 show that training with liger flce may lead to worse performance or unstable/divergent training.
 
In mixed precision training, accumulating intermediate results in fp32 is a common trick to prevent underflow/overflow issue. 
image

However, in current liger flce, gradients of weight for linear layer are accumlated in bf16 chunk by chunk.

Allocating fp32 buffer instead of the original precision for accumlations may solve the issue, but at the cost of larger memory footprints. Besides, the benchmarking shows that the changes can also improve the speed in larger batch_size * seq_len.

There are two default values for `accum_dtype`. I'm not sure which one is more preferred:
1. `None` : Accumulating in original precision
	- matching the current behavior of flce, better backward compatibility
	- still suffering from mentioned issues
2. `torch.float32` : Accmulating in fp32
  	- more memory required
	- more stable training
	- causing potential oom for users if upgrading liger kernel without knowing the change


or any alternative solutions are welcome.

I'll cleanup this PR once we decide the interface.

src/liger_kernel/ops/fused_linear_cross_entropy.py
```diff
@@ -124,16 +131,7 @@ def fused_linear_cross_entropy_forward(
...
-                input=grad_weight,
-                mat1=logits_chunk.t().to(
-                    _input_chunk.dtype
-                ),  # In an autocast scenario without bias, differing logits_chunk data types will cause an addmm operation error.
-                mat2=_input_chunk,
-                out=grad_weight,
-                alpha=1.0,
-                beta=1.0,
-            )
+            grad_weight += torch.mm(grad_logits_chunk.t(), _input_chunk).float()
```
Tcc0403, line 134: Another option is performing matmul in fp32 to maintain precision, but slowing down the speed.

shimizust:
@Tcc0403 Thanks for raising, I think the current accumulation method should be the default, as it works for most use cases.  Folks can have the option of switching to fp32 accumulation

Issue #650, opened 2025-04-03
0.5.6 report bug


hrdxwandg:
### 🐛 Describe the bug

after upgrade version from 0.5.5 to 0.5.6， when sft llama3.1-8b use trl with DataCollatorForCompletionOnlyLM  padding_free=True, it has error as follows:
```
[rank0]:   File "xxx/python3.11/site-packages/transformers/trainer.py", line 2245, in train
[rank0]:     return inner_training_loop(
[rank0]:            ^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "xxx/python3.11/site-packages/transformers/trainer.py", line 2556, in _inner_training_loop
[rank0]:     tr_loss_step = self.training_step(model, inputs, num_items_in_batch)
[rank0]:                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "xxx/python3.11/site-packages/transformers/trainer.py", line 3718, in training_step
[rank0]:     loss = self.compute_loss(model, inputs, num_items_in_batch=num_items_in_batch)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "xxx/python3.11/site-packages/trl/trainer/sft_trainer.py", line 495, in compute_loss
[rank0]:     (loss, outputs) = super().compute_loss(
[rank0]:                       ^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "xxx/python3.11/site-packages/transformers/trainer.py", line 3783, in compute_loss
[rank0]:     outputs = model(**inputs)
[rank0]:               ^^^^^^^^^^^^^^^
[rank0]:   File "xxx/python3.11/site-packages/torch/nn/modules/module.py", line 1739, in _wrapped_call_impl
[rank0]:     return self._call_impl(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "xxx/python3.11/site-packages/torch/nn/modules/module.py", line 1750, in _call_impl
[rank0]:     return forward_call(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #832, opened 2025-07-22
fixed_fused_linear_cross_entropy should pass through kwargs

npuichigo:
## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/transformers/model/loss_utils.py
+++ b/src/liger_kernel/transformers/model/loss_utils.py
@@ -23,6 +23,7 @@ def fixed_fused_linear_cross_entropy(
         reduction=reduction,
         ignore_index=ignore_index,
         softcap=final_logit_softcapping,
+        **kwargs,
     )
     if reduction == "sum":
         loss = loss / num_items_in_batch


Tcc0403:
The reason why we don't want to pass kwargs is because `**kwargs` also contains FlashAttentionKwrags, resulting in issue #650. 
More detailed explanation https://github.com/linkedin/Liger-Kernel/pull/651#discussion_r2034055850

npuichigo:
So maybe we should pass through ce_weight.

Tcc0403:
Sure, it should be fine to pass through other arguments, such as `ce_weight`, `label_smoothing`, `lse_square_scale`, etc. 
However, they are not natively supported in hf transformers, I'm not sure about the namings for these liger flce specified kwargs.
https://github.com/huggingface/transformers/blob/947a37e8f5bc50bc0e9a77c0d16b038adcb056d0/src/transformers/utils/generic.py#L858

JuanPZuluaga:
Hello, I think, it is also necessary to add `kwargs` in:

https://github.com/linkedin/Liger-Kernel/blob/24416a4286c71fcb7b148f1536f57d58e4133eda/src/liger_kernel/transformers/functional.py#L70

I'm hitting a compilation error with torch.compile. 

Basically: 

```
def liger_fused_linear_cross_entropy(
    input,
    weight,
    target,
    bias=None,
    ce_weight=None,
    ignore_index: int = -100,
    lse_square_scale: float = 0.0,
    label_smoothing: float = 0.0,
    reduction: str = "mean",
    softcap: Optional[float] = None,
    return_z_loss: bool = False,
    accum_dtype=None,
    use_token_scaling: bool = False,
    return_token_accuracy: bool = False,
    **kwargs
):
```

fixes it.

PR #886, opened 2025-09-19
Update Dev Dependencies

### LINKED ISSUE #885: Linting Deps should be updated
While inspecting the `makefile` and `setup.py`, I noticed that for linting, flake8+isort+black is declared as linting dev dependencies. But in makefile, ruff commands are used. Should the deps be updated to only use ruff and remove linting deps? WDYT?

https://github.com/linkedin/Liger-Kernel/blob/02d27648fd407391a702ef9eb91cff6a26cdf450/setup.py#L36-L38

cc: @vaibhavjindal, @shimizust

ParagEkbote:
## Summary
Fixes #885

As described in the issue, this PR updates the linting deps to only include ruff. I've also removed mkdocs, as mkdocs-material pre-installs mkdocs as well. Could you please review?

cc: @vaibhavjindal, @shimizust

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,9 +33,7 @@ def get_optional_dependencies():
         "dev": [
             "transformers>=4.49.0",
             "matplotlib>=3.7.2",
-            "flake8>=4.0.1.1",
-            "black>=24.4.2",
-            "isort>=5.13.2",
+            "ruff.=0.12.0",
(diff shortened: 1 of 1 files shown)


vaibhavjindal:
Thanks for cleaning up the dev dependencies! 🚢

PR #888, opened 2025-09-22
fix(flce): add `shift_labels` as eval mode loss condition

Tcc0403:
## Summary

Fix https://github.com/snowflakedb/ArcticTraining/issues/266

In liger flce patch, there are two branches in model.forward

1. If the model is in training mode, liger calls fused linear cross entropy to skip logit materialization and only calculates loss.
2. If the model is in eval mode, liger actually calls the original forward method,
CrossEntropy(self.lm_head(hidden_states, labels), to track full logits and loss.

However, liger doesn't check the `shift_labels` in the [eval branch](https://github.com/linkedin/Liger-Kernel/blob/d1441a0de8ee93bb3d316ba10ed449e89607825a/src/liger_kernel/transformers/model/llama.py#L249-L257), leaving loss=None without performing the loss calculation when we only passes `shift_labels` in eval mode.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/test/utils.py
+++ b/test/utils.py
@@ -14,7 +14,9 @@ from typing import Tuple
 import numpy as np
 import torch
 import torch.nn as nn
+import transformers
 
+from packaging import version
 from tokenizers import AddedToken
 from tokenizers import Tokenizer
 from tokenizers.models import BPE
@@ -149,21 +151,20 @@ class MiniModelConfig:
 
 def simple_collate_fn(data: List[Dict[str, Any]]):
     """A basic collate function to use for DataLoader"""
+    batch = {}
 
     input_ids = torch.stack([torch.tensor(item["input_ids"]) for item in data])
     attention_mask = torch.stack([torch.tensor(item["attention_mask"]) for item in data])
     labels = input_ids.clone()
-    shift_labels = nn.functional.pad(labels, (0, 1), value=-100)
-    shift_labels = shift_labels[..., 1:].contiguous()
+    batch["input_ids"] = input_ids
+    batch["attention_mask"] = attention_mask
+    batch["labels"] = labels
+    if version.parse("4.54.1") <= version.parse(transformers.__version__):
+        shift_labels = nn.functional.pad(labels, (0, 1), value=-100)
+        shift_labels = shift_labels[..., 1:].contiguous()
+        batch["shift_labels"] = shift_labels
 
-    return BatchEncoding(
-        {
-            "input_ids": input_ids,
-            "attention_mask": attention_mask,
-            "labels": labels,
-            "shift_labels": shift_labels,
-        }
-    )
+    return BatchEncoding(batch)
 
 
 def multimodal_collate_fn(data: List[Dict[str, Any]]):
@@ -177,9 +178,10 @@ def multimodal_collate_fn(data: List[Dict[str, Any]]):
 
     labels = input_ids.clone()
     batch["labels"] = labels
-    shift_labels = nn.functional.pad(labels, (0, 1), value=-100)
-    shift_labels = shift_labels[..., 1:].contiguous()
-    batch["shift_labels"] = shift_labels
+    if version.parse("4.54.1") <= version.parse(transformers.__version__):
+        shift_labels = nn.functional.pad(labels, (0, 1), value=-100)
+        shift_labels = shift_labels[..., 1:].contiguous()
+        batch["shift_labels"] = shift_labels
 
     # Collate all other keys, e.g. pixel_values, attention_mask, image_grid_thw, etc
     for key in keys:


Tcc0403:
Not sure if we should add tests for `shift_labels` functionality.

sfc-gh-sbekman:
Thank you for fixing, @Tcc0403 

question: why doesn't it shard the logits/loss computation in the eval mode?

Tcc0403:
@sfc-gh-sbekman Thank you for leaving a comment! 
> question: why doesn't it shard the logits/loss computation in the eval mode?

I don't fully understand the quetsion. What do you mean by sharding the logits/loss computation? 
The current kernel is designed for loss computation on single device, by iteratively calculating logits/loss/gradient then flushing logits. Or you mean sharding logits across devices in sequence parallel fashion? I think it can be a good direction for future improvement.

sfc-gh-sbekman:
I was asking why in eval mode forward isn't doing the same as in training mode (other than grads)

Tcc0403:
My guess is, in early design, we just assume that people might want to retrieve full logits in eval mode as well, so only enabling fused linear cross entropy in training mode.

sfc-gh-sbekman:
Won't that be a problem during training if someone combines with eval, since eval now is more memory hungry than training, unless they adjust hparams to be different for eval to use a smaller batch size. But say someone uses a bs=1 and a very long seqlen, there is no way to make bs smaller.

I personally prefer not combining eval with training and do it separately, so not a problem for me, but I can see how it might be an issue for others who want to combine. In any case I'm not asking to change anything, I was just trying to understand why it was chosen to be different.

Tcc0403:
> Won't that be a problem during training if someone combines with eval, since eval now is more memory hungry than training, unless they adjust hparams to be different for eval to use a smaller batch size. But say someone uses a bs=1 and a very long seqlen, there is no way to make bs smaller.

That's a great point. Instead of checking eval/training mode to decide whether calling flce or not, do you think we should fully rely on the new added argument `skip_logits` to make the call?

sfc-gh-sbekman:
I struggled finding documentation for `skip_logits`, found it in the code:

https://github.com/linkedin/Liger-Kernel/blob/cf43acd90a84b1528f2f182b608820a6c0a8d262/src/liger_kernel/transformers/model/gemma3.py#L102-L103

How/where would a user set it?

But otherwise probably leaving the logic as it is plus allowing a user to override the default behavior for eval would probably be a good flexible solution (and documenting it of course).

Tcc0403:
For more context, see #703 

> How/where would a user set it?

It was introduced in #704, one way to use it is just adding it to model call, `output = model(**batch, skip_logits=True)`

> But otherwise probably leaving the logic as it is plus allowing a user to override the default behavior for eval would probably be a good flexible solution (and documenting it of course).

Right, I think `skip_logits` should be able to override it. What do you think? cc @shimizust @vaibhavjindal

sfc-gh-sbekman:
> It was introduced in https://github.com/linkedin/Liger-Kernel/pull/704, one way to use it is just adding it to model call, output = model(**batch, skip_logits=True)

That's not great because one has to change end-user code, which shouldn't be hardwired to liger-kernel (think HF Transformers/HF Accelerate), rather than just replacing `AutoModelForCausalLMLoss` with `LigerForCausalLMLoss` - could this flag be propagated to `LigerForCausalLMLoss`?

perhaps one of these ideas?
```
LigerForCausalLMLoss(...., always_flce=True) # if False, only train
LigerForCausalLMLoss(...., always_fused_linear_cross_entropy=True) # same as above
LigerForCausalLMLoss(...., fused_linear_cross_entropy=Union[bool, List["train", "eval"]]) # more explicit
```

the last suggestion reuses the existing `fused_linear_cross_entropy` flag and allows in addition to True/False, adding the explicit modes - train/eval. where for BC, `True == ["train"]`.

`skip_logits` is not a very well defined name to what it implies,  perhaps it's not too late to rename? besides it's not skipping logits, it just manifests slices of them - so this is a really an incorrect/misleading arg name.

sfc-gh-sbekman:
also, @Tcc0403, perhaps I misread the code, but last time I was reading the autograd code it looked to me that it unconditionally performed grad calculation even when doing eval in flce. Did I get it wrong?

Tcc0403:
> That's not great because one has to change end-user code, which shouldn't be hardwired to liger-kernel (think HF Transformers/HF Accelerate), rather than just replacing AutoModelForCausalLMLoss with LigerForCausalLMLoss - could this flag be propagated to LigerForCausalLMLoss?


The problem is that `AutoModelForCausalLMLoss` and `LigerForCausalLMLoss` are not interchangeable, one takes `logits` and the other takes `hidden_states` + `lm_head.weight`, so the flag has to be checked somewhere before calling lm_head matmul (to fuse linear and cross entropy layer). I haven't come up with another way to do it.

> skip_logits is not a very well defined name to what it implies, perhaps it's not too late to rename? besides it's not skipping logits, it just manifests slices of them - so this is a really an incorrect/misleading arg name.

I agree. The intention of this argument is to run flce in both training and eval mode (previously in training mode only), while keeping a fall-back mechanism to generate logits for prediction/evaluation. To keep it simple, maybe just add an fused_linear_cross_entropy attribute in hf models, so we can easily set/unset it? or maybe a context manager?

> also, @Tcc0403, perhaps I misread the code, but last time I was reading the autograd code it looked to me that it unconditionally performed grad calculation even when doing eval in flce. Did I get it wrong?

You are correct, flce always calculates gradients. We just don't call flce when doing eval at the current time.

sfc-gh-sbekman:
oops, my apologies, I meant to write `AutoLigerKernelForCausalLM` (not `...Loss`). I think this flag belongs to `from_pretrained` and not inside `model()` call. Is this an issue of propagating it to where it's needed from `from_pretrained`?

I wonder if we can get HF Transformers to adopt ALST version of flce. It's pure pytorch and it performs about the same as flce (slightly slower and a tad more memory, but a smaller shard leads to the same mem usage) and it already skips grad calculation if it detects it's in eval mode.
https://github.com/deepspeedai/DeepSpeed/blob/16c1bf429f10db3e681b49fac37d641909aad69d/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L898
That way perhaps they could have a standard flag for `from_pretrained` and then they and liger-kernel could both rely on it?

> We just don't call flce when doing eval at the current time.

Hasn't this changed since https://github.com/linkedin/Liger-Kernel/pull/704? So before you unleash this new flag flce needs to become eval-aware and skip grad calculation in `forward` - please correct me if I'm wrong.

Tcc0403:
Yeah, propagating it to model factory argument seems reasonable. @shimizust @vaibhavjindal WDYT?

> I wonder if we can get HF Transformers to adopt ALST version of flce. It's pure pytorch and it performs about the same as flce (slightly slower and a tad more memory, but a smaller shard leads to the same mem usage) and it already skips grad calculation if it detects it's in eval mode.
> [deepspeedai/DeepSpeed@16c1bf4/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L898](https://github.com/deepspeedai/DeepSpeed/blob/16c1bf429f10db3e681b49fac37d641909aad69d/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L898)
> That way perhaps they could have a standard flag for from_pretrained and then they and liger-kernel could both rely on it?
> 

+1 on a standard flag. Actually, I can modify flce a bit to make it skip grad calculation in eval mode as well, then we can have both versions behave the same.

> > We just don't call flce when doing eval at the current time.
> 
> Hasn't this changed since https://github.com/linkedin/Liger-Kernel/pull/704? So before you unleash this new flag flce needs to become eval-aware and skip grad calculation in forward - please correct me if I'm wrong.

No, flce always does grad calculation in forward. TL;DR

<table>
<tr>
 <td> 
 <td colspan=2> training mode </td>
 <td colspan=2> eval mode</td>
<tr>
 <td>
 <td> FLCE call
 <td> grad calc
 <td> FLCE call
 <td> grad calc
<tr>
 <td> Before #704
 <td> Y
 <td> Y
 <td> N
 <td> N
<tr>
 <td> After #704 
 <td> Y
 <td> Y
 <td> Y if skip_logits else N
 <td> Y if skip_logits else N
</table>

sfc-gh-sbekman:
aha, thank you for clarifying, the situation with grad calculation - in other words #704 needs more follow up work - thank you for volunteering to remove that wasteful overhead, @Tcc0403 

re: integrating this feature into Transformers - would you like to drive that initiative since you guys invented this idea and feel free to re-use the pure pytorch version from Deepspeed/ALST to have a fallback for HF models that aren't supported by Liger-Kernel - that way I think HF would be more interested in making this a built-in feature, since they would want all models covered. I'm surprised they haven't done this already as it's a huge memory saver. And I'm happy to help with driving this proposal. Please let me know what you prefer.

Tcc0403:
> re: integrating this feature into Transformers - would you like to drive that initiative since you guys invented this idea and feel free to re-use the pure pytorch version from Deepspeed/ALST to have a fallback for HF models that aren't supported by Liger-Kernel - that way I think HF would be more interested in making this a built-in feature, since they would want all models covered. I'm surprised they haven't done this already as it's a huge memory saver. And I'm happy to help with driving this proposal. Please let me know what you prefer.

Sure, I'm happy to help! I have no experience of making such proposals, could you guide me through the process?

sfc-gh-sbekman:
One way to approach it is to show a memory profile w/ and w/o flce using some largish bs or seqlen (and hidden_size), where resulting logits are huge. And argue that HF Transformers users will greatly benefit from this feature being built-in as a core feature.

Then they could try and import liger-kernel (soft dependency) and check if the model that is being loaded is supported and then use it. And if either it's not supported or the user has no liger-kernel installed, fall back onto pure pytorch implementation I shared.

The trigger could be `from_pretrained(..., tiled_cross_entropy=True)` - so a more generic name which could use various backends. Or even it could be:
```
backend = "liger-kernel"  # or "deepspeed", or possibly other implementations
from_pretrained(..., tiled_cross_entropy=backend)
```
then a user can control what they want. One possibility:
```
backend = ["liger-kernel" | "deepspeed" | "auto"]
```
where `auto` will automatically try to use the most efficient solution if such available (flce) and fallback to the slower pure pytorch implementation, but still performing the same memory savings.

Is that helpful?

You can also borrow my pytorch mem profiler plots:
https://x.com/StasBekman/status/1940455219684692177
https://x.com/StasBekman/status/1942973445064798521
they are also in the paper: https://arxiv.org/abs/2506.13996 to which you can also lean as it has some amazing numbers partially thanks to tiling computation. We use liger-kernel everywhere in the paper evaluations.

Actually, why doesn't liger-kernel do tiled-mlp? that's another huge memory saving for long seq/large bs. you can probably take our pure pytorch TiledMLP and make a more efficient kernel for it. https://github.com/deepspeedai/DeepSpeed/blob/16c1bf429f10db3e681b49fac37d641909aad69d/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L898

Tcc0403:
Thank you for the advice! I'll shoot my proposal in a few days.

> Actually, why doesn't liger-kernel do tiled-mlp? that's another huge memory saving for long seq/large bs. you can probably take our pure pytorch TiledMLP and make a more efficient kernel for it. [deepspeedai/DeepSpeed@16c1bf4/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L898](https://github.com/deepspeedai/DeepSpeed/blob/16c1bf429f10db3e681b49fac37d641909aad69d/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L898)

Sure! Let me see if I can write a more efficient version.

sfc-gh-sbekman:
Fantastic! Thank you, @Tcc0403

Tcc0403:
@sfc-gh-sbekman Here are my proposal in transformers https://github.com/huggingface/transformers/issues/41306 and no-grad-forward for flce https://github.com/linkedin/Liger-Kernel/pull/894 

Please take a look and tell me if there's any issue!

sfc-gh-sbekman:
The proposal looks great, I added additional notes wrt to grad reduction nuances. e.g. DDP it troublesome. and there is an improved version in Axolotl that supports FSDP as well.

the no-grad-forward looks good as well

Thank you very much, @Tcc0403

momochen:
> > It was introduced in #704, one way to use it is just adding it to model call, output = model(**batch, skip_logits=True)
> 
> That's not great because one has to change end-user code, which shouldn't be hardwired to liger-kernel (think HF Transformers/HF Accelerate), rather than just replacing `AutoModelForCausalLMLoss` with `LigerForCausalLMLoss` - could this flag be propagated to `LigerForCausalLMLoss`?
> 
> perhaps one of these ideas?
> 
> ```
> LigerForCausalLMLoss(...., always_flce=True) # if False, only train
> LigerForCausalLMLoss(...., always_fused_linear_cross_entropy=True) # same as above
> LigerForCausalLMLoss(...., fused_linear_cross_entropy=Union[bool, List["train", "eval"]]) # more explicit
> ```
> 
> the last suggestion reuses the existing `fused_linear_cross_entropy` flag and allows in addition to True/False, adding the explicit modes - train/eval. where for BC, `True == ["train"]`.
> 
> `skip_logits` is not a very well defined name to what it implies, perhaps it's not too late to rename? besides it's not skipping logits, it just manifests slices of them - so this is a really an incorrect/misleading arg name.

+1 on setting skip_logits/tiled_loss(or other better explicit names) in LigerForCausalLMLoss instead of in model eval as it's Liger specific thing.

momochen:
> Yeah, propagating it to model factory argument seems reasonable. @shimizust @vaibhavjindal WDYT?
> 
> > I wonder if we can get HF Transformers to adopt ALST version of flce. It's pure pytorch and it performs about the same as flce (slightly slower and a tad more memory, but a smaller shard leads to the same mem usage) and it already skips grad calculation if it detects it's in eval mode.
> > [deepspeedai/DeepSpeed@16c1bf4/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L898](https://github.com/deepspeedai/DeepSpeed/blob/16c1bf429f10db3e681b49fac37d641909aad69d/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L898)
> > That way perhaps they could have a standard flag for from_pretrained and then they and liger-kernel could both rely on it?
> 
> +1 on a standard flag. Actually, I can modify flce a bit to make it skip grad calculation in eval mode as well, then we can have both versions behave the same.
> 
> > > We just don't call flce when doing eval at the current time.
> > 
> > 
> > Hasn't this changed since #704? So before you unleash this new flag flce needs to become eval-aware and skip grad calculation in forward - please correct me if I'm wrong.
> 
> No, flce always does grad calculation in forward. TL;DR
> 
> training mode	eval mode
> FLCE call	grad calc	FLCE call	grad calc
> Before #704	Y	Y	N	N
> After #704	Y	Y	Y if skip_logits else N	Y if skip_logits else N

LGTM. Let's put it as parameter in model factory init.

momochen:
> Not sure if we should add tests for `shift_labels` functionality.

Yes please add simple tests on this shift_labels flag, at least make sure output is different from shift_labels = None case

momochen (commented):
lgtm overall; please add some simple tests on validating results with and without shift_labels, thanks!

Tcc0403:
> Yes please add simple tests on this shift_labels flag, at least make sure output is different from shift_labels = None case

@momochen I've added shift_labels calculation to collate function, so all convergence tests are running with shift_labels. I'm not sure if it satisfy your request, but I feel that it doesn't make sense to add another training loops just for testing shift_labels. Since shift_labels is already a built-in feature in transformers, I think simply comparing liger and transformers should be enough. 

If you wish to have both w/ and w/o shift_labels in test, one solution is to make first half loops run with shift_labels and second half without.

momochen:
> > Yes please add simple tests on this shift_labels flag, at least make sure output is different from shift_labels = None case
> 
> @momochen I've added shift_labels calculation to collate function, so all convergence tests are running with shift_labels. I'm not sure if it satisfy your request, but I feel that it doesn't make sense to add another training loops just for testing shift_labels. Since shift_labels is already a built-in feature in transformers, I think simply comparing liger and transformers should be enough.
> 
> If you wish to have both w/ and w/o shift_labels in test, one solution is to make first half loops run with shift_labels and second half without.

Got you. Then we don't need to add separate tests.

sfc-gh-sbekman:
a new release would be lovely to be able to anchor on this fix in frameworks. Thank you!

PR #894, opened 2025-10-01
feat(ce,-flce): decouple gradients computation for no_grad mode

Tcc0403:
## Testing Done

- Hardware Type: <BLANK>


Tcc0403:
@shimizust The slower forward pass is kinda expected because: 
1. instead of one big matmul, we slice input rowwise and do multiple matmuls. It brings some kernel launch overhead. Plus, it's not gauranteed to ivnoke the most efficient kernels (tiling size, tail effect, etc... from kernel's prespective).
2. multiple cross entropy kernel launch overhead and similar issue also add up.

I just found that we can remove this line in eval mode as well, cutting another matmul for each interation should be significant.
https://github.com/linkedin/Liger-Kernel/blob/5c2a04df9e7a332fcafbc3e2d5481d0e0207f586/src/liger_kernel/ops/fused_linear_cross_entropy.py#L175

and some grad tensors allocations too

PR #895, opened 2025-10-01
Option to return hard and soft loss when using distillation

h-aurelien-lac:
## Summary
Proposition to add `return_soft_hard_loss` parameter to enable logging of soft and hard losses separately. Useful for monitoring and analysis during training

## Testing Done

- [x] test_jsd_loss.py
- [x] test_cosine_loss.py


h-aurelien-lac:
@shimizust, can I have a round of reviews please 🤗

PR #898, opened 2025-10-08
fix(flce): Remove unused instances in benchmarking script

Tcc0403:
## Summary

Before this PR, we would create multiple instances for all providers despite most instances are touched except the target one, which can lead to extremely long startup time and potential OOM.

## Testing Done

- Hardware Type: <BLANK>


Tcc0403:
Maybe the fix can be applied to other scrips as well to speed up benchmarking?

Tcc0403:
Move to #894

PR #900, opened 2025-10-09
Add support of Falcon-H1 models for liger kernels

puneeshkhanna:
## Summary
This PR enables ROPE, rmsnorm and cross entropy liger kernels for Falcon-H1 models

Falcon H1 models -
https://huggingface.co/collections/tiiuae/falcon-h1-6819f2795bc406da60fab8df

Support for fused cross entropy and swiglu to be added later.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/transformers/monkey_patch.py
+++ b/src/liger_kernel/transformers/monkey_patch.py
@@ -2147,14 +2147,10 @@ def apply_liger_kernel_to_falcon_h1(
         raise NotImplementedError("LigerSwiGLUMLP is not available for Falcon-H1 models.")
 
     if cross_entropy:
-        if transformer_version >= version.parse(SUPPORTED_TRANSFORMER_VERSION):
-            logger.info("Apply liger cross entropy")
-            from transformers.loss.loss_utils import nn
+        logger.info("Apply liger cross entropy")
+        from transformers.loss.loss_utils import nn
 
-            nn.functional.cross_entropy = liger_cross_entropy
-        else:
-            logger.warning(TRANSFORMER_DEPRECATION_WARNING)
-            modeling_falcon_h1.CrossEntropyLoss = LigerCrossEntropyLoss
+        nn.functional.cross_entropy = liger_cross_entropy
 
     if fused_linear_cross_entropy:
         raise NotImplementedError("LigerFusedLinearCrossEntropy is not available for Falcon-H1 models.")


puneeshkhanna:
@shimizust - Can you review this new PR changes up till now for support of Falcon-H1. I closed the earlier PR. I haven't yet added the testing changes. I will add them by tomorrow and push.

puneeshkhanna:
@shimizust  - for tests, I have to add changes in Liger-Kernel/test/convergence/bf16/test_mini_models_with_logits.py and Liger-Kernel/test/convergence/bf16/test_mini_models.py right and same files in fp32 ? Anywhere else ?

puneeshkhanna:
@shimizust  - Please note that Falcon H1 is hybrid architecture having Mamba as well as attention blocks. Does the testing infrastructure  support Mamba architecture and dependencies ? Is it possible to do it as separate PR too ? Right now, it is more important for the community to leverage liger kernels for Falcon-H1 models.

Tcc0403:
For testings, there is also `test_apply_liger_kernel_to_instance_for_MODEL()` to ensure modules are successfully patched in `test/transformers/test_monkey_patch.py`

So there should be 2 types of testings to support models:
1. checking patching works on instance -- `test_apply_liger_kernels_for_MODEL()` as above mentioned.
2. convergence test
	- test_mini_model.py -- mainly for testing fused_linear_cross_entropy support
	- test_mini_model_with_logits.py -- without fused_linear_cross_entropy patch
	- test_mini_model_multimodal.py -- testing ForConditionalGenerationLM with multimodal input

src/liger_kernel/transformers/monkey_patch.py
```diff
@@ -2107,6 +2107,76 @@ def apply_liger_kernel_to_internvl(
...
+
+    if cross_entropy:
+        if transformer_version >= version.parse(SUPPORTED_TRANSFORMER_VERSION):
+            logger.info("Apply liger cross entropy")
+            from transformers.loss.loss_utils import nn
+
+            nn.functional.cross_entropy = liger_cross_entropy
+        else:
+            logger.warning(TRANSFORMER_DEPRECATION_WARNING)
+            modeling_falcon_h1.CrossEntropyLoss = LigerCrossEntropyLoss
```
Tcc0403: Falcon-H1 is supported in transformers>=4.53.0 and there's no `CrossEntropyLoss` in modeling file. Simply patching `nn.functional.cross_entropy` should be enough.

puneeshkhanna: Yes right so no need of else condition

puneeshkhanna: Pushed  - please check

puneeshkhanna:
@Tcc0403 - Do you want all the testing changes to be raised in same PR or can we do it in a separate PR because there are decent amount of changes and I need to test locally at our end to ensure that the tests are passing etc ? Also we need to take care of mamba block dependencies. This may take some time. However we use the pushed changes in our SFT code base and enabling liger kernels improves speed and memory efficiencies for fine tuning H1 models and loss plot matches with and without liger kernels.

Tcc0403:
The modules we patched are independent to mamba, so I think as long as there's no mismatch between both curves, it's safe to say the patch doesn't have side effects.

For testings, ideally we want to include them in the same PR. But if you really want the support on main branch asap, I can help with that, including the fused linear cross entropy support, either directly editing your PR or I'll open one after. Let me what you prefer.

puneeshkhanna:
@Tcc0403  - I am 100% sure that there is no difference between both curves and we use internally. It would be great if you can merge this PR and raise a new PR for fused linear cross entropy support too for the community and please add me to the PR. 
image
And then I will have below pending (we l add them in coming weeks):
1. Addition of test cases - will be separate PR 
2. Also support Swiglu MLP kernel but since we have layer multipliers and MUP, I want to check that properly and then integrate - will be separate PR.

Tcc0403:
There will be follow-up PRs to add testings and flce, mlp support

puneeshkhanna:
@Tcc0403 - Just a doubt that do you see a possibility to incorporate below implementation with gate and down multipliers for Swiglu liger kernels too ?

```
class FalconH1MLP(nn.Module):
    def __init__(self, config: FalconH1Config):
        super().__init__()
        self.config = config
        self.hidden_size = config.hidden_size
        self.intermediate_size = config.intermediate_size
        self.gate_proj = nn.Linear(self.hidden_size, self.intermediate_size, bias=config.mlp_bias)
        self.up_proj = nn.Linear(self.hidden_size, self.intermediate_size, bias=config.mlp_bias)
        self.down_proj = nn.Linear(self.intermediate_size, self.hidden_size, bias=config.mlp_bias)
        self.act_fn = ACT2FN[config.hidden_act]
        self.gate_multiplier, self.down_multiplier = config.mlp_multipliers

    def forward(self, x):
        y = self.up_proj(x) * self.act_fn(self.gate_proj(x) * self.gate_multiplier)
        y = self.down_proj(y) * self.down_multiplier
        return y
```

Tcc0403:
Yes, multipliers seem easy to support. Can you file an issue for feature request?

PR #901, opened 2025-10-10
Add PolyNorm operator

### LINKED ISSUE #865: Polynorm Kernel

Recently PolyNorm is getting traction as an RMSNorm alternative. Recent Motif2.6B model used this and also released kernel for AMD. It would be  great addition to the library. 

https://arxiv.org/abs/2508.09148

https://arxiv.org/pdf/2411.03884v1

### Alternatives

_No response_

### Additional context

_No response_

0xtoward:
Resolve #865

- Add LigerPolyNormFunction
- Add Accuracy Test and Perfomance Benchmark

Performance improvement vs PyTorch autograd:
- Speed: 12-40x faster
- Memory: 6.4x reduction (10x with in-place)

Reference: 
https://github.com/BryceZhuo/PolyCom/
https://arxiv.org/abs/2508.09148
https://arxiv.org/pdf/2411.03884v1

## Summary

Introduce **PolyNorm** (Type-II PolyCom, fix r=3) as a training-side fused operator for Liger-Kernel.

image

**Kernels:** `LigerPolyNormFunction` (Triton) with **fwd+bwd fused**.
  - **Forward:** per row compute `mean(x^2)`, `mean(x^4)`, `mean(x^6)` → cache `rstd_p` (p=1,2,3; 3 scalars/row); write `y` once (no intermediate tensors).
  - **Backward (closed-form):** per row compute `S_p = ⟨grad, x^p⟩`, then  
    dX = Σ_{p=1..3} w_p * [ p * x^{p-1} * rstd_p * grad  −  (p/H) * x^{2p-1} * (rstd_p)^3 * S_p ]  
    dW_p = rstd_p * S_p,  
    dB = sum(grad).  

## Testing Done

Performance improvement vs PyTorch autograd:
- Speed: 12-40x faster
- Memory: 6.4x reduction (10x with in-place)

- Hardware Type: RTX4070

### DIFF
--- a/src/liger_kernel/ops/poly_norm.py
+++ b/src/liger_kernel/ops/poly_norm.py
@@ -296,7 +296,7 @@ def poly_norm_backward(dY, X, W, RSTD, BLOCK_SIZE, num_warps, in_place):
         dX = dY
     else:
         dX = torch.zeros_like(dY)
-    
+
     _dW = torch.empty((sm_count, 3), dtype=torch.float32, device=W.device)
     _dB = torch.empty((sm_count,), dtype=torch.float32, device=W.device)
 
@@ -382,7 +382,5 @@ class LigerPolyNormFunction(torch.autograd.Function):
             dX, dW, dB: gradients w.r.t. X, W, B
         """
         X, W, RSTD = ctx.saved_tensors
-        dX, dW, dB = poly_norm_backward(
-            grad_output, X, W, RSTD, ctx.BLOCK_SIZE, ctx.num_warps, ctx.in_place
-        )
+        dX, dW, dB = poly_norm_backward(grad_output, X, W, RSTD, ctx.BLOCK_SIZE, ctx.num_warps, ctx.in_place)
         return dX, dW, dB, None, None


src/liger_kernel/transformers/poly_norm.py
```diff
@@ -0,0 +1,37 @@
...
+    Reference:
+        https://github.com/BryceZhuo/PolyCom/
+
+    Args:
+        eps: epsilon for numerical stability (default: 1e-6)
+    """
+
+    def __init__(self, eps=1e-6):
+        super().__init__()
+        self.weight = nn.Parameter(torch.tensor([0.3, 0.4, 0.3]))
```
Tcc0403: https://github.com/BryceZhuo/PolyCom/blob/0ed8ad460f1e54eb1817163a9aecdaf7c3887643/README.md?plain=1#L143

The original implementation seems to initailize weight with (1/3, 1/3, 1/3), any considerations why it's (0.3, 0.4, 0.3) here?

src/liger_kernel/transformers/poly_norm.py
```diff
@@ -0,0 +1,37 @@
...
+        self.bias = nn.Parameter(torch.tensor(0.1))
```
Tcc0403: ditto. Why is it 0.1 instead of 1?

0xtoward: Thanks for your review! I just pushed
1. init aligned
2. dW fix: removed a redundant step. (backward speedup 1.5x than previous)
3. in_place option added (mirrors RMSNorm) (mem saved 2x than previous)

Also I tried ncu and it shows mem-bound. I tried re-computation but it hurts the performance on my machine
image

Tcc0403: >  I tried re-computation but it hurts the performance on my machine

Which part did you try replacing by recomputing? rstd cache?

0xtoward: Yeah I was stupid. Also tried increasing mem reading to overlay mem-compute steps which apparently hampers perf. 
The main mem cost is reading X (unavoidable), and since PolyNorm has only 3 scalar weights vs RMSNorm's element-wise gamma, the current implementation may be already well-optimized I guess.

Tcc0403:
Benchmarking against torch compile on H100

### speed
#### forward
download

#### backward
download

#### full
download

### memory
#### full
download

0xtoward:
> Thank you! Comments and docstring are really nice! Currently there's a formatting issue. After `make checkstyle`, feel free to ping me then I'll merge it!

Thanks for the review! Format fixed.

PR #903, opened 2025-10-10
Add flce forward for FalconH1ForCausalLM and missing tests

Tcc0403:
## Summary

This PR is follow-up of #900, adding fused linear cross entropy support and missing tests.

[FalconH1MLP ](https://github.com/huggingface/transformers/blob/8ac2b916b042b1f78b75c9eb941c0f5d2cdd8e10/src/transformers/models/falcon_h1/modeling_falcon_h1.py#L1007) requires additional multiplication after gate/down projection (`mlp_multiplier`) which requires modifying the custom kernel. Skipping it for now.

cc @puneeshkhanna 

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/test/convergence/bf16/test_mini_models_multimodal.py
+++ b/test/convergence/bf16/test_mini_models_multimodal.py
@@ -1066,6 +1066,10 @@ def run_mini_model_multimodal(
                     not MLLAMA_AVAILABLE,
                     reason="Mllama not available in this version of transformers",
                 ),
+                pytest.mark.skipif(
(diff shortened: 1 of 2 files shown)


puneeshkhanna:
thanks @Tcc0403  for the follow up PR. I will definitely try to use  fused linear cross entropy when I don't need to materialise logits. Hoping the the the CI tests will pass with the test cases that have been added. 
I will look into adding the MLP swiglu kernel too soon.

PR #910, opened 2025-10-16
[Cross-entropy-loss] return mean token accuracy metric with CE loss

kashif:
## Summary

Returns the mean token accuracy metric when minimizing the cross-entropy loss without materializing the logits

https://x.com/jeremyphoward/status/1703246293802586155

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/transformers/model/output_classes.py
+++ b/src/liger_kernel/transformers/model/output_classes.py
@@ -10,15 +10,68 @@ import torch
 
 from transformers.modeling_outputs import CausalLMOutputWithPast
 from transformers.modeling_outputs import MoeCausalLMOutputWithPast
-from transformers.models.gemma3.modeling_gemma3 import Gemma3CausalLMOutputWithPast
-from transformers.models.glm4v_moe.modeling_glm4v_moe import Glm4vMoeCausalLMOutputWithPast
-from transformers.models.internvl.modeling_internvl import InternVLCausalLMOutputWithPast
-from transformers.models.llava.modeling_llava import LlavaCausalLMOutputWithPast
-from transformers.models.paligemma.modeling_paligemma import PaliGemmaCausalLMOutputWithPast
-from transformers.models.qwen2_5_vl.modeling_qwen2_5_vl import Qwen2_5_VLCausalLMOutputWithPast
-from transformers.models.qwen2_vl.modeling_qwen2_vl import Qwen2VLCausalLMOutputWithPast
-from transformers.models.qwen3_vl.modeling_qwen3_vl import Qwen3VLCausalLMOutputWithPast
-from transformers.models.qwen3_vl_moe.modeling_qwen3_vl_moe import Qwen3VLMoeCausalLMOutputWithPast
+
+# The following model-specific outputs are optional and depend on the installed
+# transformers version. Guard their imports so our module remains importable
+# even when those models are not available in the environment.
+try:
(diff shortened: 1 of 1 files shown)


src/liger_kernel/transformers/functional.py
```diff
@@ -47,10 +48,20 @@ def liger_cross_entropy(
...
         return loss
-    return loss, z_loss
+
+    result = [loss]
+    if return_z_loss:
+        result.append(z_loss)
+    if return_token_accuracy:
+        result.append(token_accuracy)
+
+    return tuple(result) if len(result) > 1 else result[0]
```
albertvillanova: I see you tried to align with the code already present before your PR, but I find a potential ambiguity here: when a 2-tuple is returned, its 2nd element might be:
- `z_loss`: `return_z_loss=True, return_token_accuracy=False`
- `token_accuracy`: `return_z_loss=False, return_token_accuracy=True`

src/liger_kernel/transformers/functional.py
```diff
@@ -82,10 +94,20 @@ def liger_fused_linear_cross_entropy(
...
```
albertvillanova: The same here.

src/liger_kernel/transformers/model/falcon_h1.py
```diff
@@ -85,24 +86,32 @@ def lce_forward(
...
             labels=labels,
             shift_labels=shift_labels,
             hidden_size=self.config.hidden_size,
             **kwargs,
         )
+        # Unpack loss and token_accuracy if returned as tuple
+        if isinstance(result, tuple):
+            loss, token_accuracy = result
+        else:
+            loss = result
```
albertvillanova: The code before your PR assumed that `return_z_loss` was not passed within the `kwargs`: they assumed only the loss scalar was returned.
- Do you know if this assumption is justified?
- If not, then the returned type could be a 2 or 3-tuple (besides the scalar loss), with z_loss and/or token_accuracy

kashif:
@vaibhavjindal would you be able to kindly review?

kashif:
@shimizust this will be a breaking change i believe BTW

vaibhavjindal:
@kashif could you please elaborate on how it will be a breaking change? Will it break the intergration with transformers or trl?

kashif:
yes if someone is using the raw functions in their lib. then now that functions returns one more thing... but on the HF side this PR takes care of this

kashif:
@vaibhavjindal  see here https://github.com/linkedin/Liger-Kernel/pull/910/files#diff-7654a885e261ec4986c229d953681635ab4c8761c89342d2dde604f169783b35L350

vaibhavjindal:
@kashif got it. So if i understand correctly, it will make sure that liger remains compatible with newer versions from HF. However, just want to confirm it will break liger support with older transformers/trl versions?

kashif:
no i believe my changes here will work with older version of HF.. i just meant non-HF frameworks

kashif:
TRL relies on HF integration for the CE loss so in TRL I will just pin to the liger version that has these changes

kashif:
@vaibhavjindal  let me fix up the new qwen3-vl model to update its API

kashif:
@vaibhavjindal all good from my side

vaibhavjindal:
> @vaibhavjindal all good from my side

Thanks a lot! I will do some final checks on correctness and benchmarks and will try to get it merged soon.

kashif:
thank you so much.. also see here: https://github.com/huggingface/trl/pull/4302#issuecomment-3487634176

kashif:
thanks @vaibhavjindal  for the typo fix and making it more robust!

Issue #925, opened 2025-11-02
[Qwen3]: TypeError: liger_fused_linear_cross_entropy() got an unexpected keyword argument 'return_dict'


yeshsurya:
### 🐛 Describe the bug

The return_dict parameter is a standard transformers parameter that controls output format (ModelOutput object vs tuple). PEFT passes this parameter to the base model, but Liger Kernel's model implementations pass all **kwargs to LigerForCausalLMLoss, which eventually reaches liger_fused_linear_cross_entropy() that doesn't accept this parameter.

Full Stack Trace
Traceback (most recent call last):
  File "/opt/conda/envs/ptca/lib/python3.10/site-packages/transformers/trainer.py", line 3884, in compute_loss
    outputs = model(**inputs)
  File "/opt/conda/envs/ptca/lib/python3.10/site-packages/deepspeed/runtime/engine.py", line 2179, in forward
    loss = self.module(*inputs, **kwargs)
  File "/opt/conda/envs/ptca/lib/python3.10/site-packages/peft/peft_model.py", line 1850, in forward
    return self.base_model(
  File "/opt/conda/envs/ptca/lib/python3.10/site-packages/peft/tuners/tuners_utils.py", line 222, in forward
    return self.model.forward(*args, **kwargs)
  File "/opt/conda/envs/ptca/lib/python3.10/site-packages/liger_kernel/transformers/model/qwen3.py", line 95, in lce_forward
    loss = LigerForCausalLMLoss(
  File "/opt/conda/envs/ptca/lib/python3.10/site-packages/liger_kernel/transformers/model/loss_utils.py", line 58, in LigerForCausalLMLoss
    loss = fixed_fused_linear_cross_entropy(
  File "/opt/conda/envs/ptca/lib/python3.10/site-packages/liger_kernel/transformers/model/loss_utils.py", line 20, in fixed_fused_linear_cross_entropy
    loss = F.liger_fused_linear_cross_entropy(
TypeError: liger_fused_linear_cross_entropy() got an unexpected keyword argument 'return_dict'

### Reproduce

This issue occurs when:

Using PEFT (LoRA/QLoRA/etc.) with any Liger Kernel supported model
Training with transformers Trainer
The trainer's compute_loss calls model(**inputs) where inputs contains return_dict

### Versions

Operating System: Linux-5.15.0-1074-azure-x86_64-with-glibc2.35
Python version: 3.10.18
Liger Kernel version: 0.6.3
[...]

PR #926, opened 2025-11-02
[Qwen3]: If qwen3 is used along with peft config, peft adds opcl obj no…

yeshsurya:
…t injested further

## Summary
Fixes : #925
Fix TypeError: liger_fused_linear_cross_entropy() got an unexpected keyword
   argument 'return_dict' that occurs when using Liger Kernel with PEFT and
   transformers Trainer.

   The return_dict parameter is a standard transformers parameter that controls
   output format (ModelOutput vs tuple). When using PEFT with Liger Kernel models,
   this parameter is passed through **kwargs all the way to
   liger_fused_linear_cross_entropy() which doesn't accept it, causing training to
   crash.

   This PR adds kwargs.pop("return_dict", None) in all affected model files to
   remove the parameter before it reaches the loss calculation functions.

## Testing Done
  - Verified the fix resolves the TypeError with Qwen3 + PEFT + transformers Trainer
     - Tested training runs successfully complete without the error

- Hardware Type: H100 * 8 GPUs
- []  run make test to ensure correctness
- []  run make checkstyle to ensure code style
- []  run make test-convergence to ensure convergence

### DIFF
--- a/src/liger_kernel/transformers/model/qwen3.py
+++ b/src/liger_kernel/transformers/model/qwen3.py
@@ -81,6 +81,8 @@ def lce_forward(
     kept_hidden_states = hidden_states[:, slice_indices, :]
 
     shift_labels = kwargs.pop("shift_labels", None)
+    # Remove output-control parameters that shouldn't be passed to loss functions
+    kwargs.pop("return_dict", None)
     logits = None
     loss = None
 


shimizust (approved):
thanks for the fix

PR #927, opened 2025-11-03
Fix #920: Check cross entropy monkey patch

### LINKED ISSUE #920: Check if cross entropy is applied correctly in all monkey patch functions
### 🐛 Describe the bug

Some of them aren't patched to the correct object.
For example: InternVL doesn't have nn.CrossEntropy in its modeling file. The monkey patch should be applied to the loss_util file (See [comment](https://github.com/linkedin/Liger-Kernel/pull/919/files#r2463622535))
https://github.com/linkedin/Liger-Kernel/blob/da6e7059fd3b496f3bae6152e5814e72a495601a/src/liger_kernel/transformers/monkey_patch.py#L2070-L2072

Note that some monkey patch functions are outdated, so you may need to check them manually 

p.s. we only need to support transformers>=4.49.0. bwd-test ci is targeting to 4.49.0.

### Reproduce

_No response_

### Versions

liger_kernel==https://github.com/linkedin/Liger-Kernel/commit/da6e7059fd3b496f3bae6152e5814e72a495601a

0xtoward:
## Summary

Fix #920
Note: The test uses subprocess isolation because cross_entropy patches modify a global function (transformers.loss.loss_utils.nn.functional.cross_entropy). I'm wondering if there is an elegant way.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/transformers/monkey_patch.py
+++ b/src/liger_kernel/transformers/monkey_patch.py
@@ -393,8 +393,9 @@ def apply_liger_kernel_to_llava(
     from transformers.models.llava import modeling_llava
 
     if cross_entropy:
-        logger.warning(TRANSFORMER_DEPRECATION_WARNING)
-        modeling_llava.nn.CrossEntropyLoss = LigerCrossEntropyLoss
+        from transformers.loss.loss_utils import nn
+
+        nn.functional.cross_entropy = liger_cross_entropy
     if fused_linear_cross_entropy:
         if transformer_version >= version.parse("4.52.0"):
             if model is not None:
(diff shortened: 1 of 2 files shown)


Tcc0403:
I'm thinking about just patching the forward method as flce does with some modifcitions in [liger's loss_utils](https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/transformers/model/loss_utils.py), so we only need to toggle an option to switch between flce and ce with functools.partial. What do you think?

feiiiiii5:
@0xtoward This PR has been stale ~8 months since @Tcc0403's 2025-11-04 suggestion to switch to a `loss_utils`-based forward-patch approach (toggle CE/FLCE via `functools.partial`). Are you still planning to update it?

If not, I'd like to pick it up with @Tcc0403's blessing on direction — I'll comment on #920 to confirm the design direction is still current. Either way, please let me know so we don't duplicate work.

PR #934, opened 2025-11-06
Handle aux_loss for different transformer versions

vaibhavjindal:
## Summary

as only newer upcoming version of transformers exposes it.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/transformers/model/glm4v_moe.py
+++ b/src/liger_kernel/transformers/model/glm4v_moe.py
@@ -155,8 +155,8 @@ def lce_forward(
         output = output + (token_accuracy,) if token_accuracy is not None else output
         return output
 
-    # Return GLM4V MoE output with accuracy (using dict syntax to add extra field)
-    return LigerGlm4vMoeCausalLMOutputWithPast(
+    # Build output kwargs and include aux_loss only if present (depends on transformers version)
+    output_kwargs = dict(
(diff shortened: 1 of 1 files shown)


kashif:
@vaibhavjindal I thought it is returned: https://github.com/huggingface/transformers/blob/main/src/transformers/models/glm4v_moe/modeling_glm4v_moe.py#L1713

Tcc0403:
aux loss is introduced in https://github.com/huggingface/transformers/commit/59efd86da2fd2ce9d2af21b68956abf397009ec2, not in a released version yet. I think it's fine to keep it since aux_loss will be in the upcoming version anyway, and liger has its output class now so it wouldn't break the compatibility.

kashif:
thanks @Tcc0403 makes sense

shimizust (dismissed):

vaibhavjindal:
@kashif @Tcc0403 the CI tests were failing (https://github.com/linkedin/Liger-Kernel/actions/runs/19120007336/job/54638206278) because of aux_loss not being present in the current version of transformers. Also, it might break things if someone is using liger with an older version of transformers.

Pushed a small change to only pass aux_loss when Glm4vMoeModelOutputWithPast returns it.

cc @shimizust

shimizust (approved):

PR #953, opened 2025-11-26
[GRPO] chunk over vocab without materializing logits

kashif:
## Summary
Updating the forward pass to compute only the required per-token log probabilities, simplifying the loss function interface, and adding comprehensive tests to ensure correctness against the Triton implementation:

* The `chunk_forward` method in `fused_linear_ppo.py` now computes log probabilities only for selected tokens (using `selected_token_ids`), avoiding allocation of large `[B, T, V]` tensors and instead returning `[B, T]` tensors for per-token log probabilities. This greatly reduces memory usage, especially for large vocabularies.
* The loss computation in `_compute_chunk_loss` is updated to use these per-token log probabilities directly, and the interface for the loss function is changed accordingly (from `log_probs` to `per_token_logps`). [[1]](diffhunk://#diff-735e09196150d1282b5970a5fa9b0f069e11aaa7c0f78d1f6be63fd5a79dbd16L288-L307) [[2]](diffhunk://#diff-1534093f54f1b158be2da2b159e45561361e2479a7112e082232d3f21adc6a45L21-L28)

**Simplification and correctness improvements:**

* The `ppo_loss_fn` in `grpo_loss.py` is simplified: it now expects pre-gathered per-token log probabilities, removing the need for an internal `.gather()` operation and unnecessary handling of full log probability tensors.
* Redundant arguments and code paths for handling full vocabulary log probabilities are removed, further streamlining the code.

**Testing and validation:**

* A comprehensive test, `test_chunked_vs_triton_grpo_loss`, is added to ensure that the chunked, memory-optimized loss matches the Triton kernel implementation across a range of configurations, including different batch sizes, sequence lengths, hidden sizes, vocab sizes, loss types, and hyperparameters. This test checks per-token losses, KL divergences, clipping indicators, and reduced losses for correctness.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/chunked_loss/grpo_loss.py
+++ b/src/liger_kernel/chunked_loss/grpo_loss.py
@@ -107,9 +107,7 @@ class LigerFusedLinearGRPOFunction(LigerFusedLinearPPOBase):
             # If num_items_in_batch is not provided, fall back to distributed-normalized mask sum.
             if num_items_in_batch is not None:
                 if isinstance(num_items_in_batch, torch.Tensor):
-                    loss_normalizer = num_items_in_batch.to(
-                        device=per_token_loss.device, dtype=per_token_loss.dtype
-                    )
+                    loss_normalizer = num_items_in_batch.to(device=per_token_loss.device, dtype=per_token_loss.dtype)
                 else:
                     loss_normalizer = torch.as_tensor(
                         num_items_in_batch, device=per_token_loss.device, dtype=per_token_loss.dtype


kashif:
cc @vaibhavjindal for your consideration

vaibhavjindal:
Thanks for this change @kashif! I will take a look at this and try to get it merged soon.

Can you also run a benchmark for this to show the memory savings and speedups?

momochen:
Thanks @kashif for this change, looking forward to the benchmark on memory, the saving is going to be huge.

kashif:
this seems to be taking more memory, closing

Issue #954, opened 2025-11-26
[RFC] Native Ascend NPU Support for Liger Kernel


Ginray:
## 1. Background & Motivation
Ascend NPU is a default PyTorch device backend, natively compatible with ecosystems like Transformers, FlagGems, and Llama Factory. We’re also enabling Triton support (repo: [triton-ascend](https://gitcode.com/Ascend/triton-ascend)).  

Liger Kernel’s wide adoption has led to growing user requests for NPU support. This proposal aims to advance native adaptation, with community input welcome.

## 2. Proposed Implementation Steps
Adaptation proceeds in phases, with no breaking changes to existing devices.

### 2.1 Device Support Integration
Extend device detection logic to include NPU:
```python
def infer_device():
    """Get current device name based on available devices"""
    if torch.cuda.is_available():  # Works for Nvidia/AMD
        return "cuda"
    elif is_npu_available():  # Ascend NPU check
        return "npu"
    elif torch.xpu.is_available():
        return "xpu"
    else:
        return "cpu"
```

### 2.2 Operator Compatibility Guarantee
- **Native-run Operators**: Some (e.g., `geglu`, `fused_neighborhood_attention`) work on NPU via basic device adaptation.  
- **NPU-specific Adjustments**: Partial operators need `BLOCK_SIZE` tweaks to avoid Unified Buffer (UB) overflow.  

**Example: geglu Operator Modification**  
Original code:
```python
@triton.jit
def _geglu_tanh_forward_kernel(a, b, c, stride, n_cols: tl.constexpr, BLOCK_SIZE: tl.constexpr):
    program_id = tl.program_id(0).to(tl.int64)
    a += program_id * stride
    b += program_id * stride
    c += program_id * stride

    col_offsets = tl.arange(0, BLOCK_SIZE)
    mask = col_offsets < n_cols
    a_row = tl.load(a + col_offsets, mask=mask, other=0).to(tl.float32)
    b_row = tl.load(b + col_offsets, mask=mask, other=0)

    # GELU tanh approximation
    sqrt_2_over_pi = 0.7978845608028654
    a_cubed = a_row ** 3
    tanh_arg = sqrt_2_over_pi * (a_row + 0.044715 * a_cubed)
    c_row = 0.5 * a_row * (1 + tanh(tanh_arg)) * b_row
    tl.store(c + col_offsets, c_row, mask=mask)
```

[...]

PR #955, opened 2025-11-27
Add Ascend NPU device support.

Ginray:
## Summary
This PR is the first step in the adaptation of Ascend NPU to Liger Kernel: adding NPU device support. For details, refer to [[RFC] Native Ascend NPU Support for Liger Kernel](https://github.com/linkedin/Liger-Kernel/issues/954), Section 2.1: **Device Support Integration**.

## Details
Key Modifications:
1. Add the installation method and basic function adaptation for NPU.
2. Directly import via `triton.language.math` on NPU to avoid errors caused by non-existent interfaces.

## Testing Done
Verification Status:
We have conducted verification on **Atlas 800T A3**, and basic test cases such as `test_softmax` and `test_swiglu` have passed. We will continue to improve it in the future.
image

- Hardware Type: <BLANK>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,8 @@ def get_default_dependencies():
         return [
             "torch>=2.6.0",
         ]
+    elif platform == "npu":
+        return ["torch_npu==2.6.0", "triton-ascend"]
 
 
 def get_optional_dependencies():
@@ -67,7 +69,21 @@ def is_xpu_available():
     return False
 
 
-def get_platform() -> Literal["cuda", "rocm", "cpu", "xpu"]:
+def is_ascend_available() -> bool:
+    """Best-effort Ascend detection.
+
+    Checks for common Ascend environment variables and a possible `npu-smi`
+    utility if present.
+    """
+    try:
+        subprocess.run(["npu-smi", "info"], check=True)
+        return True
+    except (subprocess.SubprocessError, FileNotFoundError):
+        pass
+    return False
+
+
+def get_platform() -> Literal["cuda", "rocm", "cpu", "xpu", "npu"]:
     """
     Detect whether the system has NVIDIA or AMD GPU without torch dependency.
     """
@@ -86,6 +102,9 @@ def get_platform() -> Literal["cuda", "rocm", "cpu", "xpu"]:
             if is_xpu_available():
                 print("Intel GPU detected")
                 return "xpu"
+            elif is_ascend_available():
+                print("Ascend NPU detected")
+                return "npu"
             else:
                 print("No GPU detected")
                 return "cpu"
--- a/src/liger_kernel/ops/utils.py
+++ b/src/liger_kernel/ops/utils.py
@@ -78,6 +78,8 @@ def get_amp_custom_fwd_bwd() -> Callable:
             functools.partial(torch.amp.custom_fwd, device_type=device),
             functools.partial(torch.amp.custom_bwd, device_type=device),
         )
+    if hasattr(torch, "npu") and getattr(torch.npu, "amp", None) is not None:
+        return torch.npu.amp.custom_fwd, torch.npu.amp.custom_bwd
     return torch.cuda.amp.custom_fwd, torch.cuda.amp.custom_bwd
 
 
--- a/src/liger_kernel/utils.py
+++ b/src/liger_kernel/utils.py
@@ -18,10 +18,35 @@ def infer_device():
     """
     if torch.cuda.is_available():  # Works for both Nvidia and AMD
         return "cuda"
+    # Use Ascend NPU if available (torch.npu)
+    elif is_npu_available():
+        return "npu"
+    # XPU (Intel) if available
     elif torch.xpu.is_available():
         return "xpu"
     else:
         return "cpu"
+
+
+def is_npu_available() -> bool:
+    """Detect Ascend NPU availability."""
+    try:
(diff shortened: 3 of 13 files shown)


src/liger_kernel/ops/utils.py
```diff
@@ -78,6 +78,11 @@ def get_amp_custom_fwd_bwd() -> Callable:
             functools.partial(torch.amp.custom_fwd, device_type=device),
             functools.partial(torch.amp.custom_bwd, device_type=device),
         )
+    try:
+        if hasattr(torch, "npu") and getattr(torch.npu, "amp", None) is not None:
+            return torch.npu.amp.custom_fwd, torch.npu.amp.custom_bwd
+    except Exception:
+        pass
```
Tcc0403: What exception could it possibly be?

src/liger_kernel/utils.py
```diff
@@ -18,12 +18,35 @@ def infer_device():
...
+
+
+def get_npu_multi_processor_count() -> int:
+    """Return a heuristic multi-processor count for NPU."""
+    NPU_MULTI_PROCESSOR_COUNT = 48
+    if is_npu_available():
+        return NPU_MULTI_PROCESSOR_COUNT
+    # Reasonable default to avoid division by zero
+    return 1
+
```
Tcc0403, line 50: Is `torch.npu` going to support `get_device_properties` so we can get these numbers programmatically? If that's the case, I suggest using that method instead and leave magic number as a fallback. WDYT

setup.py
```diff
@@ -24,6 +24,11 @@ def get_default_dependencies():
         return [
             "torch>=2.6.0",
         ]
+    elif platform == "npu":
+        return [
+            "torch_npu==2.6.0",
```
zheliuyu: Is `torch_npu` necessary? It appears unused in the current implementation.

src/liger_kernel/ops/utils.py
ji-huazhong: Hi @Tcc0403. Allow me to briefly introduce torch-npu. For PyTorch, torch-npu is an out-of-tree plugin integrated via the [privateuse1](https://docs.pytorch.org/tutorials/advanced/privateuseone.html) mechanism, so the torch.npu module is not included in the official PyTorch installation package, and directly using `torch.npu` may result in an ArributedError (Unless you’ve also installed torch-npu :))

Returning to this section of the code, I believe the conditional statement on line 82 already fulfills the required security checks.

ji-huazhong:
```suggestion
    if hasattr(torch, "npu") and getattr(torch.npu, "amp", None) is not None:
        return torch.npu.amp.custom_fwd, torch.npu.amp.custom_bwd
```

src/liger_kernel/utils.py
ji-huazhong: As you mentioned, we can retrieve these values via `torch.npu.get_device_properties()`.

src/liger_kernel/ops/utils.py
Ginray: Thank you for your suggestion. The try-catch here is indeed redundant, and I have fixed it in the new commit.

src/liger_kernel/utils.py
Ginray: Thanks to all of you for your suggestions. This interface of torch.npu will be supported in the latest v7.2.0 release, and I have retained the original magic number for compatibility purposes.

src/liger_kernel/ops/utils.py
```diff
@@ -78,6 +78,8 @@ def get_amp_custom_fwd_bwd() -> Callable:
             functools.partial(torch.amp.custom_fwd, device_type=device),
             functools.partial(torch.amp.custom_bwd, device_type=device),
         )
+    if hasattr(torch, "npu") and getattr(torch.npu, "amp", None) is not None:
+        return torch.npu.amp.custom_fwd, torch.npu.amp.custom_bwd
     return torch.cuda.amp.custom_fwd, torch.cuda.amp.custom_bwd
```
Tcc0403, line 83: In case you didn't notice, it's written in the fallback part for torch<=2.4.0. Can we access npu version of amp fwd/bwd decorator via `torch.amp.custom_[fwd | bwd](device_type="npu")` in `torch > 2.4.0`?

Ginray: Thank you for your reminder. I have conducted verification and confirmed that this interface works properly with torch > 2.4.0:

Core Code:
```python
class CustomNPUDemoOp(torch.autograd.Function):
    @staticmethod
    @torch.amp.custom_fwd(device_type="npu")
    def forward(ctx, x):
        ctx.save_for_backward(x)
        return 2 * x

    @staticmethod
    @torch.amp.custom_bwd(device_type="npu")
    def backward(ctx, grad_output):
        x, = ctx.saved_tensors
        grad_x = grad_output * 2
        return grad_x
```

Verification Result:
image


The output is normal (no errors or unexpected behavior).
I have also verified this compatibility in our official API documentation. Thanks for your suggestion!

zheliuyu (approved):

setup.py
Ginray: Thank you for the suggestion. We still need to integrate NPU into the PyTorch framework as a plug-in via the PrivateUse1 functionality through the torch_npu library.

PR #962, opened 2025-12-03
Add `AutoLigerKernelForCausalLM.from_config` 

### LINKED ISSUE #943: `AutoLigerKernelForCausalLM.from_config` support

When creating a new model from scratch ideally we would want to use `AutoLigerKernelForCausalLM.from_config` and not `AutoLigerKernelForCausalLM.from_pretrained` but it looks like `from_config` doesn't take care of liger-kernel custom kwargs, e.g. I'd expect this to work:

```
            swiglu=False
            if self.using_random_model:
                # skip the weight loading for a faster startup if we are in a random model configuration mode
                return AutoLigerKernelForCausalLM.from_config(
                    model_config,
                    dtype=self.config.dtype.value,
                    swiglu=swiglu,
                )
            else:
                return AutoLigerKernelForCausalLM.from_pretrained(
                    name_or_path,
                    config=model_config,
                    dtype=self.config.dtype.value,
                    swiglu=swiglu,
                )
```
but it appears it just sub-classes the original class and we end up with:
```
[rank2]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/transformers/models/auto/auto_factory.py", line 456, in from_config
[rank2]:     return model_class._from_config(config, **kwargs)
[rank2]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank2]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/transformers/modeling_utils.py", line 277, in _wrapper
[rank2]:     return func(*args, **kwargs)
[rank2]:            ^^^^^^^^^^^^^^^^^^^^^
[rank2]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/transformers/modeling_utils.py", line 2311, in _from_config
[rank2]:     model = cls(config, **kwargs)
[rank2]:             ^^^^^^^^^^^^^^^^^^^^^
[rank2]: TypeError: Qwen3MoeForCausalLM.__init__() got an unexpected keyword argument 'swiglu'
```

Actually, I'm not even sure if it registers liger-kernel functionality at all when creating the model object via this route. Does it? If it's not it should assert.

Tcc0403:
## Summary

Fix #943 

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ def get_default_dependencies():
             "torch>=2.6.0",
         ]
     elif platform == "npu":
-        return ["torch_npu==2.6.0", "triton-ascend"]
+        return ["torch_npu==2.7.1", "triton-ascend"]
 
 
 def get_optional_dependencies():
--- a/src/liger_kernel/chunked_loss/cosine_similarity_loss.py
+++ b/src/liger_kernel/chunked_loss/cosine_similarity_loss.py
@@ -9,7 +9,13 @@ from liger_kernel.chunked_loss.fused_linear_distillation import LigerFusedLinear
 
 class LigerFusedLinearCosineSimilarityFunction(LigerFusedLinearDistillationBase):
     @staticmethod
-    def distillation_loss_fn(student_logits, teacher_logits, beta=1.0):
+    def distillation_loss_fn(
+        student_logits,
+        teacher_logits,
+        target=None,
+        ignore_index=None,
+        beta=1.0,
+    ):
         """
         Compute Cosine loss (Cosine Similarity Loss).
         Args:
--- a/src/liger_kernel/chunked_loss/fused_linear_distillation.py
+++ b/src/liger_kernel/chunked_loss/fused_linear_distillation.py
@@ -13,6 +13,8 @@ class LigerFusedLinearDistillationBase(torch.autograd.Function):
     def distillation_loss_fn(
         student_logits,
         teacher_logits,
+        target=None,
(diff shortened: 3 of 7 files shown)


Tcc0403:
@sfc-gh-sbekman quick question! Do you think we should patch liger kernel on model instances created by `from_pretrained` and `from_config`? or global patching as we currently do is fine?

Right now, `AutoLigerKernelForCausalLM.from_pretrained` and `AutoLigerKernelForCausalLM.from_config` apply liger kernel patch on global symbols instead of model instances, which might cause side effect on subsequent model initializations. For instance, you can't initialize the same model without liger patch after the first `AutoLigerKernelForCausalLM.from_pretrained` call.

sfc-gh-sbekman:
I don't have a sufficient bird's eye view that you have to tell, since your current approach always worked for my use cases.

I'd say if you want to make it more flexible, continue doing what you have been doing, but provide an additional flag like `liger_kernel_apply_scope=["class", "instance"]` with `"class"` being the default (current behavior).

Tcc0403:
Thank you for your feedback! I was wondering whether the approach brings inconvenience. If it's no concern for most people, I'll leave it what it is.

sfc-gh-sbekman:
As I suggested, if you have users complain then it'd help, otherwise nothing needs to be done (and can easily be done down the road)

PR #974, opened 2025-12-13
Fix: fix ignore_index not being applied in JSD distillation loss

roycho96:
## Summary

Fix `ignore_index` parameter not being applied in `LigerFusedLinearJSDLoss`.

The `ignore_index` parameter was accepted but never used in `distillation_loss_fn`, causing all tokens (including padding/prompt) to be included in loss computation.

### Changes
- Change `reduction='sum'` to `reduction='none'` for per-token masking
- Use `masked_fill` for dtype preservation (prevent bf16 → fp32 promotion)
- Add `clamp_min(1)` to prevent NaN when all tokens ignored
- Normalize by `num_valid_tokens` instead of `full_target.shape[0]`
- Add comprehensive ignore_index tests

## Testing Done

- Hardware Type: H100

### DIFF
--- a/test/chunked_loss/test_jsd_loss.py
+++ b/test/chunked_loss/test_jsd_loss.py
@@ -37,12 +37,14 @@ class HFJSDLoss(HFDistillationLoss):
             temperature=temperature,
         )
 
-    def distillation_loss(self, student_logits, teacher_logits, beta=0.5):
+    def distillation_loss(self, student_logits, teacher_logits, target=None, ignore_index=-100, beta=0.5):
         """
         Compute JSD loss (Jensen-Shannon Divergence Loss).
         Args:
-            student_logits (torch.Tensor): Logits of student tokens. Shape: (batch_size * seq_len,).
-            teacher_logits (torch.Tensor): Logits of teacher tokens. Shape: (batch_size * seq_len,).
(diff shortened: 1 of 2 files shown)


test/chunked_loss/test_jsd_loss.py
Tcc0403, line 1: nit: I prefer inserting this code block into `test_correctness` and add `ignore_index` as a pytest parameter, so we don't have to add so many test functions. 

https://github.com/linkedin/Liger-Kernel/blob/fe1ea95ff56aa3dddb848a03195a7c5bb9ea8e7a/test/transformers/test_fused_linear_cross_entropy.py#L195-L200

Don't forget to implement ignore index feature in `HFJSDLoss`.

Tcc0403:
Good catch! The implementation looks good to me. Just a nit change in testing file for maintainability.

test/chunked_loss/test_jsd_loss.py
roycho96: Done! Integrated `ignore_index` as a pytest parameter in `test_correctness` and implemented the feature in `HFJSDLoss.distillation_loss` with proper masking and normalization.

roycho96:
@Tcc0403 Thanks for the review! Updated as suggested.

Tcc0403 (approved):
Thank you!

Issue #868, opened 2025-08-29
Support PyTorch Tensor Parallelism


ron1x1-abba:
### 🚀 The feature, motivation and pitch

PyTorch provides native Tensor Parallel techniques for model training.

https://docs.pytorch.org/docs/stable/distributed.tensor.parallel.html

It would be great to have it supported in liger-kernel.

Issue #826, opened 2025-07-20
Tensor Parallel support


winglian:
### 🐛 Describe the bug

When enabling Tensor Parallelism with training, I get the following new error:

```
[rank1]:   File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/transformers/models/llama/modeling_llama.py", line 287, in forward                                                                        
[rank1]:     hidden_states = self.input_layernorm(hidden_states)                                                                                                                                                     
[rank1]:                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                     [rank1]:   File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1751, in _wrapped_call_impl                                                                             
[rank1]:     return self._call_impl(*args, **kwargs)                                                                                                                                                                 
[rank1]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                 [rank1]:   File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1857, in _call_impl                                                                                     
[rank1]:     return inner()                                                                                                                                                                                          
[rank1]:            ^^^^^^^                                                                                                                                                                                          
[rank1]:   File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1805, in inner                                                                                          
[rank1]:     result = forward_call(*args, **kwargs)                                                                                                                                                                  
[rank1]:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                  
[rank1]:   File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/liger_kernel/transformers/rms_norm.py", line 33, in forward                                                                               
[rank1]:     return LigerRMSNormFunction.apply(                                                                                                                                                                      
[rank1]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                      
[rank1]:   File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/torch/autograd/function.py", line 575, in apply                                                                                           
[rank1]:     return super().apply(*args, **kwargs)  # type: ignore[misc]                                                                                                                                             
[...]

PR #982, opened 2025-12-17
Changing RMS layer norm to accept DTensors.

kolehma8:
## Summary

Changing RMS layer norm to accept DTensors.

## Details
RMS layer norm parameters are NOT sharded under typical tensor parallelism implementations but the inputs (and gradients) may become sharded (DTensors). This PR resolves the issue by gathering the input tensors and performing the full compute in each device.

Relates Issues:
* https://github.com/linkedin/Liger-Kernel/issues/826
* https://github.com/linkedin/Liger-Kernel/issues/868

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/ops/rms_norm.py
+++ b/src/liger_kernel/ops/rms_norm.py
@@ -613,6 +613,13 @@ class LigerRMSNormFunction(torch.autograd.Function):
         X: (B, T, H) or (BxT, H)
         W: (H,)
         """
+        if isinstance(X, torch.distributed.tensor.DTensor):
+            # Input tensor is output of a tensor parallel module and
+            # needs to be gathered to a local tensor to compute
+            # RMSE layer norm on each TP worker.
+            # TODO: support CP.
+            X = X.full_tensor()
+
         Y, X, RSTD, BLOCK_SIZE, num_warps, casting_mode = rms_norm_forward(X, W, eps, offset, casting_mode, row_mode)
         ctx.offset = offset
         ctx.casting_mode = casting_mode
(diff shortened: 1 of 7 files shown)


shimizust:
Can you run the example e2e training script just using FSDP2 (since TP won't work until all kernels are tp-compatible) to ensure the perf/loss is correct?

src/liger_kernel/ops/rms_norm.py
```diff
@@ -558,6 +558,13 @@ def forward(ctx, X, W, eps, offset=0.0, casting_mode="llama", in_place=True, row
         X: (B, T, H) or (BxT, H)
         W: (H,)
         """
+        if isinstance(X, torch.distributed.tensor.DTensor):
+            # Input tensor is output of a tensor parallel module and
+            # needs to be gathered to a local tensor to compute
+            # RMSE layer norm on each TP worker.
+            # TODO: support CP.
+            X = X.full_tensor()
```
shimizust, line 621: For my understanding, pytorch native TP keeps activations as DTensors and lets subsequent ops decide what to do?

src/liger_kernel/utils.py
```diff
@@ -12,6 +12,33 @@ def is_peft_available():
     return PEFT_AVAILABLE
 
 
+def infer_backend():
```
shimizust: Suggest renaming to `infer_comm_backend`

src/liger_kernel/ops/rms_norm.py
kolehma8: That is my understanding as well.

src/liger_kernel/utils.py
kolehma8: Changed as suggested.

kolehma8:
> Can you run the example e2e training script just using FSDP2 (since TP won't work until all kernels are tp-compatible) to ensure the perf/loss is correct?

I did not see any benchmarks using FSDP in the scripts folder. I think we can say to a very high degree of confidence that this PR will not impact FSDP since if it did, the op would have crashed earlier (e.g. if any inputs are DTensors). Vice versa, this PR only impacts input DTensors and behavior for regular tensors remains unchanged.

PR #987, opened 2025-12-24
feat(NPU): add UB Manager for auto tiling strategy management

noemotiovon:
### Background and Motivation

When developing Ascend NPU operators, we frequently encounter compilation failures caused by UB (Unified Buffer) overflow. During compilation, Triton kernels check UB usage, and if it exceeds the capacity, an error is raised:
 `MLIRCompilationError: ub overflow`.

To address this issue, developers usually have to:

- Manually adjust block sizes
- Tune separately for different NPU models and input sizes
- Repeatedly trade off between performance and UB safety

This process is tedious, error-prone, and difficult to cover all scenarios.

------

### Solution

We implemented a **UB Manager** that provides:

- **Automatic UB capacity detection**: retrieved from device properties or environment variables
- **Best-practice-based tiling strategies**: dynamically compute safe block sizes based on UB capacity and operator parameters
- **Unified strategy registration system**: supports both fixed and conditional strategies, making it easy to extend
- **Integration with GEGLU and ROPE**: automatically handles UB constraints and prevents overflow

------

#### Automatic Capacity Detection

- Supports three sources: environment variables, device properties, and model default values
- Supports multiple Ascend models such as Ascend 910B1 / 910B4

#### Dynamic Strategy Computation

- **GEGLU**: computes a safe block size based on `n_cols` and `dtype_size`
- **ROPE**: computes `BLOCK_Q` and `BLOCK_K` based on `pad_n_q_head`, `pad_n_kv_head`, and `pad_hd`
- Uses an **80% safety margin** to balance performance and safety

#### Easy Extensibility

- Simple interface: adding a new kernel strategy only requires registering a function
- Supports parameterized strategies to adapt to different input sizes

------

### Implementation Details

- Added `ub_manager.py`: the core UB management class
- Updated `geglu.py` and `rope.py`: integrated UB-aware implementations

------

### Testing

Verified on Ascend NPU 910B4:

- GEGLU forward and backward pass tests
- ROPE forward and backward pass tests
- Works correctly across different input sizes and data types
- Hardware Type: <Ascend910B4>

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ascend-ub-manager-design.md
+++ b/src/liger_kernel/ops/backends/_ascend/ascend-ub-manager-design.md
@@ -136,11 +136,14 @@ The unified strategy uses the following parameters:
 - **`tiling_dims`**: Tuple specifying which dimensions can be tiled for each shape.
   - Each element can be:
     - `int`: single dimension index (e.g., `0` for first dimension)
-    - `tuple of ints`: multiple dimensions that can be tiled together
+    - `tuple of ints`: multiple dimensions that can be tiled together (non-empty)
   - For ROPE: `(0, 0)` means first dimension of each shape can be tiled
   - For GEGLU: `(0,)` means first dimension of the shape can be tiled
   - Length must match `len(shapes)`
   - Fixed dimensions (non-tiling) are automatically extracted from shapes and multiplied to get `unit_param`
+  - **Validation**: Raises `ValueError` if:
+    - Any `tiling_dim` is empty or invalid (e.g., empty tuple)
+    - Any dimension index is out of bounds (negative or >= shape length)
 
 ### 4. Strategy Computation Flow
 
@@ -169,6 +172,7 @@ Call _default_strategy() with:
 For each (shape, tiling_dim) pair:
   Normalize tiling_dim to set of dimension indices
   Validate tiling dimensions are within shape bounds
+  (Raises ValueError if invalid)
          │
          ▼
   Calculate unit_param:
@@ -276,7 +280,12 @@ def _default_strategy(
     
     For each shape, fixed dimensions (non-tiling) are multiplied together to get unit_param.
     
-    Returns tuple of max_safe_block_size (power of 2), one for each shape.
+    Returns:
+        Tuple of max_safe_block_size (power of 2), one for each shape.
+    
+    Raises:
+        ValueError: If any tiling_dim is empty or invalid, or if any dimension
+                    index is out of bounds for the corresponding shape.
     """
 ```
 
@@ -284,6 +293,8 @@ def _default_strategy(
 1. For each `(shape, tiling_dim)` pair:
    - Normalize `tiling_dim` to a set of dimension indices using `_normalize_tiling_dims`
    - Validate tiling dimensions are within shape bounds
+     - Raises `ValueError` if `tiling_dim` is empty or invalid
+     - Raises `ValueError` if any dimension index is out of bounds
    - Calculate `unit_param` as the product of all non-tiling dimensions
    - If all dimensions are tiling, `unit_param = 1.0`
 2. Calculate `SAFE_UB_CAPACITY_BITS = ub_capacity_bits * safety_margin`
@@ -309,17 +320,29 @@ def compute_default_tiling_strategy(
     Returns tuple of tiled shapes with same structure as input shapes.
     Tiling dimensions are replaced with computed block sizes (power of 2),
     while non-tiling dimensions are padded to next power of 2.
+    
+    Returns:
+        Tuple of tiled shapes, or None if shapes/tiling_dims are empty or
+        lengths don't match.
+    
+    Raises:
+        ValueError: If any tiling_dim is empty or invalid, or if any dimension
+                    index is out of bounds for the corresponding shape.
     """
 ```
 
 **Key Steps:**
 1. Get UB manager instance
 2. Validate `shapes` and `tiling_dims` (lengths must match, cannot be empty)
+   - Returns `None` if validation fails (empty or mismatched lengths)
 3. Set defaults for `dtype_size` (4) and `memory_multiplier` (10.0) if not provided
 4. Call `_default_strategy` to get `max_supported` (tuple of max_safe_block_size, one per shape)
+   - May raise `ValueError` if `tiling_dims` are invalid (see `_default_strategy` documentation)
 5. For each `(shape, tiling_dim, max_safe)`:
    - Normalize `tiling_dim` to a set of dimension indices
    - Validate tiling dimensions are within shape bounds
+     - Raises `ValueError` if `tiling_dim` is empty or invalid
+     - Raises `ValueError` if any dimension index is out of bounds
    - For each tiling dimension:
      - Compute `desired = triton.next_power_of_2(original_dim)`
      - Compute `final = min(desired, max_safe)`
--- a/src/liger_kernel/ops/backends/_ascend/ub_manager.py
+++ b/src/liger_kernel/ops/backends/_ascend/ub_manager.py
@@ -93,10 +93,15 @@ def _default_strategy(
         tiling_dim_set = _normalize_tiling_dims(tiling_dim)
 
         # Validate tiling dimensions are within shape bounds
-        if not tiling_dim_set or any(dim_idx < 0 or dim_idx >= len(shape) for dim_idx in tiling_dim_set):
-            # Invalid tiling_dim, use conservative default
-            max_safe_sizes.append(1)
-            continue
+        if not tiling_dim_set:
+            raise ValueError(
+                f"Invalid tiling_dim: {tiling_dim}. tiling_dim must be an int or a non-empty tuple of ints."
+            )
+        if any(dim_idx < 0 or dim_idx >= len(shape) for dim_idx in tiling_dim_set):
+            raise ValueError(
+                f"Invalid tiling_dim: {tiling_dim} for shape {shape}. "
+                f"All dimension indices must be in range [0, {len(shape)})."
+            )
 
         # Calculate unit_param: product of fixed (non-tiling) dimensions
         unit_param = 1.0
@@ -315,12 +320,15 @@ def compute_default_tiling_strategy(
         tiling_dim_set = _normalize_tiling_dims(tiling_dim)
 
         # Validate tiling dimensions are within shape bounds
-        if not tiling_dim_set or any(dim_idx < 0 or dim_idx >= len(result_shape) for dim_idx in tiling_dim_set):
-            # Invalid tiling_dim, keep original shape (with padding)
-            for dim_idx, dim_size in enumerate(result_shape):
-                result_shape[dim_idx] = triton.next_power_of_2(dim_size)
-            result.append(result_shape)
-            continue
+        if not tiling_dim_set:
+            raise ValueError(
+                f"Invalid tiling_dim: {tiling_dim}. tiling_dim must be an int or a non-empty tuple of ints."
+            )
+        if any(dim_idx < 0 or dim_idx >= len(result_shape) for dim_idx in tiling_dim_set):
+            raise ValueError(
+                f"Invalid tiling_dim: {tiling_dim} for shape {shape}. "
+                f"All dimension indices must be in range [0, {len(result_shape)})."
+            )
 
         # Replace tiling dimensions with computed block sizes
         # For each tiling dimension, compute: min(desired, max_safe)


noemotiovon:
# Ascend NPU UB Manager Design Document

## Overview

The UB Manager (Unified Buffer Manager) is a core component in **Liger-Kernel** responsible for managing the Unified Buffer (UB) capacity on Ascend NPUs. By automatically detecting UB capacity and providing best-practice-based tiling strategies, it helps Triton kernels avoid UB overflow errors while maintaining high performance.

## Design Goals

1. **Automated UB Management**: Automatically detect device UB capacity without manual configuration
2. **Best-Practice-Based**: Use proven tiling strategies to avoid UB overflow
3. **Flexible Strategy System**: Support both fixed strategies and conditional strategies to adapt to different scenarios
4. **Easy to Extend**: Simple interfaces for adding new kernel strategies
5. **Performance Optimization**: Maximize performance while ensuring UB safety

## Architecture Design

### Core Components

```
┌─────────────────────────────────────────────────────────┐
│                    UB Manager System                    │
├─────────────────────────────────────────────────────────┤
│                                                         │
│  ┌──────────────┐         ┌──────────────────┐          │
│  │  UBManager   │────────▶│ Strategy Registry│          │
│  │   (Singleton)│         │  (Best Practices)│          │
│  └──────────────┘         └──────────────────┘          │
│         │                            │                  │
│         │                            │                  │
│         ▼                            ▼                  │
│  ┌──────────────┐         ┌──────────────────┐          │
│  │   Capacity   │         │  Strategy        │          │
│  │  Detection   │         │  Functions       │          │
│  └──────────────┘         └──────────────────┘          │
│                                                         │
└─────────────────────────────────────────────────────────┘
         │                            │
         │                            │
         ▼                            ▼
┌──────────────┐         ┌──────────────────┐
│   GEGLU      │         │      ROPE        │
│   Kernel     │         │     Kernel       │
└──────────────┘         └──────────────────┘
```

### Class Diagram

```
┌─────────────────────────────────────┐
│          UBManager                  │
├─────────────────────────────────────┤
│ - _npu_model: str                   │
│ - _ub_capacity_bits: int            │
├─────────────────────────────────────┤
│ + ub_capacity_bits: int             │
│ + ub_capacity_bytes: int            │
│ + npu_model: str                    │
│ + get_tiling_strategy()             │
│ - _detect_npu_model()               │
│ - _detect_ub_capacity()             │
└─────────────────────────────────────┘
```

## Core Functionality

### 1. UB Capacity Detection

The UB Manager detects UB capacity in the following priority order:

1. **Environment Variable**: `ASCEND_UB_CAPACITY_BITS`
2. **Device Properties**: Retrieved from `torch.npu.get_device_properties(0).ub_capacity_bits`
3. **Model Defaults**: Use predefined values based on the detected NPU model

```
# Default UB capacity configuration
_DEFAULT_UB_CAPACITIES = {
    "Ascend910B1": 2097152,  # ~256 KB
    "Ascend910B4": 1572864,  # ~192 KB
    "default": 2097152,       # ~256 KB
}
```

### 2. Strategy Registration System

Strategies are registered via the `_TILING_STRATEGY_BEST_PRACTICES` dictionary and support two formats:

#### Fixed Strategy

```
("kernel_name", ub_capacity_bits): (block_size, ...)
```

Returns fixed tiling parameters directly, suitable for simple scenarios.

#### Conditional Strategy

```
("kernel_name", ub_capacity_bits): strategy_function
```

The strategy function dynamically computes tiling parameters based on input arguments:

```
def strategy_function(key_params: Optional[Union[Tuple, Dict]]) -> Optional[Tuple]:
    # Compute the tiling strategy based on key_params and UB capacity
    # Return (block_size, ...) or None
```

### 3. Strategy Lookup Flow

```
User calls get_tiling_strategy()
         │
         ▼
Build lookup key: (kernel_name, ub_capacity_bits)
         │
         ▼
Look up in _TILING_STRATEGY_BEST_PRACTICES
         │
         ├─── Not found ────▶ Return None
         │
         ▼
      Strategy found
         │
         ├─── Fixed tuple ────▶ Return directly
         │
         ▼
     Callable function
         │
         ▼
  Normalize key_params
         │
         ▼
  Call strategy function
         │
         ▼
   Return strategy result
```

## Usage Examples

### Basic Usage

```
from liger_kernel.ops.backends._ascend.ub_manager import get_tiling_strategy

# GEGLU forward
strategy = get_tiling_strategy("geglu_forward", (4096, 2))
if strategy:
    block_size = strategy[0]
    # Call kernel with block_size

# ROPE forward
strategy = get_tiling_strategy("rope_forward", (32, 32, 128, 4))
if strategy:
    BLOCK_Q, BLOCK_K = strategy
    # Call kernel with BLOCK_Q and BLOCK_K
```

### Usage Inside a Kernel

```
# GEGLU example
def geglu_forward(a, b):
    n_cols = a.shape[-1]
    dtype_size = 4 if a.dtype == torch.float32 else 2
    
    # Get strategy
    strategy = get_tiling_strategy("geglu_forward", (n_cols, dtype_size))
    
    if strategy is not None:
        block_size = strategy[0]
    else:
        block_size = triton.next_power_of_2(n_cols)  # Fallback
    
    # Call kernel
    kernel[(n_rows,)](a, b, c, BLOCK_SIZE=block_size)
```

## Extension Guide

### Adding a New Kernel Strategy

1. **Define the strategy function**:

```
def _my_kernel_strategy(key_params: Optional[Union[Tuple, Dict]]) -> Optional[Tuple]:
    """
    My kernel tiling strategy.
    
    Args:
        key_params: (param1, param2, dtype_size) or dict
    
    Returns:
        (block_size1, block_size2, ...) or None
    """
    if key_params is None:
        return None
    
    # Get UB capacity
    ub_manager = get_ub_manager()
    ub_capacity_bits = ub_manager.ub_capacity_bits
    
    # Extract parameters
    if isinstance(key_params, dict):
        param1 = key_params["param1"]
        param2 = key_params["param2"]
        dtype_size = key_params.get("dtype_size", 4)
    else:
        param1 = key_params[0]
        param2 = key_params[1]
        dtype_size = key_params[2] if len(key_params) > 2 else 4
    
    # Compute strategy
    SAFE_UB_CAPACITY = int(ub_capacity_bits * 0.80)
    # ... compute block_size based on memory estimation ...
    
    return (block_size1, block_size2)
```

**Register the strategy**:

```
_TILING_STRATEGY_BEST_PRACTICES = {
    # ... existing strategies ...
    
    # New strategies
    ("my_kernel_forward", 1572864): _my_kernel_strategy,
    ("my_kernel_backward", 1572864): _my_kernel_strategy,
}
```

**Use the strategy in the kernel**:

```
def my_kernel_forward(input):
    # Prepare parameters
    param1 = input.shape[0]
    param2 = input.shape[1]
    dtype_size = 4 if input.dtype == torch.float32 else 2
    
    # Get strategy
    strategy = get_tiling_strategy("my_kernel_forward", (param1, param2, dtype_size))
    
    if strategy:
        block_size1, block_size2 = strategy
    else:
        # Fallback logic
        block_size1, block_size2 = default_sizes()
    
    # Call kernel
    kernel[(grid_size,)](
        input,
        BLOCK_SIZE1=block_size1,
        BLOCK_SIZE2=block_size2,
    )
```

noemotiovon:
### Test Result:
In PR https://github.com/linkedin/Liger-Kernel/pull/986, the accuracy tolerance for the GEGLU operator was updated.
The previous tolerance was chosen based on the numerical properties of bfloat16 execution in the GEGLU operator, while the new (atol=1e-2, rtol=1e-2) implicitly assumes FP32-level precision.

On NPU, bfloat16 execution of the GEGLU operator involves low-precision accumulation in the underlying matmul, and the stricter tolerance may therefore lead to false negatives in correctness tests.

For this reason, the GEGLU test continues to use the previous tolerance setting instead of the updated one.


```
TOTAL                                                                8228   7118   2342     34    11%
Coverage HTML written to dir htmlcov
================================================================== slowest durations ===================================================================
11.90s call     test/transformers/test_geglu.py::test_correctness[dtype0-1.0-2e-06-2-2048-2048-4096]
2.11s call     test/transformers/test_geglu.py::test_correctness[dtype0-1.0-2e-06-9-41-341-4231]
2.10s call     test/transformers/test_geglu.py::test_correctness[dtype1-10000.0-0.006-9-41-341-4231]
2.05s call     test/transformers/test_geglu.py::test_correctness[dtype1-10000.0-0.006-2-2048-2048-4096]
0.15s call     test/transformers/test_rope.py::test_correctness[True-dtype0-1e-05-1e-05-1-128-32-32-64]
0.09s call     test/transformers/test_rope.py::test_correctness[True-dtype1-0.1-1e-05-1-128-32-32-64]
0.03s call     test/transformers/test_rope.py::test_correctness[True-dtype0-1e-05-1e-05-3-423-73-213-92]
0.03s call     test/transformers/test_rope.py::test_correctness[False-dtype0-1e-05-1e-05-3-423-73-213-92]
0.02s call     test/transformers/test_rope.py::test_functional_correctness[True-dtype0-1e-05-1e-05-1-2-2-2-8]
0.02s call     test/transformers/test_rope.py::test_correctness[False-dtype0-1e-05-1e-05-3-423-73-155-92]
0.02s call     test/transformers/test_rope.py::test_correctness[True-dtype0-1e-05-1e-05-3-423-73-155-92]
0.02s call     test/transformers/test_rope.py::test_correctness[False-dtype1-0.1-1e-05-3-423-73-213-92]
0.02s call     test/transformers/test_rope.py::test_correctness[True-dtype1-0.1-1e-05-3-423-73-213-92]
0.02s call     test/transformers/test_geglu.py::test_correctness_functional[dtype0-1.0-2e-06-2-2-8]
0.02s call     test/transformers/test_rope.py::test_correctness[True-dtype0-1e-05-1e-05-2-128-32-32-64]
0.02s call     test/transformers/test_rope.py::test_functional_correctness[True-dtype0-1e-05-1e-05-9-7-41-41-41]
0.02s call     test/transformers/test_rope.py::test_correctness[True-dtype1-0.1-1e-05-3-423-73-155-92]
0.02s call     test/transformers/test_rope.py::test_functional_correctness[True-dtype1-0.1-1e-05-1-2-2-2-8]
0.02s call     test/transformers/test_rope.py::test_functional_correctness[True-dtype1-0.1-1e-05-9-7-41-41-41]
0.02s call     test/transformers/test_rope.py::test_functional_correctness[False-dtype0-1e-05-1e-05-9-7-41-41-41]
0.02s call     test/transformers/test_rope.py::test_functional_correctness[False-dtype1-0.1-1e-05-9-7-41-41-41]
0.02s call     test/transformers/test_rope.py::test_functional_correctness[True-dtype1-0.1-1e-05-1-2-1-2-8]
0.02s call     test/transformers/test_rope.py::test_functional_correctness[True-dtype0-1e-05-1e-05-1-2-1-2-8]
0.02s call     test/transformers/test_rope.py::test_correctness[False-dtype1-0.1-1e-05-3-423-73-155-92]
0.01s call     test/transformers/test_geglu.py::test_correctness_functional[dtype1-10000.0-0.006-2-2-8]
0.01s call     test/transformers/test_geglu.py::test_correctness_functional[dtype0-1.0-2e-06-9-7-41]
0.01s call     test/transformers/test_geglu.py::test_correctness_functional[dtype1-10000.0-0.006-9-7-41]
0.01s call     test/transformers/test_rope.py::test_correctness[True-dtype1-0.1-1e-05-2-128-32-32-64]
0.01s call     test/transformers/test_rope.py::test_correctness[True-dtype1-0.1-1e-05-2-128-32-8-64]
0.01s call     test/transformers/test_rope.py::test_correctness[True-dtype0-1e-05-1e-05-2-128-32-8-64]
0.01s call     test/transformers/test_rope.py::test_correctness[True-dtype1-0.1-1e-05-1-128-32-8-64]
0.01s call     test/transformers/test_rope.py::test_correctness[True-dtype0-1e-05-1e-05-1-128-32-8-64]
0.01s call     test/transformers/test_rope.py::test_correctness[False-dtype1-0.1-1e-05-2-128-32-8-64]
0.01s call     test/transformers/test_rope.py::test_correctness[False-dtype1-0.1-1e-05-2-128-32-32-64]
0.01s call     test/transformers/test_rope.py::test_correctness[False-dtype0-1e-05-1e-05-2-128-32-32-64]
0.01s call     test/transformers/test_rope.py::test_correctness[False-dtype0-1e-05-1e-05-2-128-32-8-64]
0.01s teardown test/transformers/test_geglu.py::test_correctness[dtype0-1.0-2e-06-2-2048-2048-4096]
0.01s teardown test/transformers/test_rope.py::test_correctness[True-dtype0-1e-05-1e-05-3-423-73-213-92]
0.01s teardown test/transformers/test_rope.py::test_correctness[False-dtype0-1e-05-1e-05-3-423-73-213-92]
0.01s teardown test/transformers/test_rope.py::test_correctness[True-dtype0-1e-05-1e-05-3-423-73-155-92]
0.01s teardown test/transformers/test_rope.py::test_correctness[False-dtype0-1e-05-1e-05-3-423-73-155-92]
0.01s call     test/transformers/test_rope.py::test_correctness[False-dtype1-0.1-1e-05-1-128-32-32-64]
0.01s call     test/transformers/test_rope.py::test_correctness[False-dtype1-0.1-1e-05-1-128-32-8-64]
0.01s call     test/transformers/test_rope.py::test_functional_correctness[False-dtype1-0.1-1e-05-1-2-2-2-8]
0.01s call     test/transformers/test_rope.py::test_functional_correctness[False-dtype0-1e-05-1e-05-1-2-2-2-8]
0.01s call     test/transformers/test_rope.py::test_functional_correctness[False-dtype1-0.1-1e-05-1-2-1-2-8]
0.01s call     test/transformers/test_rope.py::test_correctness[False-dtype0-1e-05-1e-05-1-128-32-8-64]
0.01s call     test/transformers/test_rope.py::test_correctness[False-dtype0-1e-05-1e-05-1-128-32-32-64]
0.01s call     test/transformers/test_rope.py::test_functional_correctness[False-dtype0-1e-05-1e-05-1-2-1-2-8]
0.01s teardown test/transformers/test_geglu.py::test_correctness[dtype1-10000.0-0.006-2-2048-2048-4096]

(82 durations < 0.005s hidden.  Use -vv to show these durations.)
================================================================= 44 passed in 29.44s ==================================================================
```

noemotiovon:
Hi @Tcc0403 @zheliuyu @TianHao324, Merry Christmas Eve! 🎄 
When you have a moment, could you help take a look at this code? Thanks!

src/liger_kernel/ops/backends/_ascend/ops/rope.py
```diff
@@ -0,0 +1,228 @@
...
+    dtype_size = 4 if q.dtype == torch.float32 else 2
+    strategy = get_tiling_strategy("rope_forward", (pad_n_q_head, pad_n_kv_head, pad_hd, dtype_size))
+
+    if strategy is not None:
+        # Strategy provides BLOCK_Q and BLOCK_K
+        BLOCK_Q, BLOCK_K = strategy
+    else:
+        # Fallback to conservative defaults
+        BLOCK_Q = triton.next_power_of_2(pad_n_q_head)
+        BLOCK_K = triton.next_power_of_2(pad_n_kv_head)
```
TianHao324: Shouldn't this be the same logic as the corresponding one in `rope_back`?
```
BLOCK_Q = min(32, triton.next_power_of_2(pad_n_q_head))
BLOCK_K = min(32, triton.next_power_of_2(pad_n_kv_head))
```

src/liger_kernel/ops/backends/_ascend/ops/geglu.py
```diff
@@ -0,0 +1,210 @@
...
+    b = b.view(-1, n_cols)
+    c = torch.empty_like(a)
+    n_rows = a.shape[0]
+
+    # Calculate desired block size
+    desired_block_size, num_warps = calculate_settings(n_cols)
+
+    # Get tiling strategy from best practices
+    dtype_size = 4 if a.dtype == torch.float32 else 2
+    strategy = get_tiling_strategy("geglu_forward", (n_cols, dtype_size))
```
Tcc0403: nit: `dtype_size = a.element_size()`

src/liger_kernel/ops/backends/_ascend/ops/rope.py
```diff
@@ -0,0 +1,228 @@
...
+
+    n_row = batch_size * seq_len
+
+    # ensure dq and dk are contiguous
+    dq = dq.contiguous()
+    dk = dk.contiguous()
+
+    # Get tiling strategy from best practices
+    dtype_size = 4 if dq.dtype == torch.float32 else 2
+    strategy = get_tiling_strategy("rope_backward", (pad_n_q_head, pad_n_kv_head, pad_hd, dtype_size))
```
Tcc0403: ditto

src/liger_kernel/ops/backends/_ascend/ub_manager.py
```diff
@@ -0,0 +1,365 @@
...
+
+                            normalized_params = (
+                                triton.next_power_of_2(key_params.get("n_q_head", key_params.get("pad_n_q_head", 1))),
+                                triton.next_power_of_2(key_params.get("n_kv_head", key_params.get("pad_n_kv_head", 1))),
+                                triton.next_power_of_2(key_params.get("head_dim", key_params.get("pad_hd", 64))),
+                                key_params.get("dtype_size", 4),
+                            )
+                    else:
+                        # Generic: convert dict values to tuple (sorted for consistency)
+                        normalized_params = tuple(sorted(key_params.values()))
```
Tcc0403: Is there any specific reason why we need to convert dict to tuple for geglu and rope? Aren't they also accept key_params as dict?

src/liger_kernel/ops/backends/_ascend/ub_manager.py
```diff
@@ -0,0 +1,365 @@
...
+
+# Default UB capacities for different NPU models (in bits)
+_DEFAULT_UB_CAPACITIES = {
+    "Ascend910B1": 2097152,  # ~256 KB
+    "Ascend910B4": 1572864,  # ~192 KB
+    "default": 2097152,  # ~256 KB
+}
+
+
+def _geglu_default_strategy(key_params: Optional[Union[Tuple, Dict]]) -> Optional[Tuple]:
```
Tcc0403: should we cache the returned value to save lookup time?

src/liger_kernel/ops/backends/_ascend/ub_manager.py
```diff
@@ -0,0 +1,365 @@
...
+                if hasattr(dev_props, "ub_capacity_bits"):
+                    return dev_props.ub_capacity_bits
+            except Exception:
+                pass
+
+        # Fall back to model-based defaults
+        model = self._npu_model
+        return _DEFAULT_UB_CAPACITIES.get(model, _DEFAULT_UB_CAPACITIES["default"])
+
+    def get_tiling_strategy(
```
Tcc0403: or cache at this level?

Tcc0403 (commented):
This is cool! Overall lgtm, just some questions about caching and parameter normalization.

Regarding tolerances in geglu test, I think #986 doesn't comletely fix the numerical issue, it still needs further investigation. It's fine to relax the tolerance as you need.

src/liger_kernel/ops/backends/_ascend/ub_manager.py
noemotiovon: You're right. The previous implementation was incorrect.
We don't need to convert dict to tuple in `get_tiling_strategy`. `_geglu_default_strategy` and `_rope_default_strategy` already accept both dict and tuple, and handle both internally.
I've simplified the code by removing the conversion logic and passing `key_params` directly to the strategy functions. Parameter parsing is now handled inside each strategy function, which:
* Makes `get_tiling_strategy` only responsible for finding and calling strategies
* Lets strategy functions handle their own parameter formats
This keeps responsibilities clear and makes it easier to extend. When adding new strategy functions, we only need to handle parameter formats inside the function, without modifying `get_tiling_strategy`.

src/liger_kernel/ops/backends/_ascend/ub_manager.py
noemotiovon: Good idea! I've implemented an LRU cache to cache the strategy results. The cache has a configurable size limit (default 128 entries) and uses OrderedDict for LRU eviction. This should help avoid recomputing strategies for the same parameters.

src/liger_kernel/ops/backends/_ascend/ops/geglu.py
noemotiovon: Done.

noemotiovon:
Hi @Tcc0403,
Thank you so much for taking the time to review this, especially during the Christmas holidays — really appreciate it. Merry Christmas! 🎄

Your suggestions were very helpful. I’ve updated the code accordingly, added caching, and improved the overall functionality.

src/liger_kernel/ops/backends/_ascend/ops/rope.py
noemotiovon: Done.

Tcc0403:
> Thank you so much for taking the time to review this, especially during the Christmas holidays — really appreciate it. Merry Christmas! 🎄

My pleasure! Thank you for the contribution during holidays as well. Merry Christmas!

Just one more thing I forgot to mention, we should include your ub manager document for future contributors. You can put it under `src/liger_kernel/ops/backends/_ascend` directly without having to change anything since it's well-written already. After adding the document, I think we are good to merge this PR!

noemotiovon:
@Tcc0403, Thank you so much! I really appreciate it 😊
I actually wanted to add the document earlier as well, but wasn’t sure where the most appropriate place would be, haha. Thanks a lot for the guidance.
I’ll put it under `src/liger_kernel/ops/backends/_ascend` as suggested, and I’ll also expand the document based on the existing content to include the cache-related details, so it can be as comprehensive as possible for future contributors.
Thanks again for the helpful feedback!

Tcc0403:
Fantastic! Thank you so much!

Tcc0403 (approved):

src/liger_kernel/ops/backends/_ascend/ub_manager.py
```diff
@@ -0,0 +1,376 @@
...
+
+# Default UB capacities for different NPU models (in bits)
+_DEFAULT_UB_CAPACITIES = {
+    "Ascend910B1": 2097152,  # ~256 KB
+    "Ascend910B4": 1572864,  # ~192 KB
+    "default": 2097152,  # ~256 KB
+}
+
+
+def _geglu_default_strategy(key_params: Optional[Union[Tuple, Dict]]) -> Optional[Tuple]:
```
hipudding: I think the strategy here is tightly coupled with the kernel, so should it be implemented in the operator file, or in a more appropriate place?

noemotiovon: Yes, the tiling strategy is indeed tightly coupled with the kernel and the UB size.

If a strategy has already been validated in practice and confirmed to work on a specific device, I think it’s reasonable to put it directly in each kernel. In that case, it’s arguably no longer related to the UB manager. The current strategy in the UB manager feels more like a transitional or “over-general” solution — it uses a generic tiling strategy mainly to make sure things can run.

I’ve also been thinking about where this logic should live. It’s a strategy, so it makes sense to register it in the UB manager, but at the same time it’s very operator- and kernel-specific.

For now I put it in the UB manager for the former reason, but as things evolve, it’s totally possible that this will move into individual operators, or even become completely independent of the UB manager.

Tcc0403: What about defining strategies in corresponding operator files and adding a ub manager method that can  automatically grep and register all strategies during initialization?

noemotiovon: If the strategies are defined directly inside each operator, then they can indeed be called directly when the op is used, and it may seem unnecessary to register them in the UB manager. From this perspective, strategies are tightly coupled with operators.

However, strategies are not always in a one-to-one relationship with operators. For example, the tiling strategies for mRoPE and RoPE may be exactly the same. In such cases, embedding the strategies inside individual ops would lead to duplicated implementations and poor reusability.

Through our discussion, I’ve gradually realized that this “strategy” should currently be viewed as a general-purpose abstraction, rather than a private implementation detail of a specific operator. It is better treated as a reusable and shareable component.

Based on this, my plan is to first do a round of refactoring.

Tcc0403: Makes sense. A more generic strategy would probably be a function of:

Args: 
- user inputs
	- input shapes (can be multiple)
	- dtype
- ub manager attributes 
	- ub capacity 
	- safety margin (can be overridden if given in user inputs)

Retruns:
- tile shapes (depends on # of input shapes)
- num_stages, or more (optional, can be a follow-up PR or not supported)

Developers should be responsible for what is supposed to be loaded in triton code when calling generic strategy function. This function can abstract away architecture details and help developers calculate runnable tile sizes without ub overflow.

What do you think?

noemotiovon: I agree with your point. My current thinking is that the parameters can actually be divided into three categories:
1. data related to the dimensions that need tiling,
2. data related to computing the tiling values, and
3. UB-related information.

Ultimately, the goal is to derive a tiling result. As for how to abstract tensor shapes, I haven’t fully figured that out yet. It probably needs to evolve alongside the introduction of more operators, so we can have more references and arrive at a more general abstraction.
I’ll first refactor things based on this line of thinking, and once that’s done, we can discuss it together again. Also, thank you very much for taking the time out of your busy schedule to discuss this with us! 😊

noemotiovon:
Hi @Tcc0403 , requesting review for this UB Manager refactoring.

Changes:
* Removed strategy registry, unified to use _default_strategy (sufficient for current scenarios, extensible for future needs)
* Clarified parameter structure: `safety_margin`, `dtype_size`, `memory_multiplier`, `tiling_dims`, `unit_params`
* Removed caching (current computation is simple; can reconsider if complex strategies are added later)

All tests pass for GEGLU and ROPE. See `ascend-ub-manager-design.md` for details.
Please review, thanks!

src/liger_kernel/ops/backends/_ascend/ascend-ub-manager-design.md
```diff
@@ -0,0 +1,371 @@
...
+  - For GEGLU: typically 10.0 for backward, 7.0 for forward
+  - For ROPE: typically 3.0
+- **`tiling_dims`**: Dimensions that need tiling as tuple
+  - For GEGLU: `(n_cols,)`
+  - For ROPE: `(pad_n_q_head, pad_n_kv_head)`
+- **`unit_params`**: Parameters related to unit length of each tile as tuple
+  - All elements are multiplied together to get the final unit_param
+  - For GEGLU: `()` (empty tuple, unit_param = 1)
+  - For ROPE: `(pad_hd,)` (unit_param = pad_hd)
+  - For kernels with multiple factors: `(factor1, factor2, ...)` (unit_param = factor1 * factor2 * ...)
```
Tcc0403: In my opinion, unit_params is somewhat ambiguous. It’s difficult to infer its purpose without tracing through the source code.

From reading both the documentation and the implementation, my understanding is that unit_params represents the *fixed* dimensions, paired with `tiling_dims`, which represent the *flexible* dimensions. However, simply renaming `unit_params` to `fixed_dims` would not be very helpful. Since both parameters describe partial tile dimensions, achieving semantic clarity would also require renaming `tiling_dims` to something like `flexible_dims`. At that point, the two parameters still need to be interpreted together, which adds cognitive overhead and makes neither self-explanatory on its own.

Another concern is the implicit assumption that all tiling_dims share the same unit_params factor. This limitation could be addressed by allowing len(unit_params) to be:
- 0 : no factor (implicitly 1)

- 1 : a shared factor across all tiling_dims

- len(tiling_dims) : one factor per tiling dimension

While workable, this still feels somewhat indirect.

---
Given the above, I’d like to propose an alternative approach inspired by [Helion's tiling model](https://helionlang.com/api/generated/helion.language.tile.html). Instead of tiling_dims and unit_params, we could introduce:

- **`shapes`**: a sequence of full shapes 
- **`tiling_dims`**: a list of int/tuples as tiling dimensions (can't be zero length) 

Additional optional parameters could describe shared-memory behavior, such as whether tiles are loaded separately or simultaneously, and which memory formula to use when estimating total usage. I haven't fully throught through the naming and how to generalize the memory formula yet. We can start with defaulting to one of them and extend later.


### Some examples 

#### RoPE 

Args:

- **`shapes`**: `([n_q_head, hd], [n_kv_head, hd])`
	- we can pass original shapes and let tiling strategy function handle padding
	- compatible with padded shapes as well
- **`tiling_dims`**: `(0, 0)`
	- `n_q_head` and `n_kv_head` can be broken up into tiles
	- the rest of dimensions are padded to next power of 2 values
	- equivalent to current `num_params = (pad_hd,)` 

- keeping other parameters: `safety_margin`, `dtype_size`, `memory_multiplier`

Return:
- tuple of final tiling sizes: `([block_size_q, pad_hd], [block_size_kv, pad_hd])`
	- same structure as **`shapes`**
	


#### A simplified scenario for matmul, BLOCK_K is pre-defined 

Args:

- **`shapes`**: `([M, BLOCK_K], [BLOCK_K, N])`
- **`tiling_dims`**: `(0, 1)` 
	- `M` and `N` can be broken up into tiles

Return:
- tuple of final tiling sizes: `([BLOCK_M, BLOCK_K], [BLOCK_K, BLOCK_N])`
	- same structure as **`shapes`**

#### A more complicated scenario for matmul

Args:
- **`shapes`**: `([M, K], [K, N])`
- **`tiling_dims`**: `((0, 1), (0, 1))` 
	- all dimensions can be broken up into tiles
	- alternatively, we could allow `(None, None)` to indicate that all dimensions in each shape are sliceable, though I’m not sure if it's a good idea

Return:
- tuple of final tiling sizes: `([BLOCK_M, BLOCK_K], [BLOCK_K, BLOCK_N])`
	- same structure as **`shapes`**

This case would better combine with additional parameters (e.g., strides or hints for contiguous dimensions) to support coalesced memory accesses and so on. These extensions are not required in our initial version.


---
What do you think about this approach?

noemotiovon: This is a really good suggestion, and it actually touches on something that has bothered me about this interface design for a while.

Early on, I did consider directly operating on full shapes and letting a tiling strategy function handle padding, alignment, and tile-size decisions in a unified way. From an abstraction point of view, that approach is clearly more intuitive and easier to reason about.

At the time, however, fully supporting that model internally would have required writing more logic, especially to stay compatible with the existing interface and behavior. As a result, I effectively shifted part of the tiling logic to the caller by exposing it through the unit_params + tiling_dims combination. In hindsight, this does increase cognitive overhead for users and makes the API less self-explanatory, which is not ideal.

Your proposal inspired by Helion’s tiling model is much cleaner in terms of semantics and extensibility, and it nicely addresses the coupling issues between unit_params and tiling_dims. I plan to refactor the code following this tiling strategy, moving the tiling decision logic back into the strategy function rather than requiring callers to encode it implicitly.

Thanks a lot for the detailed and thoughtful feedback — it’s been very helpful 🙏

noemotiovon:
Hi @Tcc0403,

Based on the considerations above, I’ve gone ahead and refactored the code following this approach, centering the interface around shapes and tiling_dims while keeping the existing parameters (such as safety_margin, dtype_size, and memory_multiplier) compatible.

If you have some time, I’d really appreciate it if you could help review this refactor and share any feedback on the interface design or potential improvements. Thanks a lot.

src/liger_kernel/ops/backends/_ascend/ascend-ub-manager-design.md
```diff
@@ -0,0 +1,459 @@
...
+│   compute_default_tiling_strategy   │
+├─────────────────────────────────────┤
+│ + safety_margin: float               │
+│ + dtype_size: int                   │
+│ + memory_multiplier: float           │
+│ + shapes: List[List[int]]           │
+│ + tiling_dims: Tuple                │
+├─────────────────────────────────────┤
+│ Returns: List[List[int]]            │
+│   (same structure as shapes)        │
```
Tcc0403: nit: I think shapes and return values should be `Tupe[Tuple[int]]` since they are supposed to be immutable. Besides, the interface seems cleaner (to me) with tuple. 

For example, we have to pack shape values before function calls

```python
    batch_size, seq_len, n_q_head, head_dim = q.shape
    n_kv_head = k.shape[2]
    pad_hd = triton.next_power_of_2(head_dim)
    pad_n_q_head = triton.next_power_of_2(n_q_head)
    pad_n_kv_head = triton.next_power_of_2(n_kv_head)

    strategy = compute_default_tiling_strategy(
        safety_margin=0.90,
        dtype_size=dtype_size,
        memory_multiplier=3.0,
        shapes=[[pad_n_q_head, pad_hd], [pad_n_kv_head, pad_hd]],
        tiling_dims=(0, 0),
    )

    if strategy is not None and len(strategy) == 2:
        # Strategy returns [[block_size_q, pad_hd], [block_size_kv, pad_hd]]
        BLOCK_Q = strategy[0][0]
        BLOCK_K = strategy[1][0]
```

If we change it to tuple, it synergizes better with `.shape()` or `.size()` which return [`torch.Size`](https://docs.pytorch.org/docs/stable/size.html#torch.Size), a subclass of tuple, for function calls.

```python
    shapes = (q.shape[-2:], k.shape[-2:])  # Skip padding
    tile_shapes = compute_default_tiling_strategy(
        safety_margin=0.90,
        dtype_size=dtype_size,
        memory_multiplier=3.0,
        shapes=shapes, 
        tiling_dims=(0, 0),
    )

    if tile_shapes is not None and len(tile_shapes) == len(shapes):
        # Strategy returns ((block_size_q, pad_hd), (block_size_kv, pad_hd))
        q_tile_shape, k_tile_shape = tile_shapes
        BLOCK_Q, _ = q_tile_shape
        BLOCK_K, _ = k_tile_shape 
```

noemotiovon: Thanks for the suggestion! You're absolutely right about using tuples instead of lists.

Initially, we used `List[List[int]]` because:
- During the refactoring from the old API (which used `tiling_dims` as dimension values and `unit_params`), we focused on unifying the strategy function and making the interface more intuitive.
- We were building the result incrementally in the function (appending to a list), so using lists felt natural at the time.
- We didn't fully consider the immutability aspect and the synergy with PyTorch's torch.Size (which is a tuple subclass).

noemotiovon:
Hi @Tcc0403, We've updated the code to use `Tuple[Tuple[int, ...], ...]` throughout, including:
- Function signatures and return types
- All operator calls (ROPE and GEGLU)
- Documentation and examples

The new interface is indeed cleaner and more consistent with PyTorch conventions. Thanks for catching this!

src/liger_kernel/ops/backends/_ascend/ascend-ub-manager-design.md
```diff
@@ -0,0 +1,462 @@
...
+└─────────────────────────────────────┘
+
+┌─────────────────────────────────────┐
+│   _normalize_tiling_dims            │
+├─────────────────────────────────────┤
+│ Helper function to normalize        │
+│ tiling_dim (int or tuple) to set    │
+└─────────────────────────────────────┘
+```
```
Tcc0403, line 80: minor formatting issue: fit texts in boxes

Tcc0403:
That was fast! Overall LGTM, requesting reviews in case I overlook something.

src/liger_kernel/ops/backends/_ascend/ascend-ub-manager-design.md
noemotiovon: Done

Tcc0403 (approved):

src/liger_kernel/ops/backends/_ascend/ub_manager.py
```diff
@@ -0,0 +1,341 @@
...
+    max_safe_sizes = []
+
+    for shape, tiling_dim in zip(shapes, tiling_dims):
+        # Normalize tiling_dim to a set of dimension indices
+        tiling_dim_set = _normalize_tiling_dims(tiling_dim)
+
+        # Validate tiling dimensions are within shape bounds
+        if not tiling_dim_set or any(dim_idx < 0 or dim_idx >= len(shape) for dim_idx in tiling_dim_set):
+            # Invalid tiling_dim, use conservative default
+            max_safe_sizes.append(1)
```
hipudding: max_safe_sizes.append(_DEFAULT_UB_CAPACITIES * 0.8) ?

src/liger_kernel/ops/backends/_ascend/ops/geglu.py
```diff
@@ -0,0 +1,244 @@
...
+    if tile_shapes is not None and len(tile_shapes) > 0 and len(tile_shapes[0]) > 0:
+        # Strategy returns ((block_size,),)
+        adjusted_block_size = tile_shapes[0][0]
+    else:
+        # Fallback to desired block size if no best practice found (no tiling needed)
+        adjusted_block_size = desired_block_size
+    # Always use the unified NPU kernel
+    # When adjusted_block_size >= n_cols, the loop executes only once (no tiling)
+    # When adjusted_block_size < n_cols, the loop handles tiling automatically
+    _geglu_tanh_forward_kernel_npu[(n_rows,)](
```
hipudding, line 160: We may not be able to directly reuse the GPU approach here. Since the number of NPU cores is far smaller than that of GPUs, launching too many threads can lead to severe scheduling latency issues and potentially very poor performance.

In addition, each thread may access only a small amount of data, which can result in very low utilization of the Unified Buffer (UB).

I suggest setting the number of threads to match the number of NPU cores (configurable), and partitioning elements based on the following principle:
The minimum granularity should be computation-related data. For example, for element-wise operations, the minimum granularity is an element; for operations that require full-row computation, the minimum granularity is one row. Then determine how many such minimum-granularity elements can fit into the UB.

In short, the UB should be filled as much as possible, and the number of threads should be aligned with the number of cores. This helps reduce DMA transfer instructions and core scheduling overhead, thereby maximizing the utilization of the compute units.

hipudding: However, I think this PR does not have any compilation or correctness issues; the only concern is performance, which needs to be tuned according to the NPU architecture. Whether to perform the tuning in the current PR or after it is merged should be decided by the author, since the UBManager may be used by other NPU operators as well.

hipudding (approved):

src/liger_kernel/ops/backends/_ascend/ub_manager.py
noemotiovon: This branch handles fallback behavior for invalid parameters. When tiling_dims or shape does not meet the expected constraints, it is no longer possible to accurately derive a reasonable block_size based on UB capacity, so the logic must degrade gracefully.
There are two main fallback options in this situation:
1. Make block_size sufficiently small (e.g., set it to 1) to maximize memory safety and ensure that the kernel can still run correctly.
2. Avoid meaningful block partitioning (effectively using a very large block size). This may lead to UB overflow, but given that the parameters are already invalid, a reliable block size cannot be computed anyway.

src/liger_kernel/ops/backends/_ascend/ops/geglu.py
noemotiovon: At the moment, my priority was to resolve the issue that the operator was not available on the Ascend NPU, and to ensure that the operator can at least run correctly. As for how to write more efficient operators, I am still researching and exploring that area.

src/liger_kernel/ops/backends/_ascend/ub_manager.py
Tcc0403: It should be an invalid input in this case. We might want to raise a ValueError here to notify developers instead of silently generate bad block size.

src/liger_kernel/ops/backends/_ascend/ops/geglu.py
Tcc0403: > In short, the UB should be filled as much as possible, and the number of threads should be aligned with the number of cores. This helps reduce DMA transfer instructions and core scheduling overhead, thereby maximizing the utilization of the compute units.

Persistent kernel? We can work on that after npu ci is online.

src/liger_kernel/ops/backends/_ascend/ub_manager.py
noemotiovon: Agreed. For invalid parameters, we should raise a ValueError to avoid silently generating incorrect results and to make debugging easier.

noemotiovon: Done.

src/liger_kernel/ops/backends/_ascend/ops/geglu.py
noemotiovon: That makes sense. After the operator accuracy tests are validated, we’d be glad to move forward with integrating NPU CI.

PR #989, opened 2025-12-24
[Feature] Add elementwise_affine argument to LigerRMSNorm

niyunsheng:
## Summary

This PR adds the `elementwise_affine` argument to `LigerRMSNorm` to align its API with `torch.nn.RMSNorm`. This allows users to instantiate the normalization layer without learnable parameters.

## Details
- Added `elementwise_affine` (defaulting to `True`) to the `__init__` method of `LigerRMSNorm`.
- When `elementwise_affine` is set to `False`, `self.weight` is registered as `None`, preventing the allocation of unnecessary parameters.
- Updated `extra_repr` to include the `elementwise_affine` status in the printed model structure.
- The parameter is added at the end of the argument list to maintain backward compatibility for subclasses (e.g., `LigerRMSNormForGemma`) that rely on positional arguments during `super().__init__` calls.

## Testing Done

### DIFF
--- a/test/transformers/test_rms_norm.py
+++ b/test/transformers/test_rms_norm.py
@@ -77,17 +77,22 @@ class LlamaRMSNorm(nn.Module):
 
 # https://github.com/huggingface/transformers/blob/v4.44.2/src/transformers/models/gemma/modeling_gemma.py#L122
 class GemmaRMSNorm(nn.Module):
-    def __init__(self, hidden_size: int, eps: float = 1e-6):
+    def __init__(self, hidden_size: int, eps: float = 1e-6, elementwise_affine=True):
         super().__init__()
         self.eps = eps
-        self.weight = nn.Parameter(torch.ones(hidden_size))
+        self.elementwise_affine = elementwise_affine
+        if elementwise_affine:
+            self.weight = nn.Parameter(torch.ones(hidden_size))
+        else:
+            self.register_parameter("weight", None)
 
     def _norm(self, x):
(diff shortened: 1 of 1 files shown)


test/transformers/test_rms_norm.py
```diff
@@ -128,19 +149,33 @@ def test_correctness(bs, sl, hd, dtype, atol, rtol, reference, offset, casting_m
     do = torch.randn(bs, sl, hd, device=device, dtype=dtype)
 
     # reference (llama or gemma)
-    ref_rms = reference(hidden_size=hd).to(device).to(dtype)
+    if not elementwise_affine and reference == GemmaRMSNorm:
+        return pytest.skip("GemmaRMSNorm does not support elementwise_affine=False")
```
Tcc0403: Why doesn't GemmaRMSNorm support it?

niyunsheng: Because GemmaRMSNorm with elementwise_affine=False will be LigerRMSNorm

niyunsheng: > Why doesn't GemmaRMSNorm support it?

@Tcc0403  I believe adding support within this test file is a better approach and will reduce confusion. I’ll push a new commit shortly.

niyunsheng: > > Why doesn't GemmaRMSNorm support it?
> 
> @Tcc0403 I believe adding support within this test file is a better approach and will reduce confusion. I’ll push a new commit shortly.

@Tcc0403 Done.

PR #990, opened 2025-12-25
[Fix] Handle missing 'elementwise_affine' in RMSNorm extra_repr for patched models

niyunsheng:
## Summary

Fixes an AttributeError encountered in `LigerRMSNorm.extra_repr` when models are patched `in-place` (e.g., using `apply_liger_kernel_to_model`).

In my previous PR (#989 ), I added `elementwise_affine` to `extra_repr` to improve layer visibility.
However, when layers are replaced via monkey patching, the `LigerRMSNorm` constructor is typically skipped, leaving the instance without the `elementwise_affine` attribute.

## Testing Done

I verified the fix locally. The issue is reproducible via `test/transformers/test_monkey_patch.py`. After applying this fix, `test/transformers/test_monkey_patch.py` passes successfully.

### DIFF
--- a/src/liger_kernel/transformers/rms_norm.py
+++ b/src/liger_kernel/transformers/rms_norm.py
@@ -46,7 +46,7 @@ class LigerRMSNorm(nn.Module):
         )
 
     def extra_repr(self):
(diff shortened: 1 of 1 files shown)


niyunsheng:
cc @Tcc0403 Just a quick follow-up to fix the monkey-patching edge case from the previous PR. Thanks!

Tcc0403 (approved):
Thank you

PR #992, opened 2025-12-30
[NPU]: Add NPU support for the mrope operator

TianHao324:
## Summary

Mainly to complete the adaptation of the _triton_qwen2vl_mrope operator on the NPU：
1、Solving the operator ub overflow problem
2、Resolves issues with some concat operations not supported by ascend.
3、Fixed legacy issues with the rope operator

## Testing Done
image

Verified on Ascend NPU 910B4:
MROPE forward and backward pass tests
- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ops/qwen2vl_mrope.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/qwen2vl_mrope.py
@@ -6,57 +6,60 @@ from liger_kernel.ops.backends._ascend.ub_manager import compute_default_tiling_
 
 
 @triton.jit
-def _triton_rope_npu(
+def _triton_qwen2vl_mrope_npu(
     q_ptr,
     q_row_stride,
     k_ptr,
     k_row_stride,
     cos,
-    cos_row_stride,
     sin,
-    sin_row_stride,
     sl,
     bs: tl.constexpr,
-    cos_bs: tl.constexpr,
     n_qh: tl.constexpr,
     n_kh: tl.constexpr,
     hd: tl.constexpr,
+    mrope_section_t: tl.constexpr,
+    mrope_section_h: tl.constexpr,
     BLOCK_Q: tl.constexpr,
     BLOCK_K: tl.constexpr,
     BACKWARD_PASS: tl.constexpr = False,
 ):
     pid = tl.program_id(0).to(tl.int64)
-    batch_idx = pid // sl
-    cos_row_idx = pid % sl
 
-    cos = cos + tl.where(
-        cos_bs == 1,
-        cos_row_idx * cos_row_stride,
-        batch_idx * (sl * cos_row_stride) + cos_row_idx * cos_row_stride,
-    )
-    sin = sin + tl.where(
-        cos_bs == 1,
-        cos_row_idx * sin_row_stride,
-        batch_idx * (sl * sin_row_stride) + cos_row_idx * sin_row_stride,
-    )
+    t_end = mrope_section_t
+    h_end = t_end + mrope_section_h
+
+    t_cos = cos + pid * hd
+    h_cos = t_cos + bs * sl * hd
+    w_cos = h_cos + bs * sl * hd
+    t_sin = sin + pid * hd
+    h_sin = t_sin + bs * sl * hd
+    w_sin = h_sin + bs * sl * hd
 
     q_base = q_ptr + pid * q_row_stride
     k_base = k_ptr + pid * k_row_stride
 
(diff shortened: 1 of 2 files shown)


TianHao324:
Hi @Tcc0403 @zheliuyu @noemotiovon
When you have a moment, could you help take a look at this code? Thanks!

TianHao324:
@Tcc0403 It seems I don't have permission to add repository secrets,which is causing the check to fail.
```
╭─ Error ──────────────────────────────────────────────────────────────────────╮
│ Token missing. Could not authenticate client. If you have token credentials, │
│ see modal.com/docs/reference/modal.config for setup help. If you are a new   │
│ user, register an account at modal.com, then run `modal token new`.          │
╰──────────────────────────────────────────────────────────────────────────────╯
```

Tcc0403:
> @Tcc0403 It seems I don't have permission to add repository secrets,which is causing the check to fail.

Yeah, it's fine. Just ignore it. We didn't enable nvidia-ci for fork branches.

src/liger_kernel/ops/backends/_ascend/ops/qwen2vl_mrope.py
```diff
@@ -0,0 +1,285 @@
...
+
+    text_cos_vals = tl.load(t_cos + d_idx, mask=d_mask, other=0)
+    text_sin_vals = tl.load(t_sin + d_idx, mask=d_mask, other=0)
+    height_cos_vals = tl.load(h_cos + d_idx, mask=d_mask, other=0)
+    height_sin_vals = tl.load(h_sin + d_idx, mask=d_mask, other=0)
+    width_cos_vals = tl.load(w_cos + d_idx, mask=d_mask, other=0)
+    width_sin_vals = tl.load(w_sin + d_idx, mask=d_mask, other=0)
+
+    cos_vals = tl.where(pos_mask_t, text_cos_vals, tl.where(pos_mask_h, height_cos_vals, width_cos_vals))
+    sin_vals = tl.where(pos_mask_t, text_sin_vals, tl.where(pos_mask_h, height_sin_vals, width_sin_vals))
```
Tcc0403, line 56: nit: For readibility, I prefer the original method. But if benchmarking shows better performance with yours, we're welcome to upstream your method to original mrope implementation.

https://github.com/linkedin/Liger-Kernel/blob/6a383424208b1d79bca2462f7d93bcfb9d13da05/src/liger_kernel/ops/qwen2vl_mrope.py#L56-L63

src/liger_kernel/ops/backends/_ascend/ops/rope.py
```diff
@@ -239,8 +239,8 @@ def rope_backward(dq, dk, cos, sin):
         BLOCK_K, _ = k_tile_shape
     else:
         # Fallback to conservative defaults
-        BLOCK_Q = min(32, triton.next_power_of_2(pad_n_q_head))
-        BLOCK_K = min(32, triton.next_power_of_2(pad_n_kv_head))
+        BLOCK_Q = triton.next_power_of_2(pad_n_q_head)
+        BLOCK_K = triton.next_power_of_2(pad_n_kv_head)
```
Tcc0403, line 243: Isn't it meant to be a safe block size that wouldn't trigger ub overflow for all npu models?

zheliuyu: Same question here.

src/liger_kernel/ops/backends/_ascend/ops/qwen2vl_mrope.py
TianHao324: Because of differences in NPU memory mechanisms, maintaining the original syntax can lead to issues like `concat fail`. As an alternative, we used a safer method because this is a global variable that is only loaded once, minimizing performance impact. We are currently investigating the specific memory differences causing this problem.

src/liger_kernel/ops/backends/_ascend/ops/rope.py
TianHao324: After discussing with @noemotiovon , we felt that while changing to a minimum block size would prevent ub overflow. However, the purpose of our ub manager is to maximize the use of ub. If the code reaches this point, it means our current strategy doesn't support the current operator, so we should let the operator run as before. Success is great, but failure can also remind us to optimize. Of course, the probability of reaching this point is relatively low, and our strategy can still adapt to most operators.

Tcc0403: Makes sense.

Tcc0403 (approved):

PR #996, opened 2026-01-06
[NPU]: refine geglu memory_multiplier based on UB analysis

noemotiovon:
Update geglu forward and backward passes to use dtype-specific memory_multiplier values based on detailed buffer analysis:

- Forward: 4.0 (float16) or 3.0 (float32), was fixed 7.0
- Backward: 10.0 (float16) or 6.0 (float32), was fixed 10.0

Analysis distinguishes UB buffers from register-optimized temporaries and identifies peak memory usage at line 103 for backward pass.

This improves UB memory estimation accuracy and prevents overflow.

- Hardware Type: <Ascend910B4>

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ops/geglu.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/geglu.py
@@ -130,20 +130,26 @@ def geglu_forward(a, b):
     dtype_size = a.element_size()
     # GEGLU forward tiling strategy:
     # - Calculates maximum safe block size based on UB capacity
-    # - Memory analysis:
-    #   * Inputs: a, b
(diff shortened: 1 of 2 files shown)


noemotiovon:
Using the previous precision test.
```
@pytest.mark.parametrize(
    "dtype, atol, rtol",
    [
        # atol is for small values: they have more difference, so set atol higher
        # rtol is for larger values: they are very close, so set rtol lower
        (torch.float32, 1e-0, 2e-6),
        pytest.param(
            torch.bfloat16,
            1e4,
            6e-3,
            marks=pytest.mark.skipif(not supports_bfloat16(), reason="bfloat16 not supported on this GPU"),
        ),
    ],
)
```
image

PR #1005, opened 2026-01-12
feat: Align TiledMLP with DeepSpeed/ALST/Axolotl for PEFT compatibility

akshatvishu:
## Description
This PR refactors `LigerTiledGEGLUMLP` & `LigerTiledSwiGLUMLP` modules to be PEFT-aware. 
We shift from hardcoded parameter lists to dynamic parameter discovery (`self.parameters()`).

### Context
This change aligns Liger-Kernel with the architectural decisions made in Snowflake's Arctic Long Sequence Training (ALST).

The decision to move away from hardcoded lists was confirmed in discussions with @stas00 (https://github.com/snowflakedb/ArcticTraining/issues/334#issuecomment-3697214460 ). As noted by the author, the tiled compute wrapper is designed to be "blind" to internal parameters; it is the responsibility of the caller to supply the complete set of parameters for which gradients are required.

Furthermore, this refactor matches the logic used in DeepSpeed and Axolotl, both of which utilize self.parameters() to ensure robustness and native support for LoRA/PEFT configurations.
ref
- <https://github.com/deepspeedai/DeepSpeed/blob/8a9369d03e800e413a31503ceb0e5d39e390d845/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L703>
- <https://github.com/axolotl-ai-cloud/axolotl/blob/4ae6f766ad723104265ad7967d68b5ea76695a2e/src/axolotl/monkeypatch/tiled_mlp/patch.py#L56>

### Numerical Stability
In BF16, we observed accumulation drift for narrow tensors (hidden_size < 128). 
Following standard numerical stability practices, we skip these specific unstable 
configurations in tests to maintain high parity standards for production-scale shapes.

### DIFF
--- a/src/liger_kernel/transformers/tiled_mlp.py
+++ b/src/liger_kernel/transformers/tiled_mlp.py
@@ -57,11 +57,7 @@ class LigerTiledGEGLUMLP(nn.Module):
         Returns:
             Output tensor of the same shape as input
         """
-        compute_params = [
-            self.gate_proj.weight,
-            self.up_proj.weight,
-            self.down_proj.weight,
-        ]
+        compute_params = [p for p in self.parameters() if p.requires_grad]
 
         return apply_tiled_mlp(
             fn=self._mlp_forward,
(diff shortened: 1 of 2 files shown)


akshatvishu:
@Tcc0403  Thanks for the quick review! Also, I am wondering if we need to re-run/update <https://github.com/linkedin/Liger-Kernel/blob/main/benchmark/scripts/benchmark_tiled_mlp.py> after this change?

Tcc0403:
@akshatvishu Oops, I forgot the benchmark script. Yes, we need to update deepspeed implementations.

Benchmark data will be updated in https://github.com/linkedin/Liger-Kernel/tree/gh-pages/benchmarks every week

PR #1006, opened 2026-01-13
avoid pointer mutation in layer_norm kernel

TianHao324:
## Summary
Rewrite layer_norm kernel to use explicit channel offsets instead of mutating X/Y base pointers inside loops.

This improves Triton compiler optimization opportunities, enables more predictable memory access patterns, and avoids loop-carried pointer dependencies.

## Testing Done

image

- Hardware Type: Ascend NPU 910B4

### DIFF
--- a/src/liger_kernel/ops/layer_norm.py
+++ b/src/liger_kernel/ops/layer_norm.py
@@ -8,6 +8,7 @@ import triton.language as tl
 from liger_kernel.ops.utils import calculate_settings
 from liger_kernel.ops.utils import compare_version
 from liger_kernel.ops.utils import ensure_contiguous
+from liger_kernel.utils import get_npu_multi_processor_count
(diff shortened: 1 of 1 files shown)


src/liger_kernel/ops/layer_norm.py
```diff
@@ -299,8 +296,9 @@ def layer_norm_backward(dY, X, W, B, Mean, RSTD):
     )
 
     DX = DX.view(*shape)
-    DW = _DW.sum(dim=0).to(W.dtype)
-    DB = _DB.sum(dim=0).to(B.dtype)
+    DW = _DW.sum(dim=0)
```
noemotiovon: From the perspective of API contracts and type consistency, this conversion should be handled within the `layer_norm_backward` method.

TianHao324: edited

PR #1007, opened 2026-01-13
gemma3 consider loss_kwargs

jp1924:
## Summary
When applying the liger-kernel in SFTTrainer of the latest version of TRL (0.26.2), `return_token_accuracy` is also passed to input_data to compute `token_accuracy` alongside compute_loss.
However, in Gemma3, `return_token_accuracy` is applied correctly during the loss step in causal_forward but not in multimodal_forward.
Therefore, using inspect, I wrote code to separate only the kwagrs that can enter LCE from lm_kwagrs and pass them to loss_kwagrs.
Using this, it functions correctly even in the latest version of trl.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/transformers/model/loss_utils.py
+++ b/src/liger_kernel/transformers/model/loss_utils.py
@@ -73,7 +73,7 @@ def LigerForCausalLMLoss(
     return_token_accuracy: bool = False,
     **kwargs,
 ):
-	# Filter out inapplicable kwargs to liger_fused_linear_cross_entropy
+    # Filter out inapplicable kwargs to liger_fused_linear_cross_entropy
     applicable_params = inspect.signature(F.liger_fused_linear_cross_entropy).parameters
     kwargs = {k: v for k, v in kwargs.items() if k in applicable_params}
 


src/liger_kernel/transformers/model/gemma3.py
```diff
@@ -268,7 +270,11 @@ def multimodal_forward(
         shift_hidden_states = shift_hidden_states.view(-1, self.config.text_config.hidden_size)
         shift_labels = shift_labels.view(-1).to(hidden_device)
 
-        lce = LigerFusedLinearCrossEntropyLoss()
+        accept_params = inspect.signature(LigerFusedLinearCrossEntropyLoss).parameters
+        remain_params = set(lm_kwargs) - (set(accept_params) & set(lm_kwargs))
+        loss_kwargs = {k: v for k, v in lm_kwargs.items() if k not in remain_params}
+
+        lce = LigerFusedLinearCrossEntropyLoss(**loss_kwargs)
```
Tcc0403: Is it possible to use `LigerForCausalLMLoss`?

jp1924: sure!

src/liger_kernel/transformers/model/gemma3.py
```diff
@@ -250,69 +254,29 @@ def multimodal_forward(
...
-            shift_labels = shift_labels[shift_attention_mask.to(shift_labels.device) != 0].contiguous()
-        else:
-            shift_hidden_states = shift_hidden_states.contiguous()
-            shift_labels = shift_labels.contiguous()
-
-        # Flatten hidden state
-        shift_hidden_states = shift_hidden_states.view(-1, self.config.text_config.hidden_size)
-        shift_labels = shift_labels.view(-1).to(hidden_device)
-
-        lce = LigerFusedLinearCrossEntropyLoss()
```
Tcc0403: I don't think the change is equivalent to this section. 

If `attention_mask` is provided, it will remove tokens from `shift_hidden_states` and `shift_labels` where `shift_attention_mask == 0`. 

You can either
1. leave attention masking process here before calling `LigerForCausalLMLoss`, or
2. introduce a `attention_mask` argument to `LigerForCausalLMLoss` and put the process in it.

src/liger_kernel/transformers/model/gemma3.py
```diff
@@ -240,6 +240,10 @@ def multimodal_forward(
     slice_indices = slice(-logits_to_keep, None) if isinstance(logits_to_keep, int) else logits_to_keep
     kept_hidden_states = hidden_states[:, slice_indices, :]
 
+    accept_params = inspect.signature(LigerForCausalLMLoss).parameters
+    remain_params = set(lm_kwargs) - (set(accept_params) & set(lm_kwargs))
+    loss_kwargs = {k: v for k, v in lm_kwargs.items() if k not in remain_params}
+
```
Tcc0403: Move `loss_kwargs` filter logic inside `LigerForCausalLMLoss`

src/liger_kernel/transformers/model/gemma3.py
```diff
@@ -250,69 +254,29 @@ def multimodal_forward(
...
-            shift_logits = logits[..., :-1, :]
-            if attention_mask is not None:
-                # we use the input attention mask to shift the logits and labels, because it is 2D.
-                # we also crop attn mask in case it is longer, which happens in PrefixTuning with peft
-                shift_attention_mask = attention_mask[:, -shift_logits.shape[1] :].to(logits.device)
-                shift_logits = shift_logits[shift_attention_mask.to(logits.device) != 0].contiguous()
-                shift_labels = shift_labels[shift_attention_mask.to(shift_labels.device) != 0].contiguous()
-            else:
-                shift_logits = shift_logits.contiguous()
-                shift_labels = shift_labels.contiguous()
```
Tcc0403, line 324: keep attention masking process here as well.

src/liger_kernel/transformers/model/gemma3.py
```diff
@@ -250,69 +254,29 @@ def multimodal_forward(
...
-            # Flatten the tokens
-            loss_fct = nn.CrossEntropyLoss()
-
-            flat_logits = shift_logits.view(-1, self.config.text_config.vocab_size)
-            flat_labels = shift_labels.view(-1).to(shift_logits.device)
-            loss = loss_fct(flat_logits, flat_labels)
-        elif shift_labels is not None:
```
Tcc0403, line 312: not in the scope of this PR, but you can merge two conditions to reduce lines of code
```python
if labels is not None or shift_labels is not None:
	logits = logits.float()
	shift_logits = logits[,,,. :-1, :]
	shift_labels = shift_labels or labels[..., 1:] 
	...
```

jp1924:
@Tcc0403
When using LigerForCausalLMLoss in the causal_forward,
I had removed the mask calculation part from multimodal_forward because it calculated the loss without the shift_mask calculation layers before and after. That was short-sighted of me.

Therefore, I reverted it to the original code and added the kwargs filtering code only to the LigerForCausalLMLoss layer, as you requested.

src/liger_kernel/transformers/model/loss_utils.py
Tcc0403: Move `lce_param_keys` to here as a lookup table

Tcc0403:
```python
def _filter_received_keys(**kwargs):
	# inspect `fixed_fused_linear_cross_entropy` function parameters
```

src/liger_kernel/transformers/model/loss_utils.py
```diff
@@ -71,6 +73,10 @@ def LigerForCausalLMLoss(
     return_token_accuracy: bool = False,
     **kwargs,
 ):
+    accept_params = inspect.signature(LigerForCausalLMLoss).parameters
+    remain_params = set(kwargs) - (set(accept_params) & set(kwargs))
+    kwargs = {k: v for k, v in kwargs.items() if k not in remain_params}
+
```
Tcc0403: https://github.com/linkedin/Liger-Kernel/blob/41d4bcfd5717e25172c4bbfec6fa5e93877496ae/src/liger_kernel/transformers/functional.py#L70-L85

I think we should only filter out kwargs that are not applicable to `liger_fused_linear_cross_entropy`.

```suggestion
	# Filter out inapplicable kwargs to liger_fused_linear_cross_entropy
    applicable_params = inspect.signature(F.liger_fused_linear_cross_entropy).parameters
    kwargs = {k: v for k, v in kwargs.items() if k in applicable_params}

```

src/liger_kernel/transformers/model/gemma3.py
```diff
@@ -283,8 +282,15 @@ def multimodal_forward(
         }
         lce_kwargs = {k: lm_kwargs.pop(k) for k in lce_param_keys if k in lm_kwargs}
```
Tcc0403: It seems #1010 adds a kwargs filter, but I think putting it inside `LigerForCausalLMLoss` makes more sense. You can remove this part.

PR #1017, opened 2026-01-15
Fix runtime issues for FSDP/DeepSpeed training

SajanGhimire1:
- Removed unsafe usage of _MISSING_TYPE in parse_args.
- Fixed KeyError in DataModule by correcting dataset field access.
- Replaced set-based FSDP auto_wrap_policy with transformer_auto_wrap_policy.
- Corrected Lightning precision strings to valid values (bf16-mixed).
- Fixed devices argument to safely detect available GPUs.
- Added safe get() for labels in training/validation steps to avoid KeyError.

## Summary
Ensures stable and correct training across multi-GPU setups with FSDP/DeepSpeed by fixing dataset handling, auto-wrap policy, precision settings, and device detection.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/examples/lightning/training.py
+++ b/examples/lightning/training.py
@@ -34,6 +34,7 @@ class Args:
     data: str = "cais/mmlu"
     output_dir: str = "mmlu_finetuning"
     max_length: int = 2048
+    # for llam3 8B model, deepspeed will OOM with 16 on 8XA100 80G and 8 will OOM with 8XA100 40G
     batch_size: int = 4
     lr: float = 6e-6
     weight_decay: float = 0.05
@@ -70,6 +71,7 @@ class LanguageModel(pl.LightningModule):
         self.model = None
 
     def configure_model(self):
+          # https://lightning.ai/docs/pytorch/stable/advanced/model_parallel/fsdp.html#speed-up-model-initialization
         if self.model is not None:
             return
         self.model = AutoLigerKernelForCausalLM.from_pretrained(


examples/lightning/training.py
Tcc0403, line 19: What `trl` version should I use? I couldn't import `DataCollatorForCompletionOnlyLM`

SajanGhimire1: `DataCollatorForCompletionOnlyLM` class requires **trl >= 0.8.0**. Please make sure that version (or higher) is installed to avoid import issues.

Tcc0403: `trl==0.26.2` doesn't work

examples/lightning/training.py
```diff
@@ -33,7 +34,6 @@ class Args:
     data: str = "cais/mmlu"
     output_dir: str = "mmlu_finetuning"
     max_length: int = 2048
-    # for llam3 8B model, deepspeed will OOM with 16 on 8XA100 80G and 8 will OOM with 8XA100 40G
```
Tcc0403, line 36: why removing comments?

examples/lightning/training.py
SajanGhimire1: `trl==0.26.2` no longer exposes `DataCollatorForCompletionOnlyLM` in the same way. This example is intended to work with **older TRL releases where the collator exists**, specifically: `trl>=0.8.0,<0.21.0` In newer TRL versions (including 0.26.x), the collator was refactored/removed, which causes the import error.

examples/lightning/training.py
SajanGhimire1: Restore comments without functional changes

PR #1037, opened 2026-01-21
Workaround for OOM error on `benchmark_jsd`

Egor-Krivov:
## Summary
The benchmark currently requires a lot of memory with current configuration (69GBs), it's the heaviest of them all based on current results from `all_benchmark_data.csv`, so I added a workaround for GPUs with not enough memory.

## Details
Alternative implementation could be to replace all benchmark function calls with another function that would process OOM errors, but that would require changing all benchmarks. We would need to replace all `triton.testing.do_bench` and  with local function that would handle OOM errors and change `_test_memory` as well. And we would probably need to start saving some `inf` results in the csv.

## Testing Done
Tested specifically changed benchmark.

- Hardware Type: Intel GPU Max 1550

### DIFF
--- a/src/liger_kernel/utils.py
+++ b/src/liger_kernel/utils.py
@@ -131,6 +131,6 @@ def get_total_gpu_memory() -> int:
     elif device == "xpu":
         return torch.xpu.get_device_properties(0).total_memory // (1024**3)
     elif device == "npu":
-        return torch.npu.get_device_properties(0).total_memory // (1024 ** 3)
+        return torch.npu.get_device_properties(0).total_memory // (1024**3)
     else:
         raise RuntimeError(f"Unsupported device: {device}")


src/liger_kernel/utils.py
```diff
@@ -121,3 +121,14 @@ def transformers_version_dispatch(
...
+
+def get_total_gpu_memory() -> int:
+    """Returns total GPU memory in GBs."""
+    device = infer_device()
+    if device == "cuda":
+        return torch.cuda.get_device_properties(0).total_memory // (1024**3)
+    elif device == "xpu":
+        return torch.xpu.get_device_properties(0).total_memory // (1024**3)
+    else:
+        raise RuntimeError(f"Unsupported device: {device}")
```
Tcc0403, line 136: We also need to support "npu" device. @noemotiovon Do you know the equivalent api for npu `total_memory`? Is it `torch.npu.mem_get_info[0]` according to the [source code](https://github.com/Ascend/pytorch/blob/d61ec33ea54dc7a8351a8f768d2248e5a3a22f59/torch_npu/npu/__init__.py#L478C7-L478C7)?

noemotiovon: The torch.npu APIs are intentionally kept consistent with torch.cuda as much as possible.
For total memory, you can directly use:
```
torch.npu.get_device_properties(0).total_memory // (1024**3)
```

Tcc0403:
```suggestion
    if device == "cuda":
        return torch.cuda.get_device_properties(0).total_memory // (1024**3)
    elif device == "xpu":
        return torch.xpu.get_device_properties(0).total_memory // (1024**3)
    elif device == "npu":
        return torch.npu.get_device_properties(0).total_memory // (1024 ** 3)
    else:
        raise RuntimeError(f"Unsupported device: {device}")
```

src/liger_kernel/utils.py
```diff
@@ -121,3 +121,16 @@ def transformers_version_dispatch(
...
+
+def get_total_gpu_memory() -> int:
+    """Returns total GPU memory in GBs."""
+    device = infer_device()
+    if device == "cuda":
+        return torch.cuda.get_device_properties(0).total_memory // (1024**3)
+    elif device == "xpu":
+        return torch.xpu.get_device_properties(0).total_memory // (1024**3)
+    elif device == "npu":
+        return torch.npu.get_device_properties(0).total_memory // (1024 ** 3)
```
Tcc0403:
```suggestion
        return torch.npu.get_device_properties(0).total_memory // (1024**3)
```

PR #1038, opened 2026-01-22
[NPU]: use get_soc_spec for UB capacity detection

noemotiovon:
Replace hardcoded _DEFAULT_UB_CAPACITIES dictionary with dynamic detection using CANN's get_soc_spec("UB_SIZE") API. This makes UB capacity detection more accurate and device-specific.

Changes:
- Remove _DEFAULT_UB_CAPACITIES dictionary
- Use get_soc_spec("UB_SIZE") to query UB size from device (returns bytes)
- Convert bytes to bits automatically (bytes * 8)
- Improve error handling with clear RuntimeError messages
- Update documentation to reflect new detection flow

Detection priority:
1. Environment variable ASCEND_UB_CAPACITY_BITS (in bits)
2. get_soc_spec("UB_SIZE") (bytes -> bits conversion)
3. Raise RuntimeError with instructions if both fail

All UB capacity values are consistently stored and used in bits throughout the system.

Hardware Type: Ascend910B4

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ub_manager.py
+++ b/src/liger_kernel/ops/backends/_ascend/ub_manager.py
@@ -17,13 +17,6 @@ import triton
 
 from liger_kernel.utils import is_npu_available
 
-# Default UB capacities for different NPU models (in bits)
-_DEFAULT_UB_CAPACITIES = {
-    "Ascend910B1": 2097152,  # ~256 KB
-    "Ascend910B4": 1572864,  # ~192 KB
-    "default": 2097152,  # ~256 KB
-}
-
 
 def _normalize_tiling_dims(tiling_dim: Union[int, Tuple[int, ...]]) -> set:
     """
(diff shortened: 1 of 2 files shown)


noemotiovon:
I am currently running the test cases and have printed the results of the retrieved UB memory, which are as expected.

image

noemotiovon:
Hi @Tcc0403 , Could you help me review the code? I took a shortcut earlier, and now I’ve gone back and fixed it.

src/liger_kernel/ops/backends/_ascend/ub_manager.py
```diff
@@ -183,31 +176,56 @@ def _detect_npu_model(self) -> str:
...
-                dev_props = torch.npu.get_device_properties(0)
-                if hasattr(dev_props, "ub_capacity_bits"):
-                    return dev_props.ub_capacity_bits
-            except Exception:
-                pass
-
-        # Fall back to model-based defaults
-        model = self._npu_model
-        return _DEFAULT_UB_CAPACITIES.get(model, _DEFAULT_UB_CAPACITIES["default"])
+                from tbe.common.platform import get_soc_spec
```
Tcc0403, line 200: Do we need to introduce a new dependency? Or it's in one of the current packages
https://github.com/linkedin/Liger-Kernel/blob/57e98d3a2a3df2b5986c681d1b5428790e0432a9/setup.py#L28

noemotiovon: We don’t need to introduce a new dependency here because this code path is only exercised in the NPU environment, which already includes and relies on tbe-related functionality.
In the NPU environment, tbe is shipped as part of the CANN / Ascend toolchain and is exposed via environment configuration (e.g. PYTHONPATH), so it can be imported at runtime.

Tcc0403: I see, thanks for clarifying!

PR #1049, opened 2026-01-29
[NPU]: add support for grpo loss

TianHao324:
## Summary

To facilitate the integration of CI, we tested the chunked loss. Due to the differences in the NPU devices, disabling the torch compilation was necessary to pass most of the tests. However, some test cases of the group loss operator failed. The root cause was that the parent class LigerFusedLinearPPOBase did not convert the logits-related data to float32 when calculating, while the NPU has errors when computing bf16 data. Therefore, we made modifications here to first support the CI integration.

## Testing Done

image

### DIFF
--- a/test/chunked_loss/test_grpo_loss.py
+++ b/test/chunked_loss/test_grpo_loss.py
@@ -113,7 +113,7 @@ class TorchLMHeadGRPO(torch.nn.Module):
     ):
         logits = x @ self.lin.weight.t()
         if self.lin.bias is not None:
-            logits = logits + self.lin.bias.float()
+            logits = logits + self.lin.bias
         if self.temperature != 1.0:
             logits = logits / self.temperature
         # Get log probabilities
@@ -344,7 +344,7 @@ def test_correctness(
         if torch_lm_head_grpo.lin.bias is not None:
             logits = logits + torch_lm_head_grpo.lin.bias
         logits = logits / temperature
-        logps = F.log_softmax(logits.float(), dim=-1)
+        logps = F.log_softmax(logits, dim=-1)
         per_token_logps = logps.gather(dim=-1, index=selected_token_ids.unsqueeze(-1)).squeeze(-1)
 
     # Create attention mask with random padding [B, T]


src/liger_kernel/utils.py
```diff
@@ -70,7 +70,7 @@ def get_npu_multi_processor_count() -> int:
     if is_npu_available():
         NPU_MULTI_PROCESSOR_COUNT = 48
         dev_props = torch.npu.get_device_properties()
-        # The vector_core_num attribute is supported in the torch.npu v7.2.0 release version.
+        # The vector_core_num attribute is supported in the torch.npu v2.7.1 release version.
```
noemotiovon: No changes are needed here.

TianHao324:

src/liger_kernel/chunked_loss/fused_linear_ppo.py
```diff
@@ -320,6 +322,9 @@ def chunk_forward(input_chunk, weight, bias=None, temperature=1.0):
         """Forward pass computation for a single chunk without explicit reshaping."""
         # Directly compute logits via batched matrix multiplication: [B, T, H] @ [H, V] -> [B, T, V]
         logits = torch.matmul(input_chunk, weight.t())
+        # There are issues with the calculation of NPU on BF16.
+        if is_npu_available():
+            logits = logits.float()
```
Tcc0403: The root cause is the discrepancy between liger's and the one in test https://github.com/linkedin/Liger-Kernel/blob/a5be02ea2f557a1ad727f1e915e8dfa1a5525cff/test/chunked_loss/test_grpo_loss.py#L116

not specific to npu.

I think it makes more sense to remove the upcasting in test files to match the implementations.

src/liger_kernel/ops/backends/_ascend/ops/embedding.py
```diff
@@ -109,7 +109,7 @@ def embedding_backward_kernel(
 
 def get_optimal_block_size(total_elements, dtype_size, BLOCK_SIZE_N: tl.constexpr):
     # 1. Set Memory Multiplier
-    # 3.0 are empirical values based on 910B UB (192KB)
+    # 3.0 are empirical values based on Atlas 800I A2 UB (192KB)
```
Tcc0403: Do you mind opening another PR for all model changes in comments?

src/liger_kernel/chunked_loss/fused_linear_ppo.py
TianHao324: modified

src/liger_kernel/ops/backends/_ascend/ops/embedding.py
TianHao324: OK, I will make the modification in the next PR.

TianHao324:
@Tcc0403 I just updated the branch, but there was a problem with Checkstyle. It might have been caused by other PRs.

TianHao324:
@Tcc0403 May I ask if there are any other issues with this PR?

Tcc0403:
My bad, I forgot this PR.

~Isn't the parent class having logits in float?
https://github.com/linkedin/Liger-Kernel/blob/0cf4917645d0d40a3d8b197d2ed21773de0a52b1/src/liger_kernel/chunked_loss/fused_linear_ppo.py#L371~

update: Never mind, I've retrived the context now.

TianHao324:
> Isn't the parent class having logits in float?
> 
> https://github.com/linkedin/Liger-Kernel/blob/0cf4917645d0d40a3d8b197d2ed21773de0a52b1/src/liger_kernel/chunked_loss/fused_linear_ppo.py#L371

@Tcc0403  However, before this, the calculation of logits cannot be guaranteed to be of float type. Due to the type support limitations of the NPU, errors may have occurred. Just like in my first modified version, if the logits had been converted to float type at the very beginning, no errors would have arisen.
```
        logits = torch.matmul(input_chunk, weight.t())
        if bias is not None:
            logits = logits + bias  # Broadcasts bias to [B, T, V]
        if temperature != 1.0:
            logits = logits / temperature
```

test/chunked_loss/test_grpo_loss.py
Tcc0403, line 120: Should it be same casting patterns as `LigerFusedLinearPPOBase`?

no bias upcasting on L116

https://github.com/linkedin/Liger-Kernel/blob/0cf4917645d0d40a3d8b197d2ed21773de0a52b1/src/liger_kernel/chunked_loss/fused_linear_ppo.py#L361-L373

test/chunked_loss/test_grpo_loss.py
```diff
@@ -344,7 +344,7 @@ def test_correctness(
         if torch_lm_head_grpo.lin.bias is not None:
             logits = logits + torch_lm_head_grpo.lin.bias
         logits = logits / temperature
-        logps = F.log_softmax(logits.float(), dim=-1)
+        logps = F.log_softmax(logits, dim=-1)
```
Tcc0403, line 347: Could you elaborate on how this line affects the testing results?

test/chunked_loss/test_grpo_loss.py
TianHao324: You're right. The test should be consistent with the specific implementation process. I will revise it.

test/chunked_loss/test_grpo_loss.py
TianHao324: When the test data is in bfloat16 format, Torch expects to calculate log_softmax with float32 precision. However, Liger, due to the implementation of its parent class, processes the logits initially in bfloat16 precision as we mentioned above. This difference will be significantly amplified in extreme logit scenarios during LLM training and under the device limitations of the NPU, resulting in a huge disparity in the clipping detection results.

Tcc0403: Thanks for your explanation. So the problem is that `per_token_lgops`, `ref_per_token_logps` and `old_per_token_logps` have different casting path in liger's and torch ref's, leading to numerical errors caused by some precision sensitive functions, such as clipping. Do I understand it correctly?

Tcc0403: I was wondering if we should match the casting path between liger's and torch's. WDYT?

TianHao324: In my opinion, the root cause of this problem lies in the fact that torch and liger did not ensure that each step in the calculation process was of the same type, which led to the error. As long as we can ensure this, we can avoid this problem.

Tcc0403: Right, could you help me fix test ref model to match liger's?

TianHao324: OK, just no bias upcasting on L116 as you just said

TianHao324:
@Tcc0403 I have completed the revisions. I think it is now ready to be merged.

PR #1035, opened 2026-01-21
[NPU] Add Llama4_rope support on NPU

lowdy1:
## Summary
This PR implements a fully executable Llama4 RoPE operator for Ascend NPU.
1. Prevents UB overflow issues specific to NPU execution
2. Implements interleaved complex layout compatible with NPU kernels

## Testing Done
image
Test done with `python -m pytest ./test/transformers/test_llama4_rope.py -v`

Verified on Atlas 800I A2(32G)

- [x] run `make test` to ensure correctness
- [x] run `make checkstyle` to ensure code style
- [ ] run `make test-convergence` to ensure convergence

PR #1053, opened 2026-02-02
[NPU] Frequencies fusion for Llama4_rope on NPU

lowdy1:
## Summary
This PR is a descendant of https://github.com/linkedin/Liger-Kernel/pull/1035
It removes `_prepare_freqs` for simplicity and directly uses a single `freq_complex_ptr `for llama4_rope frequencies inside the Triton kernel.
By avoiding extra preprocessing and reducing load, this approach simplifies the code path and improves performance. Benchmark results show better performance compared to the previous implementation.

## Testing Done
Test done with  `python -m pytest ./test/transformers/test_llama4_rope.py -v`
Verified on Atlas 800I A2

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ops/llama4_rope.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/llama4_rope.py
@@ -40,7 +40,7 @@ def _triton_llama4_rope_npu(
     """
     Llama4 RoPE on Ascend NPU for interleaved complex layout:
     - q/k shape: (bs, sl, n_heads, hd)
-    - freqs_complex_ptr: (bs, sl, hd//2, 2)
+    - freqs_complex_ptr: (sl, hd//2, 2)
     """
     pid = tl.program_id(0).to(tl.int64)
     batch_idx = pid // sl
@@ -194,6 +194,9 @@ def llama4_rope_backward(dq, dk, freqs_cis):
         raise ValueError(f"head_dim must be even for interleaved complex layout, got {hd}")
 
     if freqs_cis.is_complex():
+        freqs_cis = freqs_cis.reshape(-1, freqs_cis.shape[-1])
+        if freqs_cis.shape[0] > sl:
+            freqs_cis = freqs_cis[:sl]
         freqs_cis = torch.view_as_real(freqs_cis)
 
     dq, dk, freqs_cis, compute_dtype = _cast_and_contiguous(dq, dk, freqs_cis)
@@ -227,7 +230,7 @@ def llama4_rope_backward(dq, dk, freqs_cis):
         dk.stride(1),
         dq.stride(2),
         dk.stride(2),
-        freqs_cis.stride(1),
+        freqs_cis.stride(0),
         sl,
         bs,
         n_qh,
--- a/src/liger_kernel/ops/llama4_rope.py
+++ b/src/liger_kernel/ops/llama4_rope.py
@@ -3,72 +3,24 @@ import triton
 import triton.language as tl
 
 
-def _prepare_freqs(freqs_cis: torch.Tensor, seq_len: int, head_dim_half: int):
-    # Split or unpack complex frequencies into real and imag parts
-    if freqs_cis.is_complex():
-        freqs_real = freqs_cis.real
-        freqs_imag = freqs_cis.imag
-    else:
-        # Already split: last dim should be 2*head_dim_half
-        if freqs_cis.shape[-1] == 2 * head_dim_half:
-            freqs_real = freqs_cis[..., :head_dim_half]
-            freqs_imag = freqs_cis[..., head_dim_half:]
-        else:
-            raise ValueError(
-                f"Unexpected freqs_cis shape for non-complex input: {freqs_cis.shape}, expected last dim = {2 * head_dim_half}"
-            )
-
-    # Canonicalize to shape (seq_len, head_dim_half):
-    # 1) Ensure the last dimension is head_dim_half
-    if freqs_real.shape[-1] != head_dim_half:
-        raise ValueError(f"Unexpected last dim for freqs: {freqs_real.shape[-1]} (expected {head_dim_half})")
-    # 2) Flatten all leading dims to a single row dimension
-    freqs_real = freqs_real.reshape(-1, head_dim_half)
-    freqs_imag = freqs_imag.reshape(-1, head_dim_half)
-    # 3) If we have fewer rows than seq_len, allow broadcasting when single row
-    if freqs_real.shape[0] < seq_len:
-        if freqs_real.shape[0] == 1:
-            freqs_real = freqs_real.expand(seq_len, -1)
-            freqs_imag = freqs_imag.expand(seq_len, -1)
-        else:
-            raise ValueError(f"Insufficient rows in freqs: {freqs_real.shape[0]} < seq_len={seq_len}")
-    # 4) If we have more rows than seq_len (e.g., batch present), take the first seq_len rows
-    elif freqs_real.shape[0] > seq_len:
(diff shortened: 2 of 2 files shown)


noemotiovon:
I remember this part also exists in the GPU—can it be modified together as well?
If you modify the GPU code, remember to add the corresponding tests.

Tcc0403:
> I remember this part also exists in the GPU—can it be modified together as well?

Yes, feel free to put it together.

lowdy1:
> > I remember this part also exists in the GPU—can it be modified together as well?
> 
> Yes, feel free to put it together.

Thank you for the suggestion. I have implemented a fused frequency loading pattern in the GPU kernel as shown below, and it achieves slightly better performance in the benchmark on A100:
```
freq_base = seq_idx * freqs_row_stride

# Tiling over dim/2
for d_start in tl.static_range(0, head_dim_half, BLOCK_SIZE):
    d_indices = d_start + tl.arange(0, BLOCK_SIZE)
    mask_d = d_indices < head_dim_half

    # Compute offsets for the block
    freq_offsets = d_indices[:, None] * 2 + tl.arange(0, 2)[None, :]

    # Load the block
    freqs_complex_block = tl.load(
        freqs_complex_ptr + freq_base + freq_offsets,
        mask=mask_d[:, None],
        other=0.0
    )

    freqs_real_block, freqs_imag_block = tl.split(freqs_complex_block)
    freqs_imag_block = freqs_imag_block * imag_sign

```
I’m also wondering whether it would be reasonable to apply a similar approach to `q` and `k`, i.e. loading them once from `q_complex_ptr` / `k_complex_ptr `and then reusing them inside the loop.

Additionally, the current `_prepare_freqs` implementation in `llama4_rope.py` supports both complex and non-complex `freqs_cis` inputs, and canonicalizes them to shape (seq_len, head_dim_half). Although my changes pass all tests, I’m not sure whether this logic should remain in `llama4_rope.py` to canonicalize `freqs_cis` as a complex or be simplified given the updated GPU path.

src/liger_kernel/ops/backends/_ascend/ops/llama4_rope.py
```diff
@@ -195,13 +154,12 @@ def llama4_rope_forward(q, k, freqs_cis):
...
         k,
-        freqs_real,
-        freqs_imag,
+        freqs_cis,
         q.stride(1),
         k.stride(1),
         q.stride(2),
         k.stride(2),
-        freqs_real.stride(0),
+        freqs_cis.stride(1),
```
Tcc0403: What's the shape of `freqs_cis` after `torch.veiw_as_real(freq_cis)`? Why is it `stride(1)`?

Tcc0403:
> Thank you for the suggestion. I have implemented a fused frequency loading pattern in the GPU kernel as shown below, and it achieves slightly better performance in the benchmark on A100:

Great work!

> I’m also wondering whether it would be reasonable to apply a similar approach to q and k, i.e. loading them once from q_complex_ptr / k_complex_ptr and then reusing them inside the loop.

But isn't that we couldn't keep whole hidden dim of q, k tensor in smem the reason why we have to read them in a loop? I might be wrong or misunderstand your question. Feel free to put up your code and I can check the logic directly.

> Additionally, the current _prepare_freqs implementation in llama4_rope.py supports both complex and non-complex freqs_cis inputs, and canonicalizes them to shape (seq_len, head_dim_half). Although my changes pass all tests, I’m not sure whether this logic should remain in llama4_rope.py to canonicalize freqs_cis as a complex or be simplified given the updated GPU path.

I assume `.view_as_real()` only changes its view, i.e. only tensor attributes are touched, so it has no kernel code to run with this op. You can check whether my assumption is true. There will be two solution depending on it:

1. If it does launch an extra kernel for this transformation, we probably want to fuse this operation inside our kernel. It can be done by checking `torch.is_complex` on freq_cis and setting a flag, maybe `IS_COMPLEX`, as kernel parameter, then we run different path in our kernel code. 

2. If not, we can keep it simple as we currently do in this PR, i.e. canonicalize freq_cis in host side

lowdy1:
> But isn't that we couldn't keep whole hidden dim of q, k tensor in smem the reason why we have to read them in a loop? I might be wrong or misunderstand your question. Feel free to put up your code and I can check the logic directly.

I was actually referring to loading `q_real` and `q_imag` together as a single block and then splitting them (similar to freqs), and doing the same for `k` like below. Never mind, I’ve already tested this approach and it doesn’t improve the benchmark performance.

```
for d_start in tl.static_range(0, head_dim_half, BLOCK_SIZE):
    d_indices_qk = d_start * 2 + tl.arange(0, BLOCK_SIZE * 2)
    mask_dqk = d_indices_qk < (head_dim_half * 2)

    if pid_h < n_q_heads:
        q_head_ptr = q_base + pid_h * q_head_stride
        q_pair = tl.load(q_head_ptr + d_indices_qk, mask=mask_dqk, other=0.0)
        q_pair = q_pair.reshape(BLOCK_SIZE, 2, can_reorder=True)
        q_real, q_imag = tl.split(q_pair)
        ...
        tl.store(q_head_ptr + d_indices_qk, new_q_pair, mask=mask_dqk)

```

> I assume `.view_as_real()` only changes its view, i.e. only tensor attributes are touched, so it has no kernel code to run with this op. You can check whether my assumption is true. There will be two solution depending on it:
> 
> 1. If it does launch an extra kernel for this transformation, we probably want to fuse this operation inside our kernel. It can be done by checking `torch.is_complex` on freq_cis and setting a flag, maybe `IS_COMPLEX`, as kernel parameter, then we run different path in our kernel code.
> 2. If not, we can keep it simple as we currently do in this PR, i.e. canonicalize freq_cis in host side

The original shape of `freq_cis` is mostly (bsz, seq_len, head_dim_half). In the current GPU implementation, `_prepare_freqs` is mainly used to canonicalize `freq_cis` into `freqs_real` and `freqs_imag` with shape (seq_len, head_dim_half). However, after calling `.view_as_real()`, the tensor becomes (bsz, seq_len, head_dim_half, 2), which is the main difference here.

Also I think it’s acceptable to preprocess `freq_cis` as shown below to effectively replace the role of `_prepare_freqs` in GPU version, returning `freq_cis` in the shape (seq_len, head_dim_half). After that, `freqs_cis.stride(1)` in kernel calling is replaced with `freqs_cis.stride(0)`
```
if freqs_cis.is_complex():
    freqs_cis = freqs_cis.reshape(-1, freqs_cis.shape[-1])
    if freqs_cis.shape[0] > sl:
        freqs_cis = freqs_cis[:sl]
    freqs_cis = torch.view_as_real(freqs_cis)
```

src/liger_kernel/ops/backends/_ascend/ops/llama4_rope.py
lowdy1: Because without preprocessing, after `torch.veiw_as_real(freq_cis)` is (bsz, seq_len, head_dim_half, 2).

Tcc0403:
Thanks for the explanation.

> Also I think it’s acceptable to preprocess freq_cis as shown below to effectively replace the role of _prepare_freqs in GPU version, returning freq_cis in the shape (seq_len, head_dim_half). After that, freqs_cis.stride(1) in kernel calling is replaced with freqs_cis.stride(0)

Yeah, feel free to apply changes to gpu implementation.

lowdy1:
> Thanks for the explanation.
> 
> > Also I think it’s acceptable to preprocess freq_cis as shown below to effectively replace the role of _prepare_freqs in GPU version, returning freq_cis in the shape (seq_len, head_dim_half). After that, freqs_cis.stride(1) in kernel calling is replaced with freqs_cis.stride(0)
> 
> Yeah, feel free to apply changes to gpu implementation.

Thanks and done.

PR #1070, opened 2026-02-05
[NPU]: NPU-optimized fused_add_rms_norm kernel

TianHao324:
## Summary

1. Because the original kernel uses n_cols as BLOCK_SIZE, and n_cols is smaller in the test, the test can pass normally. However, in the benchmark, n_cols is larger, and when running on the NPU, an ub overflow occurs. Therefore, for each row, we process it in chunks of BLOCK_SIZE.
2. Grid size is limited to NPU core count to avoid resource overflow
3. Each program handles multiple rows
4. Due to the device limitations of the NPU, there is still room for performance improvement. This modification changes it to the NPU-supported format first, but compared to the previous GPU format, the performance has improved significantly.

## Testing Done

image


TianHao324:
@Tcc0403 Sorry, I haven't submitted some performance optimization changes in time. I'm creating a new PR instead. Or should I cancel this merge and let me update the modifications?

PR #1087, opened 2026-02-09
[NPU] Add softmax implementation

lowdy1:
## Summary
This PR adds a Softmax implementation for NPU. It includes a single-block forward kernel for smaller column sizes, as well as a multi-block kernel for large column sizes to avoid NPU UB overflow.

## Testing Done
Test done with `python -m pytest test/transformers/test_softmax.py`
Hardware Type: Atlas 800I A2(32G)

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ops/softmax.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/softmax.py
@@ -16,6 +16,21 @@ def _softmax_multi_block_forward_kernel(
     n_cols,
     BLOCK_SIZE: tl.constexpr,
 ):
+    """
+    Multi-block softmax forward kernel using two-pass algorithm.
+
+    First pass computes max and sum for numerical stability.
+    Second pass normalizes and writes output.
+
+    Args:
+        Y_ptr: Output tensor pointer
+        Y_row_stride: Stride for output rows
+        X_ptr: Input tensor pointer
+        X_row_stride: Stride for input rows
+        n_rows: Number of rows to process
+        n_cols: Number of columns per row
+        BLOCK_SIZE: Block size for column processing
+    """
     row_start = tl.program_id(0)
     row_step = tl.num_programs(0)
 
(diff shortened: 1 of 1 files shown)


src/liger_kernel/ops/backends/_ascend/ops/softmax.py
```diff
@@ -0,0 +1,214 @@
...
+
+    y2d = torch.empty_like(x2d)
+    num_cores = get_npu_core_count()
+    num_programs = min(num_cores, n_rows)
+
+    if n_cols <= BLOCK_SIZE:
+        softmax_forward_kernel[(num_programs,)](y2d, x2d, x2d.stride(0), y2d.stride(0), n_rows, n_cols, BLOCK_SIZE)
+        multi_block_launch = False
+    else:
+        softmax_multi_block_forward_kernel[(n_rows,)](
```
noemotiovon: In Triton kernel implementations on NPU, it’s best for the first dimension of the grid to be `num_cores`.

src/liger_kernel/ops/backends/_ascend/ops/softmax.py
```diff
@@ -0,0 +1,214 @@
...
+            y2d,
+            dy2d.stride(0),
+            y2d.stride(0),
+            dx2d.stride(0),
+            n_rows,
+            n_cols,
+            BLOCK_SIZE=BLOCK_SIZE,
+        )
+    else:
+        softmax_multi_block_backward_kernel[(n_rows,)](
```
noemotiovon: ditto

src/liger_kernel/ops/backends/_ascend/ops/softmax.py
```diff
@@ -0,0 +1,214 @@
...
+
+    BLOCK_SIZE = triton.next_power_of_2(n_cols)
+    BLOCK_SIZE = min(BLOCK_SIZE, MAX_FUSED_BLOCK_SIZE)
+
+    y2d = torch.empty_like(x2d)
+    num_cores = get_npu_core_count()
+    num_programs = min(num_cores, n_rows)
+
+    if n_cols <= BLOCK_SIZE:
+        softmax_forward_kernel[(num_programs,)](y2d, x2d, x2d.stride(0), y2d.stride(0), n_rows, n_cols, BLOCK_SIZE)
```
noemotiovon: I don’t think we need two separate kernels. Also, the current approach of launching the kernel row by row is not optimal. It’s clear that the UB (unified buffer) isn’t being fully utilized.

I believe we should use a single unified kernel, with two nested loops inside: the outer loop iterates over rows, and the inner loop iterates over columns. If one kernel can process multiple rows, it can make better use of the UB.

However, this might introduce some additional overhead, so we still need to benchmark and see whether it actually improves performance.

Tcc0403: Agreed, multi-block kernel plus grid stride loop over rows by `num_prorams` should work.

lowdy1: Thanks for your suggestions.

lowdy1: @Tcc0403 would you mind reviewing this again?

PR #1096, opened 2026-02-11
Cleanup deprecated code for older versions of transformers.

vaibhavjindal:
## Summary

Cleanup deprecated code for older versions of transformers. We will only support versions >=4.52.0 from now on.

## Testing Done

Will get tested in [transformers-5.0.0rc1](https://github.com/linkedin/Liger-Kernel/tree/transformers-5.0.0rc1) branch

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/transformers/model/gemma.py
+++ b/src/liger_kernel/transformers/model/gemma.py
@@ -6,14 +6,12 @@ from typing import Union
 import torch
 
 from transformers.cache_utils import Cache
-from transformers.utils.deprecation import deprecate_kwarg
 
 from liger_kernel.transformers.model.loss_utils import LigerForCausalLMLoss
 from liger_kernel.transformers.model.loss_utils import unpack_cross_entropy_result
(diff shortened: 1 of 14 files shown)


src/liger_kernel/transformers/model/gemma.py
```diff
@@ -5,133 +5,14 @@
...
-    return CausalLMOutputWithPast(
-        loss=loss,
-        logits=logits,
-        past_key_values=outputs.past_key_values,
-        hidden_states=outputs.hidden_states,
-        attentions=outputs.attentions,
-    )
-
-
 @deprecate_kwarg("num_logits_to_keep", version="4.50", new_name="logits_to_keep")
```
Mecoli1219: Do we still need it as we don't support version less than 4.52.0 right now? Similar to other files

vaibhavjindal: Makes sense, let me remove it as well

Mecoli1219 (approved):
LGTM. Thanks!

PR #1112, opened 2026-02-27
Optimize Group norm forward kernel

vaibhavjindal:
## Summary

This PR refactors the **GroupNorm forward Triton kernel** to apply affine parameters (`W`, `B`) using a **single flat loop over the group’s hidden dimension** instead of looping **channel-by-channel**.

* Previously: iterate over channels in the group, then iterate over each channel’s contiguous slice, reusing one `W/B` per channel.
* Now: iterate once over all elements in the group and compute the corresponding channel via `channel = offset // hidden_size_per_channel`, then select `W/B` from that.
* Mean/RSTD computation and normalization semantics stay the same, and masking still guards out-of-bounds elements.
* Net effect: simpler indexing and better handling when per-channel slices are small, at the cost of potentially more frequent `W/B` loads (gathers) compared to the per-channel reuse before.

Gives about **~2x speedup** in the forward kernel compared to old liger kernel, without any memory changes.
Fwd speed:

group_norm_speed_forward

Full speed:

group_norm_speed_full

Memory:

group_norm_memory_full

## Testing Done

Yes, on H100.

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ops/fused_add_rms_norm.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/fused_add_rms_norm.py
@@ -23,7 +23,117 @@ def torch_dtype_to_triton(dtype):
 
 
 # -----------------------------------------------------------------------------
-# Forward Kernel - Grid-Stride Loop with Row-wise Processing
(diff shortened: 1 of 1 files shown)


Mecoli1219 (approved):
LGTM!

Issue #1051, opened 2026-01-29
[RFC] Standardized Model Config for Benchmarks


Tcc0403:
### 🚀 The feature, motivation and pitch

As we add support for more devices, benchmarking scripts increasingly rely on device-specific tensor shapes due to VRAM constraints. This leads to fragmented benchmark setups and scattered performance results that are difficult to compare or reproduce. To make benchmarking more scalable and consistent, I propose introducing a standardized benchmark model configuration.

### Motivation

Today, different kernels (and sometimes different devices) use ad-hoc shapes, which:

- Require device-specific dispatching in benchmark scripts
- Produce results that are hard to compare across hardware
- Increase maintenance burden for contributors adding new benchmarks

### Proposal

Define one or more representative “mainstream” model profiles (e.g., LLaMA-/GPT-like) with canonical parameters such as: `hidden_size`, `vocab_size`, `num_q_heads`, `num_kv_heads`, etc.

All benchmark scripts would derive their shapes from this shared config, optionally using scaled-down subsets when needed to fit memory constraints, instead of inventing per-device shapes.

This would:

- Eliminate ad-hoc device-specific shapes
- Improve comparability across kernels and hardware
- Lower the barrier for contributors writing new benchmarks
- Improve reproducibility of performance results

### Target Devices

Before finalizing the standardized config, it would be helpful to clarify which devices we officially want to support for benchmarking.

Here’s the list we currently have:

- NVIDIA H100 (80GB)
- Intel XPU GPU Max 1100 (48GB)
- NPU Atlas 900 A2 POD (64G)
- 
Please feel free to comment if I missed any devices, or if there are additional targets we should consider.

### Next Steps

- [ ]  Confirm target benchmarking devices
- [ ]  Agree on one or more baseline model profiles (e.g., LLaMA, Qwen, Gemma, etc)
- [ ]  Define canonical parameters (hidden_size, vocab_size, heads, etc.)
- [ ]  Specify scaling rules for memory-constrained devices
- [ ]  Refactor existing benchmarks to consume the standardized config

Tcc0403:
@zheliuyu what's the target npu model we are planning to add for benchmarking

zheliuyu:
@Tcc0403 The NPU models currently in use are of two types:

- Atlas 900 A2 POD (64G), which can be used for benchmark tasks.
- Atlas 800I A2 (32G), used by CI machines. This model will not be used for benchmark work.
[...]

PR #1116, opened 2026-03-02
[Test]: Refactor benchmark_geglu with standardized model configs

noemotiovon:
Introduce `benchmark_model_configs.py` with canonical LLM architecture profiles (`LLAMA_2_7B`, `LLAMA_3_8B`) and two utilities:

- **`estimate_kernel_bytes_per_token`**: runtime probe that measures peak GPU memory per token for a given kernel, providing a safe upper bound.
- **`compute_benchmark_shape`**: derives batch size and sequence length from available device memory, the model profile, and the measured bytes-per-token, replacing hardcoded shapes.

Add **`run_speed_benchmark`** and **`run_memory_benchmark`** helpers to `utils.py` to eliminate repeated `triton.testing` / `_test_memory` blocks across scripts. Also add a `--model` CLI argument to `parse_benchmark_script_args` so callers can select a model profile at runtime.

Refactor `benchmark_dyt.py`, `benchmark_geglu.py`, and `benchmark_swiglu.py` to:

- Extract a **setup factory function** to avoid duplicating layer/tensor construction between speed and memory benchmarks.
- Delegate to `run_speed_benchmark` / `run_memory_benchmark`.
- Compute `x_values`, `bsz`, and related fields dynamically via `compute_benchmark_shape` instead of hardcoded constants.
- Expand memory benchmark modes from `["full"]` to `["full", "forward", "backward"]` for consistency with speed benchmarks.

Hardware Type: NVIDIA A100-SXM4-80GB

### DIFF
--- a/benchmark/scripts/benchmark_dyt.py
+++ b/benchmark/scripts/benchmark_dyt.py
@@ -3,6 +3,9 @@ import sys
 
 import torch
 
+from benchmark_model_configs import compute_hidden_size_sweep_config
+from benchmark_model_configs import estimate_kernel_peak_memory
+from benchmark_model_configs import get_benchmark_model_config
 from utils import SingleBenchmarkRunInput
 from utils import SingleBenchmarkRunOutput
 from utils import parse_benchmark_script_args
@@ -46,17 +49,34 @@ def bench_memory_dyt(input: SingleBenchmarkRunInput) -> SingleBenchmarkRunOutput
     return run_memory_benchmark(lambda: layer(x), input.kernel_operation_mode)
 
 
+BT = 4096
+
 if __name__ == "__main__":
     args = parse_benchmark_script_args()
+    model = get_benchmark_model_config(args.model)
 
     for beta in [False, True]:
+
+        def _probe():
+            probe_input = SingleBenchmarkRunInput(
+                x=model.hidden_size,
+                kernel_provider="torch",
+                extra_benchmark_config={"BT": BT, "dtype": model.dtype, "beta": beta},
+            )
+            x, layer = _setup_dyt(probe_input)
+            return layer(x)
+
+        peak_bytes = estimate_kernel_peak_memory(probe_fn=_probe)
+        sweep_config = compute_hidden_size_sweep_config(model, peak_bytes, bt=BT)
+        x_values = [1024 * i for i in range(1, 17) if 1024 * i <= sweep_config.max_hidden_size] or [model.hidden_size]
+
         common_configs = {
             "kernel_name": f"dyt_beta={beta}",
             "x_name": "hidden_size",
             "x_label": "hidden_size",
-            "x_values": [1024 * i for i in range(1, 17)],
+            "x_values": x_values,
             "kernel_providers": ["liger", "torch", "torch_compile"],
-            "extra_benchmark_configs": [{"BT": 4096, "dtype": torch.bfloat16, "beta": beta}],
+            "extra_benchmark_configs": [{"BT": sweep_config.bt, "dtype": model.dtype, "beta": beta}],
             "overwrite": args.overwrite,
         }
 
--- a/benchmark/scripts/benchmark_geglu.py
+++ b/benchmark/scripts/benchmark_geglu.py
@@ -2,10 +2,9 @@ import math
 
 import torch
 
-from benchmark_model_configs import DEFAULT_MODEL_CONFIG
-from benchmark_model_configs import MODEL_REGISTRY
-from benchmark_model_configs import compute_benchmark_shape
-from benchmark_model_configs import estimate_kernel_bytes_per_token
+from benchmark_model_configs import compute_seq_len_sweep_config
+from benchmark_model_configs import estimate_kernel_peak_memory
+from benchmark_model_configs import get_benchmark_model_config
 from transformers.models.llama.configuration_llama import LlamaConfig
 from transformers.models.llama.modeling_llama import LlamaMLP
 from utils import SingleBenchmarkRunInput
@@ -17,7 +16,6 @@ from utils import run_speed_benchmark
 
 from liger_kernel.transformers.geglu import LigerGEGLUMLP
 from liger_kernel.utils import infer_device
-from liger_kernel.utils import get_total_gpu_memory
 
 device = infer_device()
 
@@ -60,43 +58,38 @@ def bench_memory_geglu(input: SingleBenchmarkRunInput) -> SingleBenchmarkRunOutp
 if __name__ == "__main__":
     args = parse_benchmark_script_args()
 
-    model = MODEL_REGISTRY[args.model] if args.model else DEFAULT_MODEL_CONFIG
-    total_memory_gb = get_total_gpu_memory()
-
+    model = get_benchmark_model_config(args.model)
     probe_seq_len = 1024
-    probe_input = SingleBenchmarkRunInput(
-        x=probe_seq_len,
-        kernel_provider="huggingface",
-        extra_benchmark_config={
-            "bsz": 1,
-            "hidden_size": model.hidden_size,
-            "intermediate_size": model.intermediate_size,
-            "hidden_act": "gelu_pytorch_tanh",
-            "dtype": model.dtype,
-        },
-    )
-    probe_x, probe_layer = _setup_geglu(probe_input)
-    kernel_bpt = estimate_kernel_bytes_per_token(
-        kernel_fn=lambda: probe_layer(probe_x),
-        num_tokens=probe_seq_len,
-    )
-    del probe_x, probe_layer
 
-    shape = compute_benchmark_shape(
-        total_memory_gb,
-        model,
-        kernel_bytes_per_token=kernel_bpt,
-    )
+    def _probe():
+        probe_input = SingleBenchmarkRunInput(
+            x=probe_seq_len,
+            kernel_provider="huggingface",
+            extra_benchmark_config={
+                "bsz": 1,
+                "hidden_size": model.hidden_size,
+                "intermediate_size": model.intermediate_size,
+                "hidden_act": "gelu_pytorch_tanh",
+                "dtype": model.dtype,
+            },
+        )
+        x, layer = _setup_geglu(probe_input)
+        return layer(x)
+
+    peak_bytes = estimate_kernel_peak_memory(probe_fn=_probe)
+    kernel_bpt = peak_bytes // probe_seq_len
+
+    config = compute_seq_len_sweep_config(model, kernel_bytes_per_token=kernel_bpt)
 
     common_configs = {
         "kernel_name": "geglu",
         "x_name": "T",
         "x_label": "sequence length",
-        "x_values": [2**i for i in range(10, int(math.log2(shape.seq_len)) + 1)],
+        "x_values": [2**i for i in range(10, int(math.log2(config.seq_len)) + 1)],
         "kernel_providers": ["liger", "huggingface"],
         "extra_benchmark_configs": [
             {
-                "bsz": shape.batch_size,
+                "bsz": config.batch_size,
                 "hidden_size": model.hidden_size,
                 "intermediate_size": model.intermediate_size,
                 "hidden_act": "gelu_pytorch_tanh",
--- a/benchmark/scripts/benchmark_model_configs.py
+++ b/benchmark/scripts/benchmark_model_configs.py
@@ -8,20 +8,21 @@ shared configs rather than defining ad-hoc per-script constants.
 Usage::
 
     from benchmark_model_configs import (
-        MODEL_REGISTRY, DEFAULT_MODEL_CONFIG,
-        compute_benchmark_shape,
-        estimate_kernel_bytes_per_token,
+        get_benchmark_model_config,
+        compute_seq_len_sweep_config,
+        estimate_kernel_peak_memory,
     )
 
     args = parse_benchmark_script_args()
-    model = MODEL_REGISTRY[args.model] if args.model else DEFAULT_MODEL_CONFIG
-    total_memory_gb = get_total_gpu_memory()
+    model = get_benchmark_model_config(args.model)
 
-    # Measure actual memory via a small probe, then compute safe shapes
-    bpt = estimate_kernel_bytes_per_token(kernel_fn=..., num_tokens=1024)
-    shape = compute_benchmark_shape(total_memory_gb, model, kernel_bytes_per_token=bpt)
+    # Measure actual memory via a small probe, then compute sweep config
+    peak_bytes = estimate_kernel_peak_memory(probe_fn=_probe)
+    bpt = peak_bytes // probe_num_tokens
+    config = compute_seq_len_sweep_config(model, kernel_bytes_per_token=bpt)
 """
 
+import gc
 import math
 
 from dataclasses import dataclass
@@ -30,6 +31,9 @@ from typing import Dict
 from typing import Optional
 
 import torch
+
+from liger_kernel.utils import get_total_gpu_memory
+from liger_kernel.utils import infer_device
 
 
 @dataclass(frozen=True)
@@ -56,20 +60,29 @@ class ModelConfig:
 
 
 @dataclass(frozen=True)
-class BenchmarkShapeConfig:
-    """Computed benchmark shape parameters.
-
-    Returned by :func:`compute_benchmark_shape` — these values are derived
-    from the combination of device memory, model dimensions, and kernel-specific
-    memory characteristics rather than being hardcoded per device.
+class SeqLenSweepConfig:
+    """Config for benchmarks that sweep sequence length (e.g. GEGLU, SwiGLU).
 
     Attributes:
-        batch_size: Safe batch size for the given configuration.
-        seq_len: Max sequence length for benchmark sweeps.
+        batch_size: Safe batch size for the sweep.
+        seq_len: Max sequence length (upper bound for x_values).
     """
 
     batch_size: int
     seq_len: int
+
+
+@dataclass(frozen=True)
+class HiddenSizeSweepConfig:
+    """Config for benchmarks that sweep hidden_size with fixed BT (e.g. DyT).
+
+    Attributes:
+        bt: Fixed batch * seq dimension.
+        max_hidden_size: Upper bound for hidden_size sweep.
+    """
+
+    bt: int
+    max_hidden_size: int
 
 
 # ── Model Profiles ──────────────────────────────────────────────────────────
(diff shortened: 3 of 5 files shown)


noemotiovon:
Hi @Tcc0403, I’ve reviewed your refactoring proposal for the benchmark. I’ve tried implementing it based on the issue — could we discuss what further optimizations might be needed?
Related issue: https://github.com/linkedin/Liger-Kernel/issues/1051

benchmark/scripts/benchmark_model_configs.py
```diff
@@ -0,0 +1,155 @@
...
+        device_name: Human-readable device identifier.
+        batch_size: Max comfortable batch size for this device.
+        seq_len: Max sequence length for benchmark sweeps.  Used as the upper
+            bound when generating sweep x_values, or as a fixed sequence length
+            parameter when the sweep axis is something else (e.g. vocab_size).
+    """
+
+    device_name: str
+    batch_size: int
+    seq_len: int
```
Tcc0403: I don't think `batch_size` and `seq_len` should be constant values here because they are bounded by model related configs and kernel implementations as well. Assuming that we have a large memory footprint kernel in future, OOM might still occur with hardcoding bs/sl. They should be computed after getting our target model, kernel and device. wydt?

noemotiovon: Agreed, hardcoding `batch_size/seq_len` per device is fragile. I've refactored `DeviceBenchmarkConfig` to store `total_memory_gb` instead, and added a `compute_benchmark_shape(device_cfg, model_cfg, kernel_bytes_per_token=...)` function that derives safe `batch_size/seq_len` from the combination of device memory, model dimensions (including `max_position_embeddings`), and a kernel-specific memory estimate.

However, I'd like your input on how to handle the `kernel_bytes_per_token` estimation. The challenge is:

1. Kernel memory is hard to compute analytically — it depends on the number and size of intermediate tensors, what autograd saves for backward, and whether the kernel fuses operations to reduce peak memory. For GEGLU, I estimated `~20x intermediate_size * dtype_bytes` per token, but this is a rough heuristic, not a precise derivation.
2. Different devices have different memory behaviors — the same kernel on CUDA vs NPU may have different peak memory due to allocator strategies, framework overhead, and operator implementations.

A few options I can think of:

- (A) Keep the analytical heuristic (current approach) — each benchmark script provides a conservative `kernel_bytes_per_token` estimate. Simple but imprecise.
- (B) Empirical calibration — run a small probe (e.g., single batch, small `seq_len`) to measure actual peak memory, then extrapolate. More accurate but adds runtime overhead and complexity.

What approach do you think fits best here? Or is there a simpler way I'm missing?

Tcc0403: (B) makes sense to me. I believe most OOMs occur in pure pytorch implementations. To estimate the upper bound, we only have to run once with 1024 tokens with torch's impl to get `kernel_bytes_per_token`, which shouldn't take too long.

Requiring developers to hardcode `kernel_bytes_per_token` is fragile as well.

noemotiovon: I’ve implemented a preliminary version following approach B. Could you take a look and see whether it aligns with your expectations? If it’s broadly in line with the intended direction, I’ll refine the code further and provide the corresponding design document as well.

Tcc0403: lgtm, let's see how it applies to other kernels

benchmark/scripts/benchmark_model_configs.py
```diff
@@ -0,0 +1,310 @@
...
+
+    Attributes:
+        device_name: Human-readable device identifier.
+        total_memory_gb: Total device memory in gigabytes.  Used together with
+            model and kernel information to compute safe ``batch_size`` /
+            ``seq_len`` via :func:`compute_benchmark_shape`.
+    """
+
+    device_name: str
+    total_memory_gb: float
```
noemotiovon: I believe the current device management is actually redundant. In the current computation, the only factor that affects the benchmark test shape is the global memory, and the global memory can be obtained directly through the PyTorch API. Therefore, I will remove this part of the code.

noemotiovon:
Hi @Tcc0403,  I’ve completed the benchmark refactor for element-wise level kernels.

At the moment, it seems that the only device-dependent factor is global memory, so I’ve temporarily removed the previously agreed device-specific checks.

Could you help review the code?

benchmark/scripts/benchmark_model_configs.py
```diff
@@ -0,0 +1,215 @@
...
+    import gc
+
+    if torch.cuda.is_available():
+        device_str = "cuda"
+    elif hasattr(torch, "npu") and torch.npu.is_available():
+        device_str = "npu"
+    else:
+        raise RuntimeError(
+            "No CUDA or NPU device available for memory measurement"
+        )
```
Tcc0403: call `infer_device` to prevent missing other devices
https://github.com/linkedin/Liger-Kernel/blob/27430046d73e770ea6992fcee1aeaa9b8930866f/src/liger_kernel/utils.py#L42

benchmark/scripts/benchmark_model_configs.py
```diff
@@ -0,0 +1,215 @@
...
+
+    Args:
+        total_memory_gb: Total device memory in gigabytes, typically from
+            :func:`~liger_kernel.utils.get_total_gpu_memory`.
+        model_cfg: Model architecture config.
+        kernel_bytes_per_token: Peak memory **per token** (``batch * seq_len``
+            axis).  Best obtained from :func:`estimate_kernel_bytes_per_token`.
+            Falls back to a conservative heuristic
+            (``hidden_size * dtype_bytes * 16``) when *None*.
+        memory_utilization: Fraction of total device memory to target (0\u20131).
```
Tcc0403: what's `(0\u20131)`

benchmark/scripts/benchmark_dyt.py
Tcc0403: missing a probe run for upperbound?

benchmark/scripts/benchmark_geglu.py
```diff
@@ -1,163 +1,106 @@
...
+            "hidden_act": "gelu_pytorch_tanh",
+            "dtype": model.dtype,
+        },
     )
+    probe_x, probe_layer = _setup_geglu(probe_input)
+    kernel_bpt = estimate_kernel_bytes_per_token(
+        kernel_fn=lambda: probe_layer(probe_x),
+        num_tokens=probe_seq_len,
+    )
+    del probe_x, probe_layer
```
Tcc0403: I believe input setup and cleanup should be done in `estimate_kernel_bytes_per_token`, so developers don't have to worry about them.

benchmark/scripts/benchmark_geglu.py
```diff
@@ -1,163 +1,106 @@
...
-        mem_50, mem_20, mem_80 = _test_memory(
-            full,
-            quantiles=QUANTILES,
-        )
-
-    return SingleBenchmarkRunOutput(
-        y_20=mem_20,
-        y_50=mem_50,
-        y_80=mem_80,
+    model = MODEL_REGISTRY[args.model] if args.model else DEFAULT_MODEL_CONFIG
```
Tcc0403: Should we create a function `get_model(model_name)` for readability?

noemotiovon:
I think my previous thinking wasn’t quite right. I may need to reorganize my thoughts.

benchmark/scripts/benchmark_geglu.py
noemotiovon: Good idea

benchmark/scripts/benchmark_dyt.py
noemotiovon: At first I thought that since hidden_size is a model parameter, it might not need to be probed. However, I now think that for cases where we need to avoid VRAM eviction, we should consistently scale based on x_values. A new probe has already been added here.

At the same time, I also realized that not all kernels are related to seq_len. There may be other types of kernels in the future, and for different kinds of kernels we should define matching configurations. I’d also be happy to implement this part in the future.

benchmark/scripts/benchmark_geglu.py
noemotiovon: Thanks for your suggestion!

noemotiovon:
Hi @Tcc0403, I have now completed the revisions based on the review comments above and made some adjustments to the code. First, I want to confirm that our rule is correct: we should scale based on the x_values obtained from the probe.

For the kernels I’m currently handling, each has a single x_value, but there are two situations:

1. The x_value is not a model parameter, for example seq_len. In this case, we can use the probe to compute a reasonable seq_len.

2. The x_value is a model parameter, for example hidden_size. In this case, we probably shouldn’t directly use the model parameter as the x-axis, because hidden_size is fixed within a given model. That means this kind of kernel might not really depend on the specific model. So instead, we could also infer some reasonable hidden_size values through probing. Is this the right approach?

Tcc0403:
> First, I want to confirm that our rule is correct: we should scale based on the x_values obtained from the probe.

As you mentioned, there are two situations, each targetting different metrics: 
1. `token length as variable, fixed model config`: performance across different token length (token length as x_value)
2. `model config as variable, fixed token length`: performance across different model architecture (model config as x_value)

Ideally, we want both workloads covered for every operators in our benchmark suite, so it can have comparisons in different scenarios. However, we didn't set a clear guideline at first, leaving contributors setting arbitrary configs when writing benchmark scripts. (This is also one of the reasons why I want to initiate the discussion on refactoring benchmark suite)

My bottom line is avoiding OOM on all running cases to have a fair comparison among different devices. It would be perfect if two situations can be handled. 

My current design in mind for each situations is:
1. `token length as variable, fixed model config`: what we currently do, i.e. run a probe to determine the upper bound of x_value
2. `model config as variable, fixed token length`: find a token length so that all given model configs are in the range of x_value

Take LigerGEGLUMLP as an example:
Inputs can be categorized into two groups as we mentioned: 1) token length related: batch_size, seq_len and 2) model related: hidden_size, intermediate_size

For `token length as variable, fixed model config` workload, what we currently do is fine, which is:
1. pick a model to retrieve model related configs, and set them as fixed parameters
2. run a probe with `token length=1024` to find the upper bound

For `model config as variable, fixed token length` workload, there are some problems worth discussing beforehand:
1. There might be mutliple candidates from model config that can be selected as x_value, `hidden_size` and `intermediate_size` in this case. Which one should we pick?
2. Should this fixed token length be universal across all ops? or op-specific?

Some potential solutions in my mind for each problems:
+ 1A) iteratively select one and fix the rest: there will be *two* iterations in LigerGEGLUMLP example, `(scalable hidden size, fixed intermediate size)` and `(fixed hidden size, scalable intermediate size)`, generating *two* figures with each parameter as x-axis
	- this approach shows the performance of scaling trend on the selected parameter
	- new problem comes to the surface: what fixed number should be set to? (probably set to same values from the given model config)
		- model dependent, model name required
+ 1B) scale all parameters, examining n-dimensional data points: 
	- rich datapoints can generate interactive figures with different configurations 
	- time-consuming and bloated benchmark data
	- most detailed but not realistic in my opinion
	- independent to model
+ 1C) run all model configs we have defined: for example, `LLAMA_2_7B:(hidden_size=4096, intermediate_size=11008,)` and `LLAMA_3_8B:(hidden_size=4096, intermediate_size=14336,)`
	- cost efficient
	- lack of scaling trends
	- speedup or throughput comparison instead of orginal line chart, some examples:
		- throughput: https://github.com/Dao-AILab/quack/blob/main/media/bf16_kernel_benchmarks_single_row.svg
		- speedup: https://github.com/Dao-AILab/quack/blob/main/benchmarks/visual_outputs/quack_vs_pytorch_speedup.png, but model config tuples as x-axis
+ 2) I prefer one universal token length for all operators, or coupled with model configs, such as `(BxT=2048, llama_2_7B)`, `(BxT=1024, llama_3_8B)`
	- figures can be aligned nicely as https://github.com/Dao-AILab/quack/blob/main/media/combined_kernel_benchmarks_final.png does
	- how do we ensure the token length is runnable for all ops? (maybe a dedicated test suite for benchmark)
	- how do we maintain forward-compatible with future ops? (might be overthinking, probably handle it when this actually happens)

TL;DR: solution 1B or 1C or both + 2
Since it seems to be a big change, I suggest creating a new directory before merging into ci workflows

That's my current thought, happy to discuss! Any feedback is welcome. Let me know if there's anything that needs to be considered as well.

noemotiovon:
Thanks for the thorough review and the well-structured breakdown — this really clarifies the bigger picture.

I fully agree that we need both dimensions covered:

- D1 (token-length sweep): performance scaling across different input sizes with a fixed model config — this is what the current PR delivers.
- D2 (model-config sweep): performance comparison across different model architectures at a fixed input scale — not yet implemented.

On the D2 design options, I'm aligned with your preference:

- Solution C (discrete model configs) as the primary approach — cost-efficient, produces clean bar charts (speedup/throughput) and naturally extends the existing `MODEL_REGISTRY`.
- Solution A (per-parameter sweep) as an optional enrichment for ops where individual parameter scaling trends matter.
- For universal token length, I lean towards a probe-based approach: define a candidate `BxT` per model config, then run `estimate_kernel_peak_memory` across all `(model_config, token_config)` pairs to verify safety. This keeps the framework forward-compatible with new ops — if a config would OOM, the framework reduces tokens automatically rather than crashing.

*a phased plan:*

| Phase   | Scope                                                        | Status                   |
| :------ | :----------------------------------------------------------- | :----------------------- |
| Phase 1 | D1 foundation: `ModelConfig`, probe, sweep helpers, refactored scripts, guidelines | This PR (ready to merge) |
| Phase 2 | D2 model-config sweep: new helper, `--sweep-mode` CLI, reference impl in `benchmark_geglu.py` | Follow-up PR             |
| Phase 3 | Rollout D2 to all ops, bar chart visualization, CI integration, more model profiles | Follow-up PR(s)          |

Since Phase 1 is self-contained and Phase 2/3 build on top of it, would it make sense to merge this PR as the Phase 1 baseline and then iterate on D2 in a follow-up? Happy to start on Phase 2 immediately after.

Let me know if there's anything you'd like adjusted in the design doc or the phasing. Looking forward to pushing this forward together!

noemotiovon:
# Benchmark Framework Design

This document describes the overall design of the Liger-Kernel benchmark suite, including its two benchmark dimensions, the shared infrastructure, and the phased implementation plan.

## 1. Benchmark Dimensions

Every operator should ideally be benchmarked along two orthogonal dimensions:

| Dimension | x-axis | Fixed | Goal |
|-----------|--------|-------|------|
| **D1: Token-length sweep** | sequence length (or BT) | model architecture (hidden_size, intermediate_size, ...) | Performance scaling across different input sizes |
| **D2: Model-config sweep** | model architecture parameter(s) | token length | Performance scaling across different model architectures |

### D1: Token-length sweep (implemented)

Pick a single model config from `MODEL_REGISTRY`, probe peak memory to determine safe upper bounds, then sweep sequence length (or BT for element-wise ops). This is the current default behavior for all benchmark scripts.

```
x_values:  [1024, 2048, 4096, 8192, ...]
fixed:     model=llama_3_8b (hidden_size=4096, intermediate_size=14336, ...)
output:    line chart — speed/memory vs token length
```

### D2: Model-config sweep (planned)

Fix the token-length dimension, then vary model architecture parameters across predefined `MODEL_REGISTRY` entries. This reveals how kernel performance compares across different model architectures at the same input scale.

```
x_values:  [llama_2_7b, llama_3_8b, ...]  (discrete model configs)
fixed:     BxT (determined to be safe across all configs)
output:    speedup or throughput bar chart per model config
```

## 2. D2 Design Choices

Following the maintainer discussion, we evaluated three approaches:

| Approach | Description | Pros | Cons |
|----------|-------------|------|------|
| **A: Per-parameter sweep** | Fix all but one model param, sweep it (e.g. sweep hidden_size with fixed intermediate_size, then vice versa) | Shows per-parameter scaling trend | Combinatorial; fixed values are arbitrary; model-dependent |
| **B: N-dimensional scan** | Vary all model parameters simultaneously | Most comprehensive data | Impractical runtime; data bloat |
| **C: Discrete model configs** | Run each `MODEL_REGISTRY` entry as one data point | Cost-efficient; realistic configs; easy to maintain | No continuous scaling trend |

**Decision: C as the primary approach**, with A as optional enrichment for ops where single-parameter scaling is important.

Rationale:
- C uses real-world model architectures, making results directly meaningful.
- C naturally extends the existing `MODEL_REGISTRY` infrastructure.
- C produces clean bar charts (speedup/throughput) that align across ops.
- A can be layered on later for specific ops where parameter-level scaling trends matter.

## 3. Universal Token Length for D2

For D2 benchmarks, we need a fixed token-length that is safe (no OOM) across all model configs and all operators.

### Strategy

1. **Coupled tokens**: define a `(batch_size, seq_len)` pair per model config, e.g. `(B=2, T=1024)` for llama_2_7b, `(B=1, T=1024)` for llama_3_8b. This allows adapting to each model's memory footprint while keeping the comparison fair (same total token count or same seq_len).
2. **Safety via probe**: before running D2 benchmarks, run `estimate_kernel_peak_memory` for each `(model_config, token_config)` pair. If any config would OOM, reduce token count automatically.
3. **Forward compatibility**: when new ops are added, the probe mechanism ensures safety without manual tuning. If a new op cannot fit any reasonable token count for a given model, the framework skips that data point with a warning rather than crashing.

### Proposed CLI

```bash
# D1 (existing): token-length sweep with fixed model
python benchmark_geglu.py --model llama_3_8b

# D2 (new): model-config sweep with fixed token length
python benchmark_geglu.py --sweep-mode model_config --bt 2048
```

The `--sweep-mode` flag selects the dimension. Default remains `token_length` (D1) for backward compatibility.

## 4. Infrastructure Changes

### 4.1 New config type

```python
@dataclass(frozen=True)
class ModelConfigSweepConfig:
    """Config for D2 benchmarks that sweep across model configs."""
    model_configs: List[ModelConfig]       # models to benchmark
    bt: int                                 # fixed batch * seq_len
    batch_size: int                         # safe batch size
    seq_len: int                            # safe seq_len
```

### 4.2 New helper

```python
def compute_model_config_sweep_config(
    model_configs: List[ModelConfig],
    probe_fn_factory: Callable[[ModelConfig, int], Callable[[], torch.Tensor]],
    bt: int = 2048,
    memory_utilization: float = 0.4,
) -> ModelConfigSweepConfig:
    """Find safe (batch_size, seq_len) that works across all model configs.

    For each model config, runs probe_fn_factory(model_config, bt) to measure
    peak memory, then picks the most conservative batch_size / seq_len.
    """
    ...
```

### 4.3 Script-level changes

Each benchmark script gains a D2 code path gated by `--sweep-mode`:

```python
if args.sweep_mode == "model_config":
    configs = [MODEL_REGISTRY[name] for name in MODEL_REGISTRY]
    sweep = compute_model_config_sweep_config(configs, probe_fn_factory=..., bt=args.bt)
    # x_values = model config names
    # extra_benchmark_configs = one per model config
    ...
else:
    # existing D1 logic
    ...
```

### 4.4 Visualization

D2 results produce **grouped bar charts** (speedup or throughput) rather than line charts:
- x-axis: model config names (e.g. `llama_2_7b`, `llama_3_8b`)
- bars: kernel providers (liger vs huggingface/torch)
- y-axis: speedup ratio or throughput (tokens/s)

## 5. Phased Implementation Plan

### Phase 1: Foundation (current PR)

**Status: complete**

- `ModelConfig` and `MODEL_REGISTRY` with canonical model profiles
- `estimate_kernel_peak_memory()` for runtime memory probing
- `compute_seq_len_sweep_config()` for D1 seq_len sweeps
- `compute_hidden_size_sweep_config()` for D1 hidden_size sweeps
- `run_speed_benchmark()` / `run_memory_benchmark()` shared helpers
- `--model` CLI argument
- Refactored `benchmark_geglu.py`, `benchmark_swiglu.py`, `benchmark_dyt.py`
- `BENCHMARK_GUIDELINES.md` contributor guide

### Phase 2: Model-config sweep (D2)

**Status: planned**

- [ ] Add `ModelConfigSweepConfig` dataclass
- [ ] Implement `compute_model_config_sweep_config()` with cross-model probe
- [ ] Add `--sweep-mode` and `--bt` CLI arguments to `parse_benchmark_script_args()`
- [ ] Add D2 code path to `benchmark_geglu.py` as reference implementation
- [ ] Validate on at least 2 devices (CUDA + NPU) to confirm OOM safety
- [ ] Update `BENCHMARK_GUIDELINES.md` with D2 instructions

### Phase 3: Rollout and visualization

**Status: planned**

- [ ] Port D2 support to all existing benchmark scripts
- [ ] Add bar chart / speedup visualization for D2 results
- [ ] Integrate into CI workflows
- [ ] Add more model profiles to `MODEL_REGISTRY` (e.g. Qwen-2.5-7B)

## 6. Directory Structure

```
benchmark/
├── data/
│   └── all_benchmark_data.csv
├── scripts/
│   ├── benchmark_model_configs.py      # ModelConfig, MODEL_REGISTRY, helpers
│   ├── utils.py                        # run_benchmarks, CSV, CLI
│   ├── benchmark_geglu.py              # D1 + D2
│   ├── benchmark_swiglu.py             # D1 + D2
│   ├── benchmark_dyt.py                # D1 + D2
│   └── ...
├── visualize/                          # (Phase 3) chart generation
│   └── ...
└── BENCHMARK_GUIDELINES.md             # contributor guide
```

Tcc0403:
The design plan looks good. Let's discuss more details on model sweep in phase 2

Tcc0403 (approved):
Thank you!

PR #1128, opened 2026-03-04
Add FSDP support to TiledMLP

alektebel:
## Summary
- Fix TiledMLP backward under FSDP (#893, #935): per-tile `backward()` reshards parameters mid-loop.
- Accumulate shard grads with `GradientAccumulator`, then run one `autograd.backward` after all tiles.
- Regression vs HuggingFace `LlamaMLP` at 2/4/8 GPUs, including parameter grads.

## Commits
- `feat(ops): add GradientAccumulator for tiled MLP shard grads` — Axolotl-style hook accumulator for per-shard param grads.
- `fix(tiled_mlp): FSDP-safe tiled backward via GradientAccumulator` — single batched backward so FSDP does not reshard between tiles.
- `test(tiled_mlp): FSDP regression vs HuggingFace LlamaMLP` — SwiGLU/GEGLU vs LlamaMLP under FSDP; skip 4/8 GPU if missing.

## Diff
3 files, **+200 / −9**.

## Test plan
- [x] `pytest test/transformers/test_tiled_mlp.py` (single GPU: FSDP cases skip)
- [x] 2× RTX 3060: 4 FSDP cases passed; 4- and 8-GPU skipped
- [x] `make checkstyle`

### DIFF
--- a/src/liger_kernel/ops/tiled_mlp.py
+++ b/src/liger_kernel/ops/tiled_mlp.py
@@ -78,19 +78,27 @@ class LigerTiledMLPFunction(torch.autograd.Function):
 
         x_shards = list(torch.chunk(x, chunks=shards, dim=0))
 
+        # Recompute every shard first, then run a single backward over all of them.
+        # Per-shard backward is incompatible with FSDP: the first shard's backward
+        # triggers resharding, so later shards recompute against flat local shards.
+        all_outputs = []
+        all_incoming_grads = []
+        shard_offset = 0
         for i, x_shard in enumerate(x_shards):
             x_shard.requires_grad_(x_requires_grad)
 
-            # if seqlen is not exactly divisible by shards the last step will be shorter than shard_step
             shard_step = x_shards[i].shape[0]
-            shard_offset = i * x_shards[0].shape[0]
-
-            x_shard.grad = x_grad.narrow(0, shard_offset, shard_step).view_as(x_shard)
             incoming_grad_shard = incoming_grad.narrow(0, shard_offset, shard_step).view_as(x_shard)
 
+            x_shard.grad = x_grad.narrow(0, shard_offset, shard_step).view_as(x_shard)
+
             with torch.enable_grad():
-                output = fn(mlp_module, x_shard)
-            torch.autograd.backward(output, incoming_grad_shard)
+                all_outputs.append(fn(mlp_module, x_shard))
+            all_incoming_grads.append(incoming_grad_shard)
+
+            shard_offset += shard_step
+
+        torch.autograd.backward(all_outputs, all_incoming_grads)
 
         # unflatten
         x_grad = x_grad.view(x_shape_orig)
--- a/src/liger_kernel/ops/tiled_mlp.py
+++ b/src/liger_kernel/ops/tiled_mlp.py
@@ -6,6 +6,7 @@ from typing import Optional
 
 import torch
 
+from liger_kernel.ops.utils import GradientAccumulator
 from liger_kernel.ops.utils import ensure_contiguous
 
 
@@ -44,6 +45,7 @@ class LigerTiledMLPFunction(torch.autograd.Function):
         ctx.fn = fn
         ctx.mlp_module = mlp_module
         ctx.shards = shards
+        ctx.compute_params = compute_params
         ctx.save_for_backward(x)
 
         # x.shape could be [bs, seqlen, hidden_size] or [seqlen, hidden_size] (moe experts)
@@ -61,6 +63,9 @@ class LigerTiledMLPFunction(torch.autograd.Function):
         (x,) = ctx.saved_tensors
         mlp_module = ctx.mlp_module
         shards = ctx.shards
+        compute_params = ctx.compute_params
+        if compute_params is None:
+            compute_params = list(mlp_module.parameters())
 
         x_requires_grad = x.requires_grad
         x = x.detach()
@@ -74,26 +79,42 @@ class LigerTiledMLPFunction(torch.autograd.Function):
         # flatten bs+seqlen to avoid having stride issues when narrowing into seqlen w/ bs>1
         x = x.view(-1, hidden_size)
         incoming_grad = grads[0].view(-1, hidden_size)
-        x_grad = torch.zeros_like(x)
+        x_grad = torch.zeros_like(x) if x_requires_grad else None
+
+        for param in compute_params:
+            if param.grad is not None:
+                param.grad = None
+
+        grad_accumulator = GradientAccumulator(compute_params)
+        grad_accumulator.install_hooks()
 
         x_shards = list(torch.chunk(x, chunks=shards, dim=0))
-
+        all_outputs = []
+        all_incoming_grads = []
+        shard_offset = 0
         for i, x_shard in enumerate(x_shards):
             x_shard.requires_grad_(x_requires_grad)
 
             # if seqlen is not exactly divisible by shards the last step will be shorter than shard_step
             shard_step = x_shards[i].shape[0]
-            shard_offset = i * x_shards[0].shape[0]
-
-            x_shard.grad = x_grad.narrow(0, shard_offset, shard_step).view_as(x_shard)
             incoming_grad_shard = incoming_grad.narrow(0, shard_offset, shard_step).view_as(x_shard)
 
-            with torch.enable_grad():
-                output = fn(mlp_module, x_shard)
-            torch.autograd.backward(output, incoming_grad_shard)
+            if x_grad is not None:
+                x_shard.grad = x_grad.narrow(0, shard_offset, shard_step).view_as(x_shard)
 
-        # unflatten
-        x_grad = x_grad.view(x_shape_orig)
+            with torch.enable_grad():
+                all_outputs.append(fn(mlp_module, x_shard))
+            all_incoming_grads.append(incoming_grad_shard)
+
+            shard_offset += shard_step
+
+        torch.autograd.backward(all_outputs, all_incoming_grads)
+
+        grad_accumulator.finalize_gradients()
+        grad_accumulator.cleanup()
+
+        if x_grad is not None:
+            x_grad = x_grad.view(x_shape_orig)
 
         return (None, None, x_grad, None, None)
 
--- a/test/transformers/test_tiled_mlp.py
+++ b/test/transformers/test_tiled_mlp.py
@@ -1,5 +1,9 @@
+import tempfile
+
 import pytest
 import torch
+import torch.multiprocessing as mp
+from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 
 from test.utils import supports_bfloat16
 from transformers.models.llama.configuration_llama import LlamaConfig
@@ -195,3 +199,147 @@ def test_tiled_swiglu_correctness(
         )
 
     torch.testing.assert_close(x1.grad, x2.grad, atol=atol, rtol=rtol, msg="Input gradients don't match")
+
+
+class _TorchSwiGLUMLP(torch.nn.Module):
+    def __init__(self, config):
+        super().__init__()
+        self.gate_proj = torch.nn.Linear(config.hidden_size, config.intermediate_size, bias=False)
+        self.up_proj = torch.nn.Linear(config.hidden_size, config.intermediate_size, bias=False)
+        self.down_proj = torch.nn.Linear(config.intermediate_size, config.hidden_size, bias=False)
+
+    def forward(self, x):
+        return self.down_proj(torch.nn.functional.silu(self.gate_proj(x)) * self.up_proj(x))
+
+
+class _TorchGEGLUMLP(torch.nn.Module):
+    def __init__(self, config):
+        super().__init__()
+        self.gate_proj = torch.nn.Linear(config.hidden_size, config.intermediate_size, bias=False)
+        self.up_proj = torch.nn.Linear(config.hidden_size, config.intermediate_size, bias=False)
+        self.down_proj = torch.nn.Linear(config.intermediate_size, config.hidden_size, bias=False)
+
+    def forward(self, x):
+        return self.down_proj(
+            torch.nn.functional.gelu(self.gate_proj(x), approximate="tanh") * self.up_proj(x)
+        )
+
+
+def _test_fsdp_tiled_vs_torch(
+    rank,
+    world_size,
+    bsz,
+    seq_len,
+    hidden_size,
+    intermediate_size,
+    num_shards,
+    activation,
+    dtype,
+    atol,
+    rtol,
+    file_name,
+):
+    torch.distributed.init_process_group(
+        backend="nccl",
+        init_method=f"file://{file_name}",
+        rank=rank,
+        world_size=world_size,
+    )
+    torch.cuda.set_device(rank)
+    device = f"cuda:{rank}"
+
+    if activation == "swiglu":
+        config = LlamaConfig(hidden_size=hidden_size, intermediate_size=intermediate_size, hidden_act="silu")
+        tiled_cls = LigerTiledSwiGLUMLP
+        torch_cls = _TorchSwiGLUMLP
+    else:
+        config = LlamaConfig(
+            hidden_size=hidden_size,
+            intermediate_size=intermediate_size,
+            hidden_act="gelu_pytorch_tanh",
+        )
+        tiled_cls = LigerTiledGEGLUMLP
+        torch_cls = _TorchGEGLUMLP
+
+    torch.manual_seed(42)
+    G = torch.randn(intermediate_size, hidden_size, device=device, dtype=dtype)
+    U = torch.randn(intermediate_size, hidden_size, device=device, dtype=dtype)
+    D = torch.randn(hidden_size, intermediate_size, device=device, dtype=dtype)
+    for t in (G, U, D):
+        torch.distributed.broadcast(t, src=0)
+
+    tiled_model = tiled_cls(config=config, num_shards=num_shards).to(device).to(dtype)
+    tiled_model.gate_proj.weight.data = G.clone()
+    tiled_model.up_proj.weight.data = U.clone()
+    tiled_model.down_proj.weight.data = D.clone()
+    tiled_model = FSDP(tiled_model, use_orig_params=True)
+
+    torch_model = torch_cls(config=config).to(device).to(dtype)
+    torch_model.gate_proj.weight.data = G.clone()
+    torch_model.up_proj.weight.data = U.clone()
+    torch_model.down_proj.weight.data = D.clone()
+    torch_model = FSDP(torch_model, use_orig_params=True)
+
+    torch.manual_seed(123)
+    x = torch.randn(bsz, seq_len, hidden_size, device=device, dtype=dtype) * 0.1
+    x_tiled = x.clone().requires_grad_(True)
+    x_ref = x.clone().requires_grad_(True)
+
+    out_tiled = tiled_model(x_tiled)
+    out_ref = torch_model(x_ref)
+    torch.testing.assert_close(
+        out_tiled, out_ref, atol=atol, rtol=rtol, msg=f"Rank {rank}: forward outputs don't match"
+    )
+
+    out_tiled.sum().backward()
+    out_ref.sum().backward()
+    torch.testing.assert_close(
+        x_tiled.grad, x_ref.grad, atol=atol, rtol=rtol, msg=f"Rank {rank}: input gradients don't match"
+    )
+
+    with FSDP.summon_full_params(tiled_model, with_grads=True), FSDP.summon_full_params(
+        torch_model, with_grads=True
+    ):
+        for i, (p_tiled, p_ref) in enumerate(zip(tiled_model.parameters(), torch_model.parameters())):
+            if p_tiled.grad is not None and p_ref.grad is not None:
+                torch.testing.assert_close(
+                    p_tiled.grad,
+                    p_ref.grad,
+                    atol=atol,
+                    rtol=rtol,
+                    msg=f"Rank {rank}: parameter {i} gradients don't match",
+                )
+
+    torch.distributed.destroy_process_group()
+
+
+@pytest.mark.skipif(torch.cuda.device_count() < 2, reason="requires at least 2 GPUs")
+@pytest.mark.parametrize("world_size", [ws for ws in [2, 4, 8] if ws <= torch.cuda.device_count()])
+@pytest.mark.parametrize("num_shards", [2, 4])
+@pytest.mark.parametrize("activation", ["swiglu", "geglu"])
+@pytest.mark.parametrize(
+    "bsz, seq_len, hidden_size, intermediate_size",
+    [(2, 512, 128, 256)],
+)
+def test_fsdp_tiled_mlp_matches_torch(world_size, num_shards, activation, bsz, seq_len, hidden_size, intermediate_size):
+    """TiledMLP + FSDP should match a torch MLP baseline on forward and all gradients."""
+    atol, rtol = 1e-3, 1e-3
+    with tempfile.NamedTemporaryFile() as f:
+        mp.spawn(
+            _test_fsdp_tiled_vs_torch,
+            args=(
+                world_size,
+                bsz,
+                seq_len,
+                hidden_size,
+                intermediate_size,
+                num_shards,
+                activation,
+                torch.float32,
+                atol,
+                rtol,
+                f.name,
+            ),
+            nprocs=world_size,
+            join=True,
+        )
--- a/test/transformers/test_tiled_mlp.py
+++ b/test/transformers/test_tiled_mlp.py
@@ -1,16 +1,36 @@
+import tempfile
+
 import pytest
 import torch
+import torch.multiprocessing as mp
 
 from test.utils import supports_bfloat16
+from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 from transformers.models.llama.configuration_llama import LlamaConfig
+from transformers.models.llama.modeling_llama import LlamaMLP
 
 from liger_kernel.transformers.geglu import LigerGEGLUMLP
 from liger_kernel.transformers.swiglu import LigerSwiGLUMLP
 from liger_kernel.transformers.tiled_mlp import LigerTiledGEGLUMLP
 from liger_kernel.transformers.tiled_mlp import LigerTiledSwiGLUMLP
+from liger_kernel.utils import infer_comm_backend
 from liger_kernel.utils import infer_device
 
 device = infer_device()
+
+_NDEV = torch.cuda.device_count() if torch.cuda.is_available() else 0
+
+
+def _need_gpus(n):
+    return pytest.mark.skipif(_NDEV < n, reason=f"requires {n} GPUs")
+
+
+# HuggingFace LlamaMLP is the torch reference used by test_swiglu / test_geglu.
+# silu → SwiGLU, gelu_pytorch_tanh → GEGLU.
+_FSDP_MLP_KINDS = {
+    "swiglu": {"hidden_act": "silu", "tiled_cls": LigerTiledSwiGLUMLP},
+    "geglu": {"hidden_act": "gelu_pytorch_tanh", "tiled_cls": LigerTiledGEGLUMLP},
+}
 
 
 @pytest.mark.parametrize(
@@ -195,3 +215,99 @@ def test_tiled_swiglu_correctness(
         )
 
     torch.testing.assert_close(x1.grad, x2.grad, atol=atol, rtol=rtol, msg="Input gradients don't match")
+
+
+def _test_fsdp_tiled_mlp(
+    rank, world_size, mlp_kind, num_shards, bsz, seq_len, hidden_size, intermediate_size, file_name
+):
+    """FSDP-wrapped TiledMLP vs FSDP-wrapped HuggingFace LlamaMLP (torch).
+
+    Issue #893: tiled backward must not reshard FSDP params between sequence
+    tiles. Matching *parameter* grads (via summon_full_params) is the actual
+    regression; input grads can still agree when shard grads are wrong.
+    """
+    torch.distributed.init_process_group(
+        backend=infer_comm_backend(),
+        init_method=f"file://{file_name}",
+        rank=rank,
+        world_size=world_size,
+    )
+    torch.cuda.set_device(rank)
+    device = f"cuda:{rank}"
+
+    spec = _FSDP_MLP_KINDS[mlp_kind]
+    config = LlamaConfig(hidden_size=hidden_size, intermediate_size=intermediate_size, hidden_act=spec["hidden_act"])
+    dtype = torch.float32
+    atol, rtol = 1e-3, 1e-3
+
+    torch.manual_seed(42)
+    G = torch.randn(intermediate_size, hidden_size, device=device, dtype=dtype)
+    U = torch.randn(intermediate_size, hidden_size, device=device, dtype=dtype)
+    D = torch.randn(hidden_size, intermediate_size, device=device, dtype=dtype)
+    for t in (G, U, D):
+        torch.distributed.broadcast(t, src=0)
+
+    tiled_model = spec["tiled_cls"](config=config, num_shards=num_shards).to(device).to(dtype)
+    ref_model = LlamaMLP(config=config).to(device).to(dtype)
+    for model in (tiled_model, ref_model):
+        model.gate_proj.weight.data = G.clone()
+        model.up_proj.weight.data = U.clone()
+        model.down_proj.weight.data = D.clone()
+    tiled_model = FSDP(tiled_model, use_orig_params=True, device_id=rank)
+    ref_model = FSDP(ref_model, use_orig_params=True, device_id=rank)
+
+    torch.manual_seed(123)
+    x = torch.randn(bsz, seq_len, hidden_size, device=device, dtype=dtype) * 0.1
+    torch.distributed.broadcast(x, src=0)
+    x_tiled = x.clone().requires_grad_(True)
+    x_ref = x.clone().requires_grad_(True)
+
+    out_tiled = tiled_model(x_tiled)
+    out_ref = ref_model(x_ref)
+    torch.testing.assert_close(
+        out_tiled, out_ref, atol=atol, rtol=rtol, msg=f"Rank {rank}: forward outputs don't match"
+    )
+
+    out_tiled.sum().backward()
+    out_ref.sum().backward()
+    torch.testing.assert_close(
+        x_tiled.grad, x_ref.grad, atol=atol, rtol=rtol, msg=f"Rank {rank}: input gradients don't match"
+    )
+
+    with FSDP.summon_full_params(tiled_model, with_grads=True), FSDP.summon_full_params(ref_model, with_grads=True):
+        for i, (p_tiled, p_ref) in enumerate(zip(tiled_model.parameters(), ref_model.parameters())):
+            assert p_tiled.grad is not None, f"Rank {rank}: tiled param {i} has no grad"
+            assert p_ref.grad is not None, f"Rank {rank}: ref param {i} has no grad"
+            torch.testing.assert_close(
+                p_tiled.grad,
+                p_ref.grad,
+                atol=atol,
+                rtol=rtol,
+                msg=f"Rank {rank}: parameter {i} gradients don't match",
+            )
+
+    torch.distributed.destroy_process_group()
+
+
+@pytest.mark.skipif(_NDEV < 2, reason="requires at least 2 GPUs")
+@pytest.mark.parametrize("mlp_kind", ["swiglu", "geglu"])
+@pytest.mark.parametrize(
+    "world_size, num_shards, bsz, seq_len, hidden_size, intermediate_size",
+    [
+        pytest.param(2, 2, 2, 512, 128, 256, id="2gpu-2shards-even"),
+        pytest.param(2, 4, 2, 127, 128, 256, id="2gpu-4shards-ragged"),
+        pytest.param(4, 2, 2, 512, 128, 256, id="4gpu-2shards-even", marks=_need_gpus(4)),
+        pytest.param(4, 4, 2, 255, 256, 512, id="4gpu-4shards-ragged", marks=_need_gpus(4)),
+        pytest.param(8, 2, 2, 512, 128, 256, id="8gpu-2shards-even", marks=_need_gpus(8)),
+        pytest.param(8, 4, 2, 511, 256, 512, id="8gpu-4shards-ragged", marks=_need_gpus(8)),
+    ],
+)
+def test_fsdp_tiled_mlp(mlp_kind, world_size, num_shards, bsz, seq_len, hidden_size, intermediate_size):
+    """TiledMLP + FSDP must match HuggingFace LlamaMLP (torch) on forward, input grads, and param grads."""
+    with tempfile.NamedTemporaryFile() as f:
+        mp.spawn(
+            _test_fsdp_tiled_mlp,
+            args=(world_size, mlp_kind, num_shards, bsz, seq_len, hidden_size, intermediate_size, f.name),
+            nprocs=world_size,
+            join=True,
+        )
--- a/src/liger_kernel/ops/utils.py
+++ b/src/liger_kernel/ops/utils.py
@@ -13,9 +13,12 @@ Modifications made by Yanning Chen, 2024.
 import functools
 import importlib
 import operator
+import threading
 
 from contextlib import contextmanager
 from typing import Callable
+from typing import List
+from typing import Optional
 
 import torch
 import triton
@@ -151,6 +154,57 @@ def get_npu_core_count(default: int = 20) -> int:
         return int(props.get("num_vectorcore", default))
     except Exception:
         return default
+
+
+class GradientAccumulator:
+    """
+    Hook-based gradient accumulator for TiledMLP (Axolotl pattern).
+
+    Reference:
+    https://github.com/axolotl-ai-cloud/axolotl/blob/main/src/axolotl/monkeypatch/tiled_mlp/base.py
+    """
+
+    def __init__(
+        self,
+        params: List[torch.nn.Parameter],
+        dtype: Optional[torch.dtype] = None,
+    ):
+        self.params = [p for p in params if p.requires_grad]
+        self.grad_accumulation_dtype = dtype or torch.float32
+        self.accumulated_grads = {}
+        self.hooks = []
+        self.lock = threading.Lock()
+
+        for param in self.params:
+            if param.grad is not None:
+                self.accumulated_grads[param] = param.grad.to(self.grad_accumulation_dtype)
+                param.grad = None
+            else:
+                self.accumulated_grads[param] = torch.zeros_like(
+                    param, dtype=self.grad_accumulation_dtype
+                )
+
+    def install_hooks(self):
+        def create_hook(param):
+            def hook(grad):
+                with self.lock:
+                    self.accumulated_grads[param] += grad.to(self.grad_accumulation_dtype)
+                    return None
+
+            return hook
+
+        for param in self.params:
+            self.hooks.append(param.register_hook(create_hook(param)))
+
+    def finalize_gradients(self):
+        for param in self.params:
+            param.grad = self.accumulated_grads[param].to(param.dtype)
+
+    def cleanup(self):
+        for hook in self.hooks:
+            hook.remove()
+        self.hooks.clear()
+        del self.accumulated_grads
 
 
 def set_large_grf_mode(kernel_args: dict):


Tcc0403:
I'm not entirely if `summon_full_params` context is workable.

According to https://docs.pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.FullyShardedDataParallel.summon_full_params

> This can not be used within a forward or backward pass. Nor can forward and backward be started from within this context.

Is mutli-gpu environment accessible to you? If not, could you provide mutli-gpu test script that I can run directly?

alektebel:
Hi @Tcc0403 ,

You're right, after checking the docs, `summon_full_params` is indeed not usable within a forward/backward pass, so I'm pivoting to a `no_sync`-based context instead, as discussed in the #935  thread.

I don't have multi-GPU access locally, I'm renting 2× RTX 3060s (~$0.09/h) to properly validate this. Will update the PR with a multi-GPU test script and results as soon as I get positive outcomes.

Tcc0403:
https://docs.pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.FullyShardedDataParallel.no_sync

> This should only be used on the root FSDP instance and will recursively apply to all children FSDP instances.

doc says only apply on the root FSDP instance. It's really tricky to make it work with fsdp.

alektebel:
Hi @Tcc0403 ,

You were right about the `no_sync` context issue. Being `TiledMLP` a wrapped object, it's impossible for it to know when it is being wrapped by `FSDP` or other object, and `FSDP` in this case is an external dependency, so this aproach doesn't seem promising.

I propose two different potential solutions:

- **FSDP2 fix:** One comes about changing the FSDP used for the newer `fully_shard` (`FSDP`) in pytorch. Below I've tested for the bug in the 2x3060RTX env and ran without an error
- **FSDP1 fix:** The second one is a fix in the `backward` method of TiledMLP.  Instead of calling `.autograd()` each shard, we call `.grad()`, which doesn't trigger a full backward method, therefore allowing us to accumulate the gradients correctly. Afterwards, `.autograd()` is called to propagate these grads to the rest of the computational graph.

## The FSDP2 fix

After analyzing more deeply the problem, I found it tricky to stop the gradient flow to take place in FSDP, being this an external application, when doing the n forward-backward steps in tiledMLP. Researching, I found that actually, there is a newer version of FSDP that can handle this issue in pytorch. Here is the code I ran in a 2x3060 cloud instance that actually ran succesfully:
```bash
(main) root@C.32454339:/workspace$ cat > test_fsdp2.py << 'EOF'
import os
from types import SimpleNamespace

import torch
import torch.distributed as dist
from torch.distributed.fsdp import fully_shard

from liger_kernel.transformers.tiled_mlp import LigerTiledSwiGLUMLP


def run():
    dist.init_process_group("nccl")
    rank = dist.get_rank()
    torch.cuda.set_device(rank)

    config = SimpleNamespace(hidden_size=32, intermediate_size=64)

    model = LigerTiledSwiGLUMLP(config, num_shards=4).cuda()
    fully_shard(model)

    x = torch.randn(4, 16, 32, device=f"cuda:{rank}", requires_grad=True)
    model(x).sum().backward()

    if rank == 0:
        print("OK — no error")

    dist.destroy_process_group()


if __name__ == "__main__":
    run()
EOF
torchrun --nproc_per_node=2 test_fsdp2.py
W0306 16:19:01.108000 1700 site-packages/torch/distributed/run.py:852] 
W0306 16:19:01.108000 1700 site-packages/torch/distributed/run.py:852] *****************************************
W0306 16:19:01.108000 1700 site-packages/torch/distributed/run.py:852] Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed. 
W0306 16:19:01.108000 1700 site-packages/torch/distributed/run.py:852] *****************************************
OK — no error
```
FSDP2's per-parameter DTensor sharding handles multiple backward passes through the same parameters correctly by design, unlike FSDP1's flat parameter state machine.

Source:
https://github.com/pytorch/pytorch/issues/114299


Whereas if we use the current HEAD FSDP, with the same functionlity as above, I get an error:


```bash

(main) root@C.32454339:/workspace$ cat > test_fsdp1.py << 'EOF'
import os
from types import SimpleNamespace

import torch
import torch.distributed as dist
from torch.distributed.fsdp import FullyShardedDataParallel as FSDP

from liger_kernel.transformers.tiled_mlp import LigerTiledSwiGLUMLP


def run():
    dist.init_process_group("nccl")
    rank = dist.get_rank()
    torch.cuda.set_device(rank)

    config = SimpleNamespace(hidden_size=32, intermediate_size=64)

    model = FSDP(
        LigerTiledSwiGLUMLP(config, num_shards=4).cuda(),
        device_id=torch.device(f"cuda:{rank}"),
    )

    x = torch.randn(4, 16, 32, device=f"cuda:{rank}", requires_grad=True)
    model(x).sum().backward()

    if rank == 0:
        print("OK — no error")

    dist.destroy_process_group()


if __name__ == "__main__":
    run()
EOF
torchrun --nproc_per_node=2 test_fsdp1.py
W0306 16:23:22.581000 1819 site-packages/torch/distributed/run.py:852] 
W0306 16:23:22.581000 1819 site-packages/torch/distributed/run.py:852] *****************************************
W0306 16:23:22.581000 1819 site-packages/torch/distributed/run.py:852] Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed. 
W0306 16:23:22.581000 1819 site-packages/torch/distributed/run.py:852] *****************************************
[rank1]: Traceback (most recent call last):
[rank1]:   File "/workspace/test_fsdp1.py", line 33, in <module>
[rank1]:     run()
[rank1]:   File "/workspace/test_fsdp1.py", line 24, in run
[rank1]:     model(x).sum().backward()
[rank1]:   File "/venv/main/lib/python3.12/site-packages/torch/_tensor.py", line 630, in backward
[rank1]:     torch.autograd.backward(
[rank1]:   File "/venv/main/lib/python3.12/site-packages/torch/autograd/__init__.py", line 364, in backward
[rank1]:     _engine_run_backward(
[rank1]:   File "/venv/main/lib/python3.12/site-packages/torch/autograd/graph.py", line 865, in _engine_run_backward
[rank1]:     return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
[rank1]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/venv/main/lib/python3.12/site-packages/torch/autograd/function.py", line 317, in apply
[rank1]:     return user_fn(self, *args)
[rank1]:            ^^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/workspace/Liger-Kernel/src/liger_kernel/ops/utils.py", line 40, in wrapper
[rank1]:     return fn(ctx, *args, **kwargs)
[rank1]:            ^^^^^^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/workspace/Liger-Kernel/src/liger_kernel/ops/tiled_mlp.py", line 92, in backward
[rank1]:     output = fn(mlp_module, x_shard)
[rank1]:              ^^^^^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/workspace/Liger-Kernel/src/liger_kernel/transformers/tiled_mlp.py", line 102, in _mlp_forward
[rank1]:     gate = module.gate_proj(x)
[rank1]:            ^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/venv/main/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1776, in _wrapped_call_impl
[rank1]:     return self._call_impl(*args, **kwargs)
[rank1]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/venv/main/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1787, in _call_impl
[rank1]:     return forward_call(*args, **kwargs)
[rank1]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/venv/main/lib/python3.12/site-packages/torch/nn/modules/linear.py", line 134, in forward
[rank1]:     return F.linear(input, self.weight, self.bias)
[rank1]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank1]: RuntimeError: setStorage: sizes [32, 64], strides [1, 32], storage offset 0, and itemsize 4 requiring a storage size of 8192 are out of bounds for storage of size 0
[rank0]: Traceback (most recent call last):
[rank0]:   File "/workspace/test_fsdp1.py", line 33, in <module>
[rank0]:     run()
[rank0]:   File "/workspace/test_fsdp1.py", line 24, in run
[rank0]:     model(x).sum().backward()
[rank0]:   File "/venv/main/lib/python3.12/site-packages/torch/_tensor.py", line 630, in backward
[rank0]:     torch.autograd.backward(
[rank0]:   File "/venv/main/lib/python3.12/site-packages/torch/autograd/__init__.py", line 364, in backward
[rank0]:     _engine_run_backward(
[rank0]:   File "/venv/main/lib/python3.12/site-packages/torch/autograd/graph.py", line 865, in _engine_run_backward
[rank0]:     return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/venv/main/lib/python3.12/site-packages/torch/autograd/function.py", line 317, in apply
[rank0]:     return user_fn(self, *args)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/workspace/Liger-Kernel/src/liger_kernel/ops/utils.py", line 40, in wrapper
[rank0]:     return fn(ctx, *args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/workspace/Liger-Kernel/src/liger_kernel/ops/tiled_mlp.py", line 92, in backward
[rank0]:     output = fn(mlp_module, x_shard)
[rank0]:              ^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/workspace/Liger-Kernel/src/liger_kernel/transformers/tiled_mlp.py", line 102, in _mlp_forward
[rank0]:     gate = module.gate_proj(x)
[rank0]:            ^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/venv/main/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1776, in _wrapped_call_impl
[rank0]:     return self._call_impl(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/venv/main/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1787, in _call_impl
[rank0]:     return forward_call(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/venv/main/lib/python3.12/site-packages/torch/nn/modules/linear.py", line 134, in forward
[rank0]:     return F.linear(input, self.weight, self.bias)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: RuntimeError: setStorage: sizes [32, 64], strides [1, 32], storage offset 0, and itemsize 4 requiring a storage size of 8192 are out of bounds for storage of size 0
``` 

## The FSDP1 fix

 I want to propose a solution related to the current `FSDP`, in case we don't want to change it for any reason. My proposal is that, being the main problem related to this issue, is that a full backward is triggered when calling the `backward`method at `TiledMLP` many times (n_shards), which interferes with `FSDP`'s state machine, corrupting the resharding state of the FSDP. 

**proposed solution**
Instead of calling `torch.autograd.backward()` inside the tiling loop, we use `torch.autograd.grad()`, which returns gradients as plain tensors without writing to `.grad` or firing any registered hooks.  `FSDP`'s post-backward reshard hook never fires during the loop. Param gradients are accumulated manually across shards and written to `.grad` exactly once after the loop, which is the single gradient event `FSDP` expects.

Please find updated these changes on the PR branch.

I've tested this fix on the 2x3060 RTX environment and found no problems. Also, I've tried benchmarking it against the old version of `TiledMLP` and found no significant change in memory or speed.

With that, here would be the passing test I mentioned, for this fix.

Importantly, this fix requires no knowledge of `FSDP` inside `LigerTiledSwiGLUMLP`

```bash
(main) root@C.32454339:/workspace$ cat > test_fsdp1.py << 'EOF'
import os
from types import SimpleNamespace

import torch
import torch.distributed as dist
from torch.distributed.fsdp import FullyShardedDataParallel as FSDP

from liger_kernel.transformers.tiled_mlp import LigerTiledSwiGLUMLP


def run():
    dist.init_process_group("nccl")
    rank = dist.get_rank()
    torch.cuda.set_device(rank)

    config = SimpleNamespace(hidden_size=32, intermediate_size=64)

    model = FSDP(
        LigerTiledSwiGLUMLP(config, num_shards=4).cuda(),
        device_id=torch.device(f"cuda:{rank}"),
    )

    x = torch.randn(4, 16, 32, device=f"cuda:{rank}", requires_grad=True)
    model(x).sum().backward()

    if rank == 0:
        print("OK — no error")

    dist.destroy_process_group()


if __name__ == "__main__":
    run()
EOF
torchrun --nproc_per_node=2 test_fsdp1.py
W0306 18:13:26.722000 1029 site-packages/torch/distributed/run.py:852] 
W0306 18:13:26.722000 1029 site-packages/torch/distributed/run.py:852] *****************************************
W0306 18:13:26.722000 1029 site-packages/torch/distributed/run.py:852] Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed. 
W0306 18:13:26.722000 1029 site-packages/torch/distributed/run.py:852] *****************************************
OK — no error
```

Let me know what you think.

Tcc0403:
> Instead of calling torch.autograd.backward() inside the tiling loop, we use torch.autograd.grad(), which returns gradients as plain tensors without writing to .grad or firing any registered hooks. FSDP's post-backward reshard hook never fires during the loop. Param gradients are accumulated manually across shards and written to .grad exactly once after the loop, which is the single gradient event FSDP expects.

Sounds good. Let's try this out.

alektebel:
Hi @Tcc0403 ,

I believe the changes are updated on the PR branch. This is the code I ran too:

1. Check the TiledMLP run succesfully wrapped on an FSDP
2. FSDP(TiledMLP) behaves exactly in a functional way as a normal TiledMLP


## Succesfull code run on the PR brach

```bash
 root@C.32658063:/workspace/Liger-Kernel$ cat > test_fsdp1.py << 'EOF'
import os
from types import SimpleNamespace

import torch
import torch.distributed as dist
from torch.distributed.fsdp import FullyShardedDataParallel as FSDP

from liger_kernel.ops.tiled_mlp import apply_tiled_mlp
from liger_kernel.transformers.tiled_mlp import LigerTiledSwiGLUMLP

ATOL = 1e-4
RTOL = 1e-4


def run():
    dist.init_process_group("nccl")
    rank = dist.get_rank()
    world_size = dist.get_world_size()
    torch.cuda.set_device(rank)

    config = SimpleNamespace(hidden_size=32, intermediate_size=64)

    # Shared input — identical on all ranks.
    # requires_grad=True is required: LigerTiledMLPFunction.forward() runs under
    # torch.no_grad(), so the output only has a grad_fn (and is differentiable)
    # when the input itself requires grad.
    torch.manual_seed(0)
    x_full = torch.randn(4, 16, 32, device=f"cuda:{rank}", requires_grad=True)

    # --- Reference: plain model on rank 0 only ---
    if rank == 0:
        torch.manual_seed(42)
        ref_model = LigerTiledSwiGLUMLP(config, num_shards=4).cuda()
        # detach+clone gives an independent leaf so ref backward graph is isolated
        x_ref = x_full.detach().clone().requires_grad_(True)
        ref_out = ref_model(x_ref)
        ref_out.sum().backward()
        ref_grads = {n: p.grad.clone() for n, p in ref_model.named_parameters()}
        ref_out = ref_out.detach()

    # --- FSDP model (same seed so weights are identical at init) ---
    torch.manual_seed(42)
    fsdp_model = FSDP(
        LigerTiledSwiGLUMLP(config, num_shards=4).cuda(),
        device_id=torch.device(f"cuda:{rank}"),
        use_orig_params=True,
    )

    # Each rank processes its own batch shard
    batch_per_rank = x_full.shape[0] // world_size
    x_local = x_full[rank * batch_per_rank:(rank + 1) * batch_per_rank].detach().clone().requires_grad_(True)

    fsdp_out = fsdp_model(x_local)
torchrun --nproc_per_node=2 test_fsdp1.py{name}': max_diff={max_gdiff:.6f}"erence.
W0310 23:10:49.691000 2370 site-packages/torch/distributed/run.py:852] 
W0310 23:10:49.691000 2370 site-packages/torch/distributed/run.py:852] *****************************************
W0310 23:10:49.691000 2370 site-packages/torch/distributed/run.py:852] Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed. 
W0310 23:10:49.691000 2370 site-packages/torch/distributed/run.py:852] *****************************************
OK — forward match  (max_diff=0.00e+00)
OK — gradient match
```


## Bug reproducing code run on the repo HEAD 
Please find below the code to reproduce the bug at HEAD:

```bash
(main) root@C.32658063:/workspace/Liger-Kernel$ cat > test_fsdp1.py << 'EOF'
import os
from types import SimpleNamespace

import torch
import torch.distributed as dist
from torch.distributed.fsdp import FullyShardedDataParallel as FSDP

from liger_kernel.ops.tiled_mlp import apply_tiled_mlp
from liger_kernel.transformers.tiled_mlp import LigerTiledSwiGLUMLP

ATOL = 1e-4
RTOL = 1e-4


def run():
    dist.init_process_group("nccl")
    rank = dist.get_rank()
    world_size = dist.get_world_size()
    torch.cuda.set_device(rank)

    config = SimpleNamespace(hidden_size=32, intermediate_size=64)

    # Shared input — identical on all ranks.
    # requires_grad=True is required: LigerTiledMLPFunction.forward() runs under
    # torch.no_grad(), so the output only has a grad_fn (and is differentiable)
    # when the input itself requires grad.
    torch.manual_seed(0)
    x_full = torch.randn(4, 16, 32, device=f"cuda:{rank}", requires_grad=True)

    # --- Reference: plain model on rank 0 only ---
    if rank == 0:
        torch.manual_seed(42)
        ref_model = LigerTiledSwiGLUMLP(config, num_shards=4).cuda()
        # detach+clone gives an independent leaf so ref backward graph is isolated
        x_ref = x_full.detach().clone().requires_grad_(True)
        ref_out = ref_model(x_ref)
        ref_out.sum().backward()
        ref_grads = {n: p.grad.clone() for n, p in ref_model.named_parameters()}
        ref_out = ref_out.detach()

    # --- FSDP model (same seed so weights are identical at init) ---
    torch.manual_seed(42)
    fsdp_model = FSDP(
        LigerTiledSwiGLUMLP(config, num_shards=4).cuda(),
        device_id=torch.device(f"cuda:{rank}"),
        use_orig_params=True,
    )

    # Each rank processes its own batch shard
    batch_per_rank = x_full.shape[0] // world_size
    x_local = x_full[rank * batch_per_rank:(rank + 1) * batch_per_rank].detach().clone().requires_grad_(True)

    fsdp_out = fsdp_model(x_local)
torchrun --nproc_per_node=2 test_fsdp1.py{name}': max_diff={max_gdiff:.6f}"erence.
W0310 23:13:58.381000 2921 site-packages/torch/distributed/run.py:852] 
W0310 23:13:58.381000 2921 site-packages/torch/distributed/run.py:852] *****************************************
W0310 23:13:58.381000 2921 site-packages/torch/distributed/run.py:852] Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed. 
W0310 23:13:58.381000 2921 site-packages/torch/distributed/run.py:852] *****************************************
[rank0]: Traceback (most recent call last):
[rank0]:   File "/workspace/Liger-Kernel/test_fsdp1.py", line 88, in <module>
[rank0]:     run()
[rank0]:   File "/workspace/Liger-Kernel/test_fsdp1.py", line 54, in run
[rank0]:     fsdp_out.sum().backward()
[rank0]:   File "/venv/main/lib/python3.12/site-packages/torch/_tensor.py", line 630, in backward
[rank0]:     torch.autograd.backward(
[rank0]:   File "/venv/main/lib/python3.12/site-packages/torch/autograd/__init__.py", line 364, in backward
[rank0]:     _engine_run_backward(
[rank0]:   File "/venv/main/lib/python3.12/site-packages/torch/autograd/graph.py", line 865, in _engine_run_backward
[rank0]:     return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/venv/main/lib/python3.12/site-packages/torch/autograd/function.py", line 317, in apply
[rank0]:     return user_fn(self, *args)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/workspace/Liger-Kernel/src/liger_kernel/ops/utils.py", line 40, in wrapper
[rank0]:     return fn(ctx, *args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/workspace/Liger-Kernel/src/liger_kernel/ops/tiled_mlp.py", line 92, in backward
[rank0]:     output = fn(mlp_module, x_shard)
[rank0]:              ^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/workspace/Liger-Kernel/src/liger_kernel/transformers/tiled_mlp.py", line 102, in _mlp_forward
[rank0]:     gate = module.gate_proj(x)
[rank0]:            ^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/venv/main/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1776, in _wrapped_call_impl
[rank0]:     return self._call_impl(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/venv/main/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1787, in _call_impl
[rank0]:     return forward_call(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/venv/main/lib/python3.12/site-packages/torch/nn/modules/linear.py", line 134, in forward
[rank0]:     return F.linear(input, self.weight, self.bias)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: RuntimeError: size mismatch, got input (8), mat (8x32), vec (2048)
W0310 23:14:03.576000 2921 site-packages/torch/distributed/elastic/multiprocessing/api.py:1010] Sending process 2985 closing signal SIGTERM
E0310 23:14:03.741000 2921 site-packages/torch/distributed/elastic/multiprocessing/api.py:984] failed (exitcode: 1) local_rank: 0 (pid: 2984) of binary: /venv/main/bin/python3

```


What more tests come to your mind in order to make sure that this doesn't break anything and behaves as expected? Should I integrate this from above into the test suite? Only thing, it will only work on a 2+ gpu instance.

Tcc0403:
Add a test similar to https://github.com/linkedin/Liger-Kernel/blob/8c4be2b3b0f508f4e9e6c84491cc142516529210/test/transformers/test_rms_norm.py#L304

alektebel:
Hi @Tcc0403 ,

Added a test for FSDP-TiledMLP compatibility. Ran it on 2x3060 env without issues:

```bash

test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-2-256-512-1-2] PASSED                                                            [ 81%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-2-256-512-2-2] PASSED                                                            [ 82%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-2-256-512-4-2] PASSED                                                            [ 83%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-2-512-1024-1-2] PASSED                                                           [ 84%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-2-512-1024-2-2] PASSED                                                           [ 85%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-2-512-1024-4-2] PASSED                                                           [ 86%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-1-128-256-1-2] PASSED                                                            [ 87%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-1-128-256-2-2] PASSED                                                            [ 88%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-1-128-256-4-2] PASSED                                                            [ 90%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-2-256-512-1-2] PASSED                                                                [ 91%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-2-256-512-2-2] PASSED                                                                [ 92%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-2-256-512-4-2] PASSED                                                                [ 93%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-2-512-1024-1-2] PASSED                                                               [ 94%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-2-512-1024-2-2] PASSED                                                               [ 95%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-2-512-1024-4-2] PASSED                                                               [ 96%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-1-128-256-1-2] PASSED                                                                [ 97%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-1-128-256-2-2] PASSED                                                                [ 98%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-1-128-256-4-2] PASSED                                                                [100%]
```

One question: **¿Is adding 2, 4, 8 world size overkill for the tests?**

I've left the combinations to test as just 2 world size:

```python
@pytest.mark.parametrize("world_size", [2])  # extend to [2, 4, 8] on multi-GPU hosts
```

The test was added to `test/transformers/test_tiled_mlp.py`, changes are now visible in the PR branch.

Let me know if it would be better to have 2, 4, 8 gpu instance test, skipped in case not available by the setup.

test/transformers/test_tiled_mlp.py
```diff
@@ -195,3 +198,90 @@ def test_tiled_swiglu_correctness(
...
+
+    out = model(x)
+    out.sum().backward()
+
+    ref_out = ref_model(x)
+    ref_out.sum().backward()
+
+    # Assert
+    torch.testing.assert_close(out, ref_out, atol=atol, rtol=rtol)
+
```
Tcc0403: missing gradients comparison

test/transformers/test_tiled_mlp.py
```diff
@@ -195,3 +198,90 @@ def test_tiled_swiglu_correctness(
...
+    model = LigerTiledSwiGLUMLP(config=config, num_shards=num_shards).to(device).to(dtype)
+    model.gate_proj.weight.data = G.clone()
+    model.up_proj.weight.data = U.clone()
+    model.down_proj.weight.data = D.clone()
+
+    # Wrap with FSDP
+    model = FSDP(model, use_orig_params=True)
+
+    # Reference: same weights, no FSDP
+    ref_model = LigerTiledSwiGLUMLP(config=config, num_shards=num_shards).to(device).to(dtype)
```
Tcc0403: test against torch implentation as reference

test/transformers/test_tiled_mlp.py
```diff
@@ -195,3 +198,90 @@ def test_tiled_swiglu_correctness(
...
+    ref_out = ref_model(x)
+    ref_out.sum().backward()
+
+    # Assert
+    torch.testing.assert_close(out, ref_out, atol=atol, rtol=rtol)
+
+
+@pytest.mark.skipif(torch.cuda.device_count() < 2, reason="requires at least 2 GPUs")
+
+@pytest.mark.parametrize("world_size", [2])  # extend to [2, 4, 8] on multi-GPU hosts
```
Tcc0403: Extend to [2, 4, 8] is fine. Just make sure they are skipped if gpus are not present.

alektebel:
Hi @Tcc0403 ,

I've added tests covering:

- Gradients comparison
- torch implementation sanity check
- Testing on 2, 4 or 8 gpus if available.

I had to relax the gradient tolerance to 1e-3 (from 1e-5) to account for expected numerical differences in bfloat16 + FSDP sharding/fusion. This is in line with PyTorch's own tolerances for bfloat16 (rtol ~1e-2) and common practice in transformer kernels.

I've benchmarked the modified TiledMLP against the current HEAD implementation. While there's a small memory overhead in some cases (likely due to tiling/fusion trade-offs), we retain clear memory savings overall, and these benefits scale strongly with increasing sequence length / token count (see plot below)."

Backward optimizations:
The last commit also includes several optimizations for TiledMLP’s backward pass. These improvements aim to prevent premature resharding during the tiled backward recompute loop, which helps avoid unnecessary communication and can further improve scaling efficiency and memory usage in large distributed setups.

In the below image `HEAD`refers to the implementation of TiledMLP at the HEAD of this repo, vs `Custom` which is the implementation I'm currently developing in this PR. As you can see, we still see memory benefits, and the benefits scale well when increasing token size.
image


I've commited the changes and they are available in the PR branch. 


These are the passing tests:

```bash
(main) root@C.33135850:/workspace/Liger-Kernel$ python3 -m pytest test/transformers/test_tiled_mlp.py -v
========================================================================= test session starts =========================================================================
platform linux -- Python 3.12.12, pytest-9.0.2, pluggy-1.6.0 -- /venv/main/bin/python3
cachedir: .pytest_cache
rootdir: /workspace/Liger-Kernel
configfile: pyproject.toml
plugins: cov-7.0.0, anyio-4.12.0
collected 144 items                                                                                                                                                   

test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-None-dtype0-1.0-2e-06-1-1024-128-256] PASSED                                             [  0%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-None-dtype0-1.0-2e-06-2-1024-64-256] PASSED                                              [  1%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-None-dtype0-1.0-2e-06-4-127-128-256] PASSED                                              [  2%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-None-dtype1-1.0-1.0-1-1024-128-256] PASSED                                               [  2%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-None-dtype1-1.0-1.0-2-1024-64-256] SKIPPED (Skipping unstable BF16 configuration: hi...) [  3%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-None-dtype1-1.0-1.0-4-127-128-256] PASSED                                                [  4%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-2-dtype0-1.0-2e-06-1-1024-128-256] PASSED                                                [  4%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-2-dtype0-1.0-2e-06-2-1024-64-256] PASSED                                                 [  5%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-2-dtype0-1.0-2e-06-4-127-128-256] PASSED                                                 [  6%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-2-dtype1-1.0-1.0-1-1024-128-256] PASSED                                                  [  6%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-2-dtype1-1.0-1.0-2-1024-64-256] SKIPPED (Skipping unstable BF16 configuration: hidde...) [  7%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-2-dtype1-1.0-1.0-4-127-128-256] PASSED                                                   [  8%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-4-dtype0-1.0-2e-06-1-1024-128-256] PASSED                                                [  9%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-4-dtype0-1.0-2e-06-2-1024-64-256] PASSED                                                 [  9%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-4-dtype0-1.0-2e-06-4-127-128-256] PASSED                                                 [ 10%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-4-dtype1-1.0-1.0-1-1024-128-256] PASSED                                                  [ 11%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-4-dtype1-1.0-1.0-2-1024-64-256] SKIPPED (Skipping unstable BF16 configuration: hidde...) [ 11%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[True-4-dtype1-1.0-1.0-4-127-128-256] PASSED                                                   [ 12%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-None-dtype0-1.0-2e-06-1-1024-128-256] PASSED                                            [ 13%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-None-dtype0-1.0-2e-06-2-1024-64-256] PASSED                                             [ 13%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-None-dtype0-1.0-2e-06-4-127-128-256] PASSED                                             [ 14%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-None-dtype1-1.0-1.0-1-1024-128-256] PASSED                                              [ 15%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-None-dtype1-1.0-1.0-2-1024-64-256] SKIPPED (Skipping unstable BF16 configuration: h...) [ 15%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-None-dtype1-1.0-1.0-4-127-128-256] PASSED                                               [ 16%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-2-dtype0-1.0-2e-06-1-1024-128-256] PASSED                                               [ 17%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-2-dtype0-1.0-2e-06-2-1024-64-256] PASSED                                                [ 18%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-2-dtype0-1.0-2e-06-4-127-128-256] PASSED                                                [ 18%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-2-dtype1-1.0-1.0-1-1024-128-256] PASSED                                                 [ 19%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-2-dtype1-1.0-1.0-2-1024-64-256] SKIPPED (Skipping unstable BF16 configuration: hidd...) [ 20%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-2-dtype1-1.0-1.0-4-127-128-256] PASSED                                                  [ 20%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-4-dtype0-1.0-2e-06-1-1024-128-256] PASSED                                               [ 21%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-4-dtype0-1.0-2e-06-2-1024-64-256] PASSED                                                [ 22%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-4-dtype0-1.0-2e-06-4-127-128-256] PASSED                                                [ 22%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-4-dtype1-1.0-1.0-1-1024-128-256] PASSED                                                 [ 23%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-4-dtype1-1.0-1.0-2-1024-64-256] SKIPPED (Skipping unstable BF16 configuration: hidd...) [ 24%]
test/transformers/test_tiled_mlp.py::test_tiled_geglu_correctness[False-4-dtype1-1.0-1.0-4-127-128-256] PASSED                                                  [ 25%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-None-dtype0-1.0-2e-06-2-512-512-1024] PASSED                                            [ 25%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-None-dtype0-1.0-2e-06-1-1024-256-512] PASSED                                            [ 26%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-None-dtype0-1.0-2e-06-4-127-128-256] PASSED                                             [ 27%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-None-dtype1-1.0-1.0-2-512-512-1024] SKIPPED (bfloat16 tests disabled due to numeric...) [ 27%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-None-dtype1-1.0-1.0-1-1024-256-512] SKIPPED (bfloat16 tests disabled due to numeric...) [ 28%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-None-dtype1-1.0-1.0-4-127-128-256] SKIPPED (bfloat16 tests disabled due to numerica...) [ 29%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-2-dtype0-1.0-2e-06-2-512-512-1024] PASSED                                               [ 29%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-2-dtype0-1.0-2e-06-1-1024-256-512] PASSED                                               [ 30%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-2-dtype0-1.0-2e-06-4-127-128-256] PASSED                                                [ 31%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-2-dtype1-1.0-1.0-2-512-512-1024] SKIPPED (bfloat16 tests disabled due to numerical ...) [ 31%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-2-dtype1-1.0-1.0-1-1024-256-512] SKIPPED (bfloat16 tests disabled due to numerical ...) [ 32%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-2-dtype1-1.0-1.0-4-127-128-256] SKIPPED (bfloat16 tests disabled due to numerical i...) [ 33%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-4-dtype0-1.0-2e-06-2-512-512-1024] PASSED                                               [ 34%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-4-dtype0-1.0-2e-06-1-1024-256-512] PASSED                                               [ 34%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-4-dtype0-1.0-2e-06-4-127-128-256] PASSED                                                [ 35%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-4-dtype1-1.0-1.0-2-512-512-1024] SKIPPED (bfloat16 tests disabled due to numerical ...) [ 36%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-4-dtype1-1.0-1.0-1-1024-256-512] SKIPPED (bfloat16 tests disabled due to numerical ...) [ 36%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[True-4-dtype1-1.0-1.0-4-127-128-256] SKIPPED (bfloat16 tests disabled due to numerical i...) [ 37%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-None-dtype0-1.0-2e-06-2-512-512-1024] PASSED                                           [ 38%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-None-dtype0-1.0-2e-06-1-1024-256-512] PASSED                                           [ 38%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-None-dtype0-1.0-2e-06-4-127-128-256] PASSED                                            [ 39%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-None-dtype1-1.0-1.0-2-512-512-1024] SKIPPED (bfloat16 tests disabled due to numeri...) [ 40%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-None-dtype1-1.0-1.0-1-1024-256-512] SKIPPED (bfloat16 tests disabled due to numeri...) [ 40%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-None-dtype1-1.0-1.0-4-127-128-256] SKIPPED (bfloat16 tests disabled due to numeric...) [ 41%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-2-dtype0-1.0-2e-06-2-512-512-1024] PASSED                                              [ 42%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-2-dtype0-1.0-2e-06-1-1024-256-512] PASSED                                              [ 43%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-2-dtype0-1.0-2e-06-4-127-128-256] PASSED                                               [ 43%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-2-dtype1-1.0-1.0-2-512-512-1024] SKIPPED (bfloat16 tests disabled due to numerical...) [ 44%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-2-dtype1-1.0-1.0-1-1024-256-512] SKIPPED (bfloat16 tests disabled due to numerical...) [ 45%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-2-dtype1-1.0-1.0-4-127-128-256] SKIPPED (bfloat16 tests disabled due to numerical ...) [ 45%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-4-dtype0-1.0-2e-06-2-512-512-1024] PASSED                                              [ 46%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-4-dtype0-1.0-2e-06-1-1024-256-512] PASSED                                              [ 47%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-4-dtype0-1.0-2e-06-4-127-128-256] PASSED                                               [ 47%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-4-dtype1-1.0-1.0-2-512-512-1024] SKIPPED (bfloat16 tests disabled due to numerical...) [ 48%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-4-dtype1-1.0-1.0-1-1024-256-512] SKIPPED (bfloat16 tests disabled due to numerical...) [ 49%]
test/transformers/test_tiled_mlp.py::test_tiled_swiglu_correctness[False-4-dtype1-1.0-1.0-4-127-128-256] SKIPPED (bfloat16 tests disabled due to numerical ...) [ 50%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-2-256-512-1-2] PASSED                                                            [ 50%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-2-256-512-2-2] PASSED                                                            [ 51%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-2-256-512-4-2] PASSED                                                            [ 52%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-2-512-1024-1-2] PASSED                                                           [ 52%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-2-512-1024-2-2] PASSED                                                           [ 53%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-2-512-1024-4-2] PASSED                                                           [ 54%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-1-128-256-1-2] PASSED                                                            [ 54%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-1-128-256-2-2] PASSED                                                            [ 55%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype0-1e-05-1e-05-1-128-256-4-2] PASSED                                                            [ 56%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-2-256-512-1-2] PASSED                                                                [ 56%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-2-256-512-2-2] PASSED                                                                [ 57%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-2-256-512-4-2] PASSED                                                                [ 58%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-2-512-1024-1-2] PASSED                                                               [ 59%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-2-512-1024-2-2] PASSED                                                               [ 59%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-2-512-1024-4-2] PASSED                                                               [ 60%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-1-128-256-1-2] PASSED                                                                [ 61%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-1-128-256-2-2] PASSED                                                                [ 61%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_swiglu[dtype1-0.1-0.1-1-128-256-4-2] PASSED                                                                [ 62%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype0-0.001-0.001-2-256-512-1-2] PASSED                                                   [ 63%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype0-0.001-0.001-2-256-512-2-2] PASSED                                                   [ 63%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype0-0.001-0.001-2-256-512-4-2] PASSED                                                   [ 64%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype0-0.001-0.001-2-512-1024-1-2] PASSED                                                  [ 65%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype0-0.001-0.001-2-512-1024-2-2] PASSED                                                  [ 65%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype0-0.001-0.001-2-512-1024-4-2] PASSED                                                  [ 66%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype0-0.001-0.001-1-128-256-1-2] PASSED                                                   [ 67%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype0-0.001-0.001-1-128-256-2-2] PASSED                                                   [ 68%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype0-0.001-0.001-1-128-256-4-2] PASSED                                                   [ 68%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype1-0.1-0.1-2-256-512-1-2] PASSED                                                       [ 69%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype1-0.1-0.1-2-256-512-2-2] PASSED                                                       [ 70%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype1-0.1-0.1-2-256-512-4-2] PASSED                                                       [ 70%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype1-0.1-0.1-2-512-1024-1-2] PASSED                                                      [ 71%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype1-0.1-0.1-2-512-1024-2-2] PASSED                                                      [ 72%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype1-0.1-0.1-2-512-1024-4-2] PASSED                                                      [ 72%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype1-0.1-0.1-1-128-256-1-2] PASSED                                                       [ 73%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype1-0.1-0.1-1-128-256-2-2] PASSED                                                       [ 74%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_swiglu[dtype1-0.1-0.1-1-128-256-4-2] PASSED                                                       [ 75%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype0-0.001-0.001-2-256-512-1-2] PASSED                                                    [ 75%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype0-0.001-0.001-2-256-512-2-2] PASSED                                                    [ 76%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype0-0.001-0.001-2-256-512-4-2] PASSED                                                    [ 77%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype0-0.001-0.001-2-512-1024-1-2] PASSED                                                   [ 77%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype0-0.001-0.001-2-512-1024-2-2] PASSED                                                   [ 78%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype0-0.001-0.001-2-512-1024-4-2] PASSED                                                   [ 79%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype0-0.001-0.001-1-128-256-1-2] PASSED                                                    [ 79%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype0-0.001-0.001-1-128-256-2-2] PASSED                                                    [ 80%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype0-0.001-0.001-1-128-256-4-2] PASSED                                                    [ 81%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype1-0.1-0.1-2-256-512-1-2] PASSED                                                        [ 81%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype1-0.1-0.1-2-256-512-2-2] PASSED                                                        [ 82%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype1-0.1-0.1-2-256-512-4-2] PASSED                                                        [ 83%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype1-0.1-0.1-2-512-1024-1-2] PASSED                                                       [ 84%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype1-0.1-0.1-2-512-1024-2-2] PASSED                                                       [ 84%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype1-0.1-0.1-2-512-1024-4-2] PASSED                                                       [ 85%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype1-0.1-0.1-1-128-256-1-2] PASSED                                                        [ 86%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype1-0.1-0.1-1-128-256-2-2] PASSED                                                        [ 86%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_vs_torch_geglu[dtype1-0.1-0.1-1-128-256-4-2] PASSED                                                        [ 87%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype0-1e-05-1e-05-2-256-512-1-2] PASSED                                                             [ 88%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype0-1e-05-1e-05-2-256-512-2-2] PASSED                                                             [ 88%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype0-1e-05-1e-05-2-256-512-4-2] PASSED                                                             [ 89%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype0-1e-05-1e-05-2-512-1024-1-2] PASSED                                                            [ 90%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype0-1e-05-1e-05-2-512-1024-2-2] PASSED                                                            [ 90%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype0-1e-05-1e-05-2-512-1024-4-2] PASSED                                                            [ 91%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype0-1e-05-1e-05-1-128-256-1-2] PASSED                                                             [ 92%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype0-1e-05-1e-05-1-128-256-2-2] PASSED                                                             [ 93%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype0-1e-05-1e-05-1-128-256-4-2] PASSED                                                             [ 93%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype1-0.1-0.1-2-256-512-1-2] PASSED                                                                 [ 94%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype1-0.1-0.1-2-256-512-2-2] PASSED                                                                 [ 95%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype1-0.1-0.1-2-256-512-4-2] PASSED                                                                 [ 95%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype1-0.1-0.1-2-512-1024-1-2] PASSED                                                                [ 96%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype1-0.1-0.1-2-512-1024-2-2] PASSED                                                                [ 97%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype1-0.1-0.1-2-512-1024-4-2] PASSED                                                                [ 97%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype1-0.1-0.1-1-128-256-1-2] PASSED                                                                 [ 98%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype1-0.1-0.1-1-128-256-2-2] PASSED                                                                 [ 99%]
test/transformers/test_tiled_mlp.py::test_fsdp_tiled_geglu[dtype1-0.1-0.1-1-128-256-4-2] PASSED                                                                 [100%]
```
Let me know if you'd like further details or have specific scenarios you want verified.

test/transformers/test_tiled_mlp.py
alektebel: Could you clarify on the torch implementation as reference? FSDP would be already from torch. You mean using an MLP from torch, wrapping with a FSDP, and checking it works as expected (Or MLP vs TiledMLP + FSDP)? Or comparing the FSDP1 with FSDP2? At the moment, I just compared the TiledMLP wrapped with FSDP agains the LigerSwiGLUMLP wrapped with FSDP.

test/transformers/test_tiled_mlp.py
alektebel: Okey, done

src/liger_kernel/ops/tiled_mlp.py
```diff
@@ -74,28 +77,82 @@ def backward(ctx, *grads) -> tuple:
...
+
+        # Initialize param grad accumulators as None for lazy allocation
+        param_grads: List[Optional[torch.Tensor]] = [None for _ in params]
 
         x_shards = list(torch.chunk(x, chunks=shards, dim=0))
 
+        # Calculate cumulative offsets for correct gradient slicing when shards are uneven
+        shard_offset = 0
         for i, x_shard in enumerate(x_shards):
+            x_shard = x_shard.detach()
```
Tcc0403: Why do we need to detach?

src/liger_kernel/ops/tiled_mlp.py
```diff
@@ -74,28 +77,82 @@ def backward(ctx, *grads) -> tuple:
...
-        x_grad = torch.zeros_like(x)
+        x_grad = torch.zeros_like(x) if x_requires_grad else None
+
+        # Initialize param grad accumulators as None for lazy allocation
+        param_grads: List[Optional[torch.Tensor]] = [None for _ in params]
 
         x_shards = list(torch.chunk(x, chunks=shards, dim=0))
 
+        # Calculate cumulative offsets for correct gradient slicing when shards are uneven
+        shard_offset = 0
```
Tcc0403: Could you elaborate on the scenarios in which the original offset calculation fails? 
```py
shard_offset = i * x_shards[0].shape[0]
```

src/liger_kernel/ops/tiled_mlp.py
```diff
@@ -39,12 +39,14 @@ def forward(
         mlp_module: torch.nn.Module,
         x: torch.Tensor,
         shards: int,
-        compute_params: Optional[List[torch.nn.Parameter]] = None,
+        *params: torch.nn.Parameter,
```
Tcc0403: Could you explain why we use position  args instead of the original list?

src/liger_kernel/ops/tiled_mlp.py
```diff
@@ -74,28 +77,82 @@ def backward(ctx, *grads) -> tuple:
...
+                        grad_outputs=incoming_grad_shard,
+                    )
+                else:
+                    local_grads = []
+
+            # Process gradients
+            grad_idx = 0
+            if x_requires_grad and x_grad is not None:
+                x_grad.narrow(0, shard_offset, shard_step).copy_(local_grads[grad_idx])
+                grad_idx += 1
```
Tcc0403: Split local grad into local_x_grad and local_param_grad then we don't have to maintain awkward grad_idx logic

src/liger_kernel/ops/tiled_mlp.py
```diff
@@ -74,28 +77,82 @@ def backward(ctx, *grads) -> tuple:
...
+        # Return gradients: (fn, mlp_module, x, shards, *params)
+        # Clone param_grads to ensure they're not views into local_grads
+        final_param_grads = []
+        for param_idx, p in enumerate(params):
+            if param_grads[param_idx] is not None:
+                final_param_grads.append(param_grads[param_idx].clone())
+            else:
+                final_param_grads.append(torch.zeros_like(p))
+
+        return (None, None, x_grad, None, *final_param_grads)
```
Tcc0403:
```suggestion
        #  (fn, mlp_module, x, shards, *params)
        return (None, None, x_grad, None, *final_param_grads)
```

src/liger_kernel/ops/tiled_mlp.py
```diff
@@ -74,28 +77,82 @@ def backward(ctx, *grads) -> tuple:
...
+            x_grad = x_grad.view(x_shape_orig)
+
+        # Return gradients: (fn, mlp_module, x, shards, *params)
+        # Clone param_grads to ensure they're not views into local_grads
+        final_param_grads = []
+        for param_idx, p in enumerate(params):
+            if param_grads[param_idx] is not None:
+                final_param_grads.append(param_grads[param_idx].clone())
+            else:
+                final_param_grads.append(torch.zeros_like(p))
```
Tcc0403:
```suggestion
        # Clone param_grads to ensure they're not views into local_grads
        final_param_grads = [param_grad.clone() if param_grad is not None else torch.zeros_like(params[idx]) for idx, param_grad in enumerate(param_grads)]
```

Tcc0403 (commented):
I left some comments on areas where I didn't quite understand, hoping that we can improve readability with clarification 

Besides, the axolotl's hook approach seems to be applicable to your current implementation. I wonder if you are interested in trying this way

Ref: https://github.com/axolotl-ai-cloud/axolotl/blob/dd78f2e0cc5cc6458daaad02cc29b649ff1046f5/src%2Faxolotl%2Fmonkeypatch%2Ftiled_mlp%2Fbase.py

src/liger_kernel/ops/tiled_mlp.py
alektebel: This would be a consequence of using .grad() vs .autograd() 

* .backward() computes gradients on the leaf tensors, without having to return anything
* .grad() return tensors per input specified. 

In the past implementation, we only called .autograd(), which populated automatically the gradients, without having to worry about input definitions. The input in this case was a list of torch.nn.Parameter. Now that we are using .grad() instead of .autograd(), we need to take into account that .grad() returns a tuple with the gradients, depending on the inputs that we pass the function. I noticed this when expanding the test suite on the gradients.

src/liger_kernel/ops/tiled_mlp.py
alektebel: Actually, we don't need to detach per se, since x is the leaf of this graph, so we wouldn't extend the gradients back past x.

Back when I edited this part, I added this .detach(), thinking it would skip x.grad computation entirely as an optimization.

after looking into it, perhaps by calling autograd.grad() on the shards we can automatically Get x.grad by coalescing, instead of having to accumulate them manually.

This way, x.grad could be correctly computed when computing the x_shard[i].grad, due to how autograd.grad() works. What do you think of exploring this aproach?

src/liger_kernel/ops/tiled_mlp.py
alektebel: This assumes that all shards have an equal size, but in odd cases where n_tokens is not divisible by n_shards, this won't have a correct offset

src/liger_kernel/ops/tiled_mlp.py
alektebel: Okey, will change on next commit

src/liger_kernel/ops/tiled_mlp.py
alektebel: Good simplification, it makes sense

alektebel: Will change it on next commit

alektebel (commented):
Hi @Tcc0403 ,

Responding to the suggestions and comments about the code. I'll add the axolotl contrasting test in a follow up commit around this week.

Thank you

alektebel:
Hi @Tcc0403 ,

Sorry for the delay. I'm thoroughly reviewing the implementation so that it's as clean and correct as possible, making sure I'm implementing the hook based aproach correctly. I might need a little more time

Tcc0403:
No worries, take your time 😉

AlirezAkbary:
Hi @alektebel ! I previously volunteered to work on #893, and I’d like to collaborate on this draft rather than open a brand new PR.

I noticed that your April update mentioned refining the Axolotl-inspired hook approach, but the branch was subsequently force-pushed in June and July. 

The current implementation now recomputes all shards and performs a single combined backward instead of using hooks. Can you confirm whether this intentionally replaces the hook-based approach?

My main concern with the current implementation is that retaining every shard’s autograd graph until the combined backward may eliminate TiledMLP’s peak-memory advantage. 

I can start with a focused multi-GPU regression test covering gradient correctness and peak memory.

If you’re open to collaborating, would you prefer to add me as a collaborator to your fork, or have me prepare small commits in my fork for you to review and cherry-pick?

alektebel:
Hi @AlirezAkbary ! Thank you for volunteering. Yes, the progress here has been a bit stalled since I had a busy season at work, and I couldn’t find the time. I will add you as a collaborator.

The plan would be to implement a gradient accumulator based aproach inspired by Axolotl implementation in order to avoid the state inconsistencies that occur when combining both TiledMLP and FSDP algorithms together. I'm pushing the current work on the gradient -accumulator-based aproach today to the fork.

In the last WIP, I was basically cleaning up things from the PR, and making tests, so the multi-GPU tests you are proposing sound good. What were you planning to use for that? I was basically renting a really cheap pair of GPUs on vast.ai.

alektebel:
Hi @AlirezAkbary , @Tcc0403 , 

Done. For the sake of clarity, I have overriden the past commits, and I'm explaining below the content of each of them:

225ff573 -> Write the GradientAccumulator utility in utils.py.
ef68da98 -> Actually use the GradientAccumulator to accumulate the gradients, avoiding the state inconsistencies.
516beb8e ->  Tests with 2, 4, and 8 GPUs (skipped if missing) to assure that the defect no longer occurs, as well as behaviour against a vanilla PyTorch MLP.

@AlirezAkbary  What could be a good aproach to test this on 4, 8 gpus? Do you have any hardware for that?

Happy to hear any comments.

Thank you

AlirezAkbary:
Hi @alektebel, It looks like the invitation expired before I accepted it. Could you please resend it?

alektebel:
Done! Lmk if you want to talk about this PR vía discord @AlirezAkbary

AlirezAkbary:
Hi @alektebel ! thanks for adding me to the fork. 

@alektebel @Tcc0403 
I tested a simpler implementation of the same gradient-accumulation approach, and I completed the multi-GPU investigation. Tests are passing!


I have two branches:

- **Mergeable fix:** [`alireza/fsdp-tiledmlp-fix`](https://github.com/alektebel/Liger-Kernel/tree/alireza/fsdp-tiledmlp-fix)
- **Extended validation:** [`alireza/fsdp-tiledmlp-validation`](https://github.com/alektebel/Liger-Kernel/tree/alireza/fsdp-tiledmlp-validation)


### What belongs in the PR

The mergeable branch adds two commits on top of the current PR head:

1. the backward implementation
2. focused regression tests for micro-batch gradient accumulation and FSDP1 flat parameters (on top of the existing tests in this PR)


### Why there is a separate validation branch


The mergeable branch contains only the focused regression tests needed for this fix.

The validation branch answers the broader questions that came up during the investigation: does the fix work with DDP, FSDP1, and FSDP2; at 2, 4, and 8 GPUs; in FP32 and BF16; with gradient accumulation and `no_sync()`; and without losing TiledMLP’s memory advantage?

I kept this matrix separate because it requires up to eight GPUs and is too expensive for normal CI. It is reproducible evidence for reviewing the fix, not part of the proposed merge.


### Test setup

- GPUs: 8× NVIDIA A100-SXM4 40 GB
- Installation: `.[dev]` dependencies
- PyTorch: 2.14.0+cu130
- CUDA runtime: 13.0
- Triton: 3.8.0
- Transformers: 5.17.0
- pytest: 9.1.1
 
 
 ### Results

| Test suite | Result |
|---|---:|
| Mergeable test file | 63 passed, 24 existing BF16 skips |
| Distributed correctness matrix | 80 passed |
| `no_sync()` accumulation | 18 passed |
| Isolated memory comparisons | 24 passed |

The validation branch passed all **122 tests**.

In every memory comparison, tiled execution used **68–71% less temporary peak memory** than untiled execution.

I also saved the complete pytest logs, JUnit reports, package list, environment information, and exact commit metadata.


### Proposed PR update

The mergeable branch is based directly on the current PR head. If it looks good, my preferred option is to fast-forward the two commits (keeping the validation commit stays on its separate branch).

If you prefer, you can cherry-pick `f3aab5a` and `487b170` instead. Let me know which option you would like.

AlirezAkbary:
Hey @alektebel 
Just following up on this. If you're happy with the changes, I'm happy to fast-forward `fsdp-tiledmlp-support` with the two commits from `alireza/fsdp-tiledmlp-fix`.

PR #1129, opened 2026-03-04
Kolehma8/dist swiglu

kolehma8:
## Summary

D tensor support for swiglu. All computations are performed element wise with the local tensors and the output tensor distributed layout is mirrored from the inputs.

## Testing Done

Tested with 4 and 8 H100 GPUs locally using NCCL as the communication backend.

- Hardware Type: <BLANK>


src/liger_kernel/ops/swiglu.py
```diff
@@ -104,13 +104,31 @@ class LigerSiLUMulFunction(torch.autograd.Function):
     @staticmethod
     @ensure_contiguous
     def forward(ctx, a, b):
+        if isinstance(a, torch.distributed.tensor.DTensor):
```
vaibhavjindal: Can we also add a check for b? to make sure that a and mb are not incompatible?

test/convergence/fp32/test_mini_models.py
```diff
@@ -1618,12 +1618,12 @@ def run_mini_model(
                     not LLAMA4_AVAILABLE,
                     reason="Llama4 not available in this version of trasnformers",
                 ),
-                pytest.mark.xfail(
-                    reason=(
-                        "RuntimeError: Expected query, key, and value to have the same dtype, but got query.dtype:"
-                        " float key.dtype: c10::BFloat16 and value.dtype: c10::BFloat16 instead."
-                    )
-                ),
+                # pytest.mark.xfail(
```
vaibhavjindal, line 1621: need to uncomment it

kolehma8: Actually it was commented originally. make checkstyle added a space to make it pass the formatting checks.

src/liger_kernel/ops/swiglu.py
kolehma8: Added handling for case where the inputs are mixed regular and D-tensors.

PR #1138, opened 2026-03-10
[NPU]:Adding the single_block strategy to the softmax operator

TianHao324:
## Summary

Add single-block softmax kernel for small column sizes: 
    Processes entire row in one block when n_cols <= BLOCK_SIZE.
    Uses 2D tensor to process multiple rows simultaneously for better UB utilization.
The modified version has a performance improvement of about 50% compared to the original. However, it still has a gap compared to HuggingFace. We will continue to conduct research in the future.

## Testing Done

image


TianHao324:

PR #1139, opened 2026-03-10
[NPU] FIX bug in softmax

lowdy1:
## Summary
This PR fixes a bug in the softmax implementation where the loop iteration count was computed incorrectly. It also renames the functions to `_softmax_forward` and `_softmax_backward` to match the source kernels and avoid potential import errors.

## Testing Done
Device: Atlas 800I A2

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ops/softmax.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/softmax.py
@@ -326,7 +326,7 @@ class LigerSoftmaxFunction(torch.autograd.Function):
     @staticmethod
     @ensure_contiguous
     def forward(ctx, input_: torch.Tensor):
-        y, BLOCK_SIZE = _softmax_forward(input_)
+        y, BLOCK_SIZE, ROWS_PER_BLOCK, multi_block_launch = _softmax_forward(input_)
         ctx.save_for_backward(y)
         ctx.BLOCK_SIZE = BLOCK_SIZE
         ctx.ROWS_PER_BLOCK = ROWS_PER_BLOCK


lowdy1:
@Tcc0403 Could you please review this ASAP? This PR fixes a subtle bug in the NPU softmax implementation that causes incorrect behavior under certain conditions but still passes all tests.

src/liger_kernel/ops/backends/_ascend/ops/softmax.py
```diff
@@ -88,7 +88,8 @@ def _softmax_multi_block_forward_kernel(
         BLOCK_SIZE: Block size for column processing
     """
     row_start = tl.program_id(0)
-    row_step = tl.num_programs(0)
+    num_prog = tl.num_programs(0)
+    row_step = tl.cdiv(n_rows, num_prog)
```
Tcc0403, line 92: Shouldn't `tl.cdiv(n_rows, num_prog)` be the number of iterations each program requires? If it is larger than number of programs, wouldn't it skip some of indices? (referring to L95)

```py
for row_idx in tl.range(row_start, n_rows, row_step):
```

Tcc0403: @lowdy1 would you like to take a look?

lowdy1: Thanks for your careful review, and apologies for the subtle mistake. Refer to https://github.com/linkedin/Liger-Kernel/pull/1147  please.

PR #1141, opened 2026-03-11
[NPU]:Resolve the ub overflow issue of the kl_div operator

TianHao324:
## Summary

In the original code, the calculation logic is different when log_target is true or false, and the occupied ub is also different. Therefore, using a single logic to calculate the block size is inaccurate and will lead to ub overflow when log_target is false, because at this time, the ub occupies more space. So, we take log_target as a judgment condition when calculating ub.
image

## Testing Done

image


TianHao324:

PR #1144, opened 2026-03-12
[NPU] Add group norm support on NPU

orangeH25:
## Summary
This PR introduces a functional GroupNorm operator for Ascend NPU.

Key improvements:

- Fixes the runtime error `grid should be less than 65536!`  and `ub overflow` that occurs when the original GPU-oriented liger-kernel GroupNorm implementation is executed on NPU.
- Adjusts the kernel launch and tiling strategy to comply with Ascend NPU execution constraints.
- Resolves numerical accuracy issues with PyTorch reference outputs.

While the current implementation is still slower than the HuggingFace implementation in end-to-end benchmarks, it provides a stable and functional GroupNorm path for Ascend NPU.

This PR mainly focuses on correctness and NPU compatibility. Further kernel-level optimizations will be explored in follow-up work.
## Testing Done
image

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ops/group_norm.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/group_norm.py
@@ -143,8 +143,10 @@ def _group_norm_backward_kernel(
     N_inv = 1.0 / hidden_size
     row_offsets = tl.arange(0, BLOCK_SIZE_M)
     col_offsets_base = tl.arange(0, BLOCK_SIZE_N)
-    DW_scratch_base = DW_scratch_ptr + pid * DW_scratch_stride
-    DB_scratch_base = DB_scratch_ptr + pid * DB_scratch_stride
+
+    if COMPUTE_PARAM_GRAD:
+        DW_scratch_base = DW_scratch_ptr + pid * DW_scratch_stride
+        DB_scratch_base = DB_scratch_ptr + pid * DB_scratch_stride
 
     # Persistent-program loop over row tiles.
     for block_m in tl.range(pid, grid_m, num_progs):
@@ -240,11 +242,6 @@ def _group_norm_backward_kernel(
                     dB_partial = tl.sum(tl.where(mask, DY_block, 0.0), axis=1)
                     tl.atomic_add(DW_scratch_base + global_channel, dW_partial, mask=row_mask)
                     tl.atomic_add(DB_scratch_base + global_channel, dB_partial, mask=row_mask)
-                else:
-                    dW_block = tl.where(mask, DY_block * x_hat, 0.0)
-                    dB_block = tl.where(mask, DY_block, 0.0)
-                    tl.atomic_add(DW_scratch_base + global_channel, dW_block, mask=mask)
-                    tl.atomic_add(DB_scratch_base + global_channel, dB_block, mask=mask)
 
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


orangeH25:
Hi @Tcc0403 , please take a look. Thanks!

src/liger_kernel/ops/backends/_ascend/ops/group_norm.py
```diff
@@ -0,0 +1,476 @@
...
+    compute_param_grad = single_channel_tile
+
+    DX = torch.empty((batch_size, num_groups, hidden_size), dtype=X.dtype, device=X.device)
+    if compute_param_grad:
+        DW_scratch = torch.zeros((grid, num_channels), dtype=torch.float32, device=W.device)
+        DB_scratch = torch.zeros((grid, num_channels), dtype=torch.float32, device=W.device)
+    else:
+        # Placeholder buffers (unused in kernel when COMPUTE_PARAM_GRAD=False)
+        DW_scratch = torch.empty((1, 1), dtype=torch.float32, device=W.device)
+        DB_scratch = torch.empty((1, 1), dtype=torch.float32, device=W.device)
```
Tcc0403: If the 2-level sum approach I suggest doesn't work, is it possible to just pass None in triton-ascend backend

src/liger_kernel/ops/backends/_ascend/ops/group_norm.py
```diff
@@ -0,0 +1,476 @@
...
+                if SINGLE_CHANNEL_TILE:
+                    dW_partial = tl.sum(tl.where(mask, DY_block * x_hat, 0.0), axis=1)
+                    dB_partial = tl.sum(tl.where(mask, DY_block, 0.0), axis=1)
+                    tl.atomic_add(DW_scratch_base + global_channel, dW_partial, mask=row_mask)
+                    tl.atomic_add(DB_scratch_base + global_channel, dB_partial, mask=row_mask)
+                else:
+                    dW_block = tl.where(mask, DY_block * x_hat, 0.0)
+                    dB_block = tl.where(mask, DY_block, 0.0)
+                    tl.atomic_add(DW_scratch_base + global_channel, dW_block, mask=mask)
+                    tl.atomic_add(DB_scratch_base + global_channel, dB_block, mask=mask)
```
Tcc0403: L377 says
```
# Placeholder buffers (unused in kernel when COMPUTE_PARAM_GRAD=False)
```
, which contradicts what this block does. Shouldn't it be no-op here?

src/liger_kernel/ops/backends/_ascend/ops/group_norm.py
Tcc0403: Can placeholder buffers set to `None` in triton-ascend? to avoid accidently access in device code.

src/liger_kernel/ops/backends/_ascend/ops/group_norm.py
orangeH25: Removed.
> I originally kept it to preserve the kernel structure for potential future experiments, but it’s not needed in the current implementation.

src/liger_kernel/ops/backends/_ascend/ops/group_norm.py
orangeH25: Addressed.

orangeH25:
Hi @Tcc0403, changes applied. Appreciate another review when you have time, thanks!

src/liger_kernel/ops/backends/_ascend/ops/group_norm.py
```diff
@@ -0,0 +1,474 @@
...
+                DX_ptr + batch_idx[:, None] * X_row_stride + group_idx[:, None] * X_col_stride + col_offsets[None, :]
+            )
+            tl.store(DX_ptrs, DX_block.to(X_ptr.dtype.element_ty), mask=mask)
+
+            if COMPUTE_PARAM_GRAD:
+                if SINGLE_CHANNEL_TILE:
+                    dW_partial = tl.sum(tl.where(mask, DY_block * x_hat, 0.0), axis=1)
+                    dB_partial = tl.sum(tl.where(mask, DY_block, 0.0), axis=1)
+                    tl.atomic_add(DW_scratch_base + global_channel, dW_partial, mask=row_mask)
+                    tl.atomic_add(DB_scratch_base + global_channel, dB_partial, mask=row_mask)
```
Tcc0403, line 244: I wonder if we can accumulate dw and db over grid loop and store it after, similar to https://github.com/linkedin/Liger-Kernel/blob/db14ea856fc58b113a94fa78fdeebbdaae440d08/src/liger_kernel/ops/rms_norm.py#L390

With this approach, we can avoid using atomic_add and potentially handle the scenario where `num_col_blocks>1`.

The solution is not trivial and not gauranteed to achieve better performance, leaving the comment here as a future works direction.

Tcc0403 (approved):
LGTM! I left a comment about a potential improvement, but it can be done in another PR!

src/liger_kernel/ops/backends/_ascend/ops/group_norm.py
orangeH25: Thanks for the suggestion! I’ll look into this.

PR #1153, opened 2026-03-19
[Optimize, NPU] Remove tl.where from _rms_norm_forward/backward_kernel_tiled()

zheliuyu:
## Summary

When the `mask` has a large shape, `tl.where` is not NPU-friendly in `triton-ascend`, leading to low kernel performance. When writing kernels, it's best to use alternative logic instead. This can result in significant performance improvements.

**Will these changes affect accuracy?** Since the masking operation has already been applied when loading `X_block`, it will not affect the calculation result during `tl.sum`.

### Accuracy first

The shapes in `test_rms_norm.py` are too small to trigger the `_rms_norm_forward_kernel_tiled` kernel, so we need a new configuration.

```python
@pytest.mark.parametrize(
    "bs, sl, hd",
    [
        (2, 2048, 4096),
        (2, 2048, 8192),
        (2, 2048, 16384),
        (2, 2048, 32768),
    ],
)
@pytest.mark.parametrize(
    "dtype, atol, rtol",
    [
        (torch.float32, 1e-4, 1e-6),
    ],
)
@pytest.mark.parametrize(
    "reference, offset, casting_mode",
    [
        (LlamaRMSNorm, 0.0, "llama"),
        (GemmaRMSNorm, 1.0, "gemma"),
        pytest.param(
            BaseRMSNorm,
            0.0,
            "none",
            marks=pytest.mark.skipif(device == "npu", reason="Ascend NPU does not support this test"),
        ),
    ],
)
@pytest.mark.parametrize(
    "in_place",
    [
        True,
        False,
    ],
)
@pytest.mark.parametrize(
    "elementwise_affine",
    [
        True,
        False,
    ],
)
```

#### Env
image

#### Results after code modification

image

### Benchmark test

The test cases in `benchmark_rms_norm.py` should keep the same shapes as those in `test_rms_norm.py`.

```python
common_configs = {
    ...
    "x_values": [2**i for i in range(12, 16)],
    ...
}
```

#### Before Optimization

forward
rms_norm_speed_forward

backward
rms_norm_speed_backward

full
rms_norm_speed_full

memory
rms_norm_memory_full

[all_benchmark_data_raw.csv](https://github.com/user-attachments/files/26131388/all_benchmark_data_raw.csv)

#### After Optimization

forward
rms_norm_speed_forward

backward
rms_norm_speed_backward

full
rms_norm_speed_full

memory
rms_norm_memory_full

[all_benchmark_data_optimized.csv](https://github.com/user-attachments/files/26131403/all_benchmark_data_optimized.csv)

- Hardware Type: Atlas 900 A2 PoD

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ops/rms_norm.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/rms_norm.py
@@ -199,7 +199,7 @@ def _rms_norm_forward_kernel_tiled(
                 X_block = X_block.to(tl.float32)
 
             # Accumulate sum of squares (only for valid elements)
-            sum_square += tl.sum(tl.where(mask, X_block * X_block, 0.0))
+            sum_square += tl.sum(X_block * X_block)
 
         # Compute rstd for this row
         mean_square = sum_square / n_cols
@@ -456,7 +456,7 @@ def _rms_norm_backward_kernel_tiled(
                     m = dY_block
 
             # Accumulate sum(m * X)
-            sum_m_X += tl.sum(tl.where(mask, m * X_block, 0.0))
+            sum_m_X += tl.sum(m * X_block)
 
         # Compute the correction factor
         correction_factor = -(1.0 / n_cols) * rstd * rstd * sum_m_X


zheliuyu:
@Tcc0403 @TianHao324  PR is ready for review. Thanks.

PR #1196, opened 2026-04-17
[Gemma 4] Add apply_liger_kernel_to_gemma4_text (dense text, 31B-targeted)

lardinator:
## Summary

Adds Liger-Kernel support for Google's **Gemma 4** family (text, dense variants — primary target `google/gemma-4-31B`). Follows the exact integration pattern of the Gemma 3 port: module-level class swaps (`Gemma4RMSNorm`, `Gemma4TextMLP`), class-forward swap on `Gemma4ForCausalLM`, and `MODEL_TYPE_TO_APPLY_LIGER_FN["gemma4_text"]` registration so `AutoLigerKernelForCausalLM` routes transparently.

### What you get

- `LigerRMSNormForGemma4` (ones-init, no `(1+w)` offset, fp32 compute, handles `with_scale=False` for `v_norm`)
- `LigerGEGLUMLPForGemma4` (thin wrapper over `LigerGEGLUMLP` that absorbs `Gemma4TextMLP`'s `layer_idx` arg)
- Fused-linear-CE `causal_forward` on both `Gemma4ForCausalLM` (stock HF) and `Gemma4TextForCausalLM` (some users extract this subclass to dodge HF issue #45200's mm_token_type_ids check)
- Registration + exports from `liger_kernel.transformers`

### What you don't get (explicit non-goals)

- Multimodal (`Gemma4ForConditionalGeneration`, `Gemma4VisionModel`, `Gemma4AudioModel`)
- MoE (`Gemma4TextExperts`/`Gemma4TextRouter`, used by 26B-A4B)
- PLE (Per-Layer Embeddings — used by E2B/E4B)
- Double-wide MLP on KV-shared layers (31B doesn't use either)
- **Rope kernel swap**: HF Gemma 4's `apply_rotary_pos_emb(x, cos, sin)` is single-tensor; `liger_rotary_pos_emb(q, k, cos, sin)` is dual-tensor. Signatures are incompatible. `rope=True` is a no-op with a warning. The large memory win (LCE) is unaffected.

## Headline: Peak HBM at seq_len=8192

Measured on **AMD MI250X (LUMI)** with a 6-layer Gemma-4-shaped mini model at full 31B vocab (262,144), `bf16`, batch=1, seq=8192, forward+backward:

| Peak HBM | Value |
|---|---|
| HF baseline | **41.47 GB** |
| Liger (patched) | **10.89 GB** |
| **Saved** | **30.58 GB (73.7% reduction)** |

Driver: fused-linear-CE (`skip_logits=True`) eliminates the 262,144 × 8192 × bf16 logits tensor (~4 GB materialized) **plus** its gradient, activation buffers, and softcap intermediates — combined ~30 GB freed.

Loss parity on the same configuration: `loss_HF = 12.6841440`, `loss_liger = 12.6840858` (abs diff ~6e-5 — numerically identical).

### fp32 (kernel correctness)

Whole-model forward, 6-layer Gemma-4-shaped config, vocab 32000, shape `(2, 256, 32000)`:

| statistic | value |
|---|---|
| max \|logits_liger − logits_hf\| | **2.55e-03** |
| mean abs diff | 5.11e-05 |
| p99 abs diff | 3.84e-04 |

### bf16 (expected dtype noise)

| statistic | value |
|---|---|
| max abs diff | 7.97e-01 |
| mean abs diff | 3.05e-02 |
| p99 abs diff | 1.92e-01 |

Interpretation: kernels are numerically correct (fp32 max 2.55e-3); bf16 drift is 6-layer dtype-inherent precision noise, well inside industry-standard ranges.

## Discovered during LUMI verification (fixed in-PR)

Three plan-level assumptions that only showed up under real model construction — all fixed here:

1. **`v_norm` uses `with_scale=False`** — no weight parameter, so the naive subclass broke class-level swap. Fix: `LigerRMSNormForGemma4` accepts `with_scale` kwarg and delegates forward to a plain-torch path when weight is absent.
2. **`Gemma4TextMLP(config, layer_idx)` signature** (vs `LigerGEGLUMLP(config)`) — crashed at layer construction. Fix: `LigerGEGLUMLPForGemma4` wrapper absorbs the extra arg.
3. **`apply_rotary_pos_emb(x, cos, sin)` single-tensor signature** (vs `liger_rotary_pos_emb(q, k, cos, sin)`) — incompatible. Fix: `rope=True` becomes a no-op + warning for Gemma 4.

Plus: `Gemma4RMSNorm` inherits `Gemma3nRMSNorm` (not `Gemma3RMSNorm`) — reusing `LigerRMSNormForGemma3` would silently diverge training via its `(1+w)` offset. New subclass is mandatory, not cosmetic.

## Testing Done

- Hardware Type: AMD MI250X (LUMI HPC)

### CPU-runnable (included in `make test`)

\`\`\`
$ python -m pytest test/transformers/test_monkey_patch.py::test_apply_liger_kernel_to_instance_for_gemma4_text -v
test/transformers/test_monkey_patch.py::test_apply_liger_kernel_to_instance_for_gemma4_text PASSED [100%]
======================== 1 passed, 1 warning in 21.27s =========================
\`\`\`

### checkstyle

\`\`\`
$ ruff check --output-format=concise .
All checks passed!
$ ruff format --check .
264 files already formatted
\`\`\`

### LUMI (MI250X, ROCm 6.2.4, PyTorch 2.7.1, transformers 5.5.4)

- [x] `pytest test/convergence/bf16/test_mini_models.py -k mini_gemma4_text` — MI250X ROCm
- [x] `pytest test/convergence/bf16/test_mini_models_with_logits.py -k mini_gemma4_text` — MI250X ROCm
- [x] Gemma 3 regression (3 tests) — all pass, no Gemma 3 behaviour change
- [x] Revert helper functional test — identical outputs (`max diff 0.00e+00`) after revert
- [x] HBM benchmark at seq_len=8192 — **30.58 GB saved**, **73.7% reduction**

Tolerance selection for `mini_gemma4_text`:
- `loss_atol=5e-2` (vs gemma3's `1e-2`) — 6-layer mini (minimum for sliding+global cycle) has ~0.05 bf16 drift vs gemma3's 4-layer which fits 1e-2
- `logprobs_atol=5e-1` — accumulated bf16 noise flips 3 of ~20,480 top-k logprob ranks on near-tie tokens

---

🤖 Drafted with [Claude Code](https://claude.com/claude-code) (Claude Opus 4.7). No auto-generation skill was used end-to-end — scaffolding and LUMI-verified fixes were developed and reviewed interactively.


src/liger_kernel/transformers/geglu.py
```diff
@@ -20,3 +20,18 @@ def __init__(self, config):
...
+
+    HF's Gemma4TextMLP is instantiated as ``Gemma4TextMLP(config, layer_idx)``;
+    swapping in plain LigerGEGLUMLP (single-arg) breaks model construction.
+    This subclass accepts and ignores ``layer_idx`` — 31B has
+    ``use_double_wide_mlp=false``, so the layer_idx never needed to feed the
+    doubled intermediate_size path. Forward is inherited unchanged.
+    """
+
+    # Internal monkey-patch helper only — not part of the public API surface.
+    def __init__(self, config, layer_idx=None):
```
Mecoli1219, line 35: Even though the 31B target has `use_double_wide_mlp=False`, a class-level swap affects all model constructions, so someone loading a Gemma 4 variant with double-wide MLP enabled would get silently incorrect weights. 

Can you match HF's conditional doubling logic: https://github.com/huggingface/transformers/blob/74a2a4d0c790788a3b65b975510577f8dc6e81d5/src/transformers/models/gemma4/modeling_gemma4.py#L1030-L1032 to keep the 31B path unchanged (no doubling) while correctly handling future variants?

src/liger_kernel/transformers/monkey_patch.py
```diff
@@ -3182,6 +3328,7 @@ def __init__(self, hidden_size, eps=1e-6, **kwargs):
     "gemma2": apply_liger_kernel_to_gemma2,
     "gemma3_text": apply_liger_kernel_to_gemma3_text,
     "gemma3": apply_liger_kernel_to_gemma3,
+    "gemma4_text": apply_liger_kernel_to_gemma4_text,
```
Mecoli1219, line 3331: Can we also support `gemma4` for users who load via `Gemma4Config`

src/liger_kernel/transformers/monkey_patch.py
```diff
@@ -1238,6 +1239,151 @@ def apply_liger_kernel_to_gemma3(
             raise TypeError("The model must be Gemma3ForConditionalGeneration.")
 
 
+def apply_liger_kernel_to_gemma4_text(
+    rope: bool = True,
```
Mecoli1219: Can you change the default value to False given that it is a no-op here?

Mecoli1219 (changes requested):
Overall looks good. Really exciting to have Gemma 4 patched in Liger! Thanks for the contribution! Left a few comments, mostly around edge cases for future Gemma 4 variants. 

Related to https://github.com/linkedin/Liger-Kernel/issues/1186. Still missing multimodal (`Gemma4ForConditionalGeneration`) and MoE (`Gemma4TextExperts`) 
  patching.

lardinator:
Thanks for the review @Mecoli1219! Addressed all three points:

**1. Double-wide MLP edge case** (`geglu.py`)
`LigerGEGLUMLPForGemma4` now replicates HF's conditional logic — re-creates `gate_proj`/`up_proj`/`down_proj` with `2 * intermediate_size` when `config.use_double_wide_mlp=True` AND the layer is KV-shared (`layer_idx >= num_hidden_layers - num_kv_shared_layers > 0`). The 31B production path (`use_double_wide_mlp=False`, `num_kv_shared_layers=0`) is unchanged.

**2. `gemma4` model type** (`monkey_patch.py:3331`)
Added `"gemma4": apply_liger_kernel_to_gemma4_text` to `MODEL_TYPE_TO_APPLY_LIGER_FN`.

**3. `rope` default** (`monkey_patch.py:1243`)
Changed default from `True` to `False` since it's a documented no-op on Gemma 4.

Added 7 unit tests in `test/transformers/test_geglu.py` covering:
- `layer_idx=None` default (no doubling)
- `use_double_wide_mlp=False` (31B config — no doubling)
- Non-KV-shared layers with flag enabled (no doubling)
- KV-shared layers with flag enabled (2x doubled, shapes verified)
- `num_kv_shared_layers=0` edge case (no doubling)
- Forward/backward shape correctness with doubled MLP

Re: multimodal (`Gemma4ForConditionalGeneration`) and MoE (`Gemma4TextExperts`) — happy to tackle those in a follow-up PR once this lands, if that works for you.

src/liger_kernel/transformers/monkey_patch.py
```diff
@@ -3329,6 +3329,7 @@ def __init__(self, hidden_size, eps=1e-6, **kwargs):
     "gemma3_text": apply_liger_kernel_to_gemma3_text,
     "gemma3": apply_liger_kernel_to_gemma3,
     "gemma4_text": apply_liger_kernel_to_gemma4_text,
+    "gemma4": apply_liger_kernel_to_gemma4_text,
```
Mecoli1219: We should move this to the PR that supports the multimodal `Gemma4ForConditionalGeneration`.

Mecoli1219 (changes requested):
Hi @lardinator! Thanks for the quick turnaround on the review comments. The double-wide MLP fix and tests look great. Overall LGTM. I'll approve the PR after the `MODEL_TYPE_TO_APPLY_LIGER_FN` is fixed, sync with main branch, and stylecheck pass.       
                                                                                                                  
For the follow-ups (multimodal Gemma4ForConditionalGeneration and MoE Gemma4TextExperts), we can do that in separate PRs. Also, we have some Claude skills in this repo: https://github.com/linkedin/Liger-Kernel/tree/main/.claude/skills, including one for automatically generating monkey-patches. Feel free to try them out in the follow-up development!

Mecoli1219:
Hi @lardinator . Can you fix the checkstyle? Thanks

Mecoli1219 (approved):
LGTM! Thanks for your contribution!

thad0ctor:
@lardinator @Mecoli1219 

I was working on a local fork of this PR and testing on my setup and noticed the following on my 5090 and RTX 6000 (sm_120) when one of the tests failed:

### `mini_gemma4_text` logprobs_atol=5e-1 fails on Blackwell

**Why this matters:** the test will fail in CI / locally for any contributor
on Blackwell hardware (RTX 5090, RTX PRO 6000, B100/B200) — false-negative
that blocks PRs unrelated to Gemma 4.

**Failure** (sm_120, CUDA 12.8, torch 2.9.1):
```
Mismatch at index (4, 123, 1):
  HF    = -7.4056
  Liger = -7.9989
  drift = 0.5933   (5e-1 + 0.093)
```
1 of ~20k top-k slots, deterministic at the test's seed.

**Cause:** cuBLAS bf16 GEMM (HF path) and Triton bf16 GEMM (Liger path) pick
different tensor-core kernels on sm_100/sm_120 than on sm_80/86/89/90. The
fp32 reference is identical; the gap between two equivalent bf16 routes at
one near-tie top-k slot widens from ~0.5 to ~0.59.

**Why a tolerance bump is a stopgap, not a fix.** The 0.59 single-slot delta
is the visible tip; the underlying noise floor is across the full vocab × seq
× batch logit tensor. Configs that consume logprobs as scalars amplify it:
- **DPO / IPO / KTO**: `β × Δlogprob` per pair → bias-prone if Triton-vs-cuBLAS
  rounding isn't symmetric across token IDs.
- **PPO / RLHF KL term**: cross-arch reference-vs-policy scoring bakes in a
  systematic offset.
- **Speculative decoding**: arch-mismatched draft/target reduces acceptance.
- **E2B specifics**: PLE injection norms accumulate per-layer additively;
  KV-shared layers consume earlier layers' K/V; double-wide MLP doubles the
  bf16 accumulation surface — all compound the per-layer drift.
- **Long pretraining**: Liger-on-Blackwell and Liger-on-Hopper diverge into
  different bit-exact checkpoints from the same seed.

For most workloads this is noise floor (loss/params match, Adam smooths
per-step noise). For the configs above it's a real risk that bumping the
tolerance hides.

**Fix options** (cleanest first):
1. **Cast lm_head matmul to fp32** in the FLCE path on bf16 models. Closes
   the gap at the most-visible boundary (logits → logprobs). Small memory
   cost; eliminates cross-arch divergence in eval logprobs.
2. **Compare top-k sets, not per-slot deltas.** Mask near-ties (rank
   ambiguity that bf16 inherently can't disambiguate) and assert on stable
   slots. Catches real regressions, ignores arch-noise rank flips.
3. **`consider torch.set_float32_matmul_precision('high')` in the test fixture** —
   forces tf32 accumulators in cuBLAS, often tightens HF's path enough.
4. **Per-arch tolerance gate**: `5e-1` for sm < 100, `7e-1` for sm ≥ 100,
   via `torch.cuda.get_device_capability()`.

**Repro:**
```bash
pytest -v --no-cov "test/convergence/bf16/test_mini_models.py::test_mini_model[mini_gemma4_text-32-1e-05-dtype26-0.05-0.01-0.5-0.01-0.01-0.01]"
```
Pass on Ampere, fail on sm_120.

Mecoli1219:
Hi @thad0ctor , thank you for testing it out on other types of gpu! It should be fine for some numerical drift in convergence tests, especially in bf16 precision.

Issue #956, opened 2025-11-28
[Feature Request] Add SwiGLU support for Qwen3-VL models


matthewdm0816:
### Problem Description
Can we add **SwiGLU** kernel support for **Qwen3-VL** models in Liger-Kernel?

Currently, it appears that passing `swiglu=True` in the `apply_liger_kernel_to_qwen3_vl` monkey patch function is a no-op and does not activate the kernel.

### References
* **Original PR:** #911
* **Relevant Discussion:** [Comment Link](https://github.com/linkedin/Liger-Kernel/pull/911#issuecomment-3585290131)

dongchany:
Opened PR #1175 for this.

  It fixes the missing Qwen3-VL SwiGLU patch path so `apply_liger_kernel_to_qwen3_vl(..., swiglu=True)` is
  no longer a no-op.

  What was missing:
  - `Qwen3VLTextMLP` was not being replaced with `LigerSwiGLUMLP`
  - instantiated `decoder_layer.mlp` modules were not being patched either

  Validation run locally:
  - `python -m pytest -q test/transformers/test_monkey_patch.py -k 'qwen3_vl and not moe'`
  - `python -m pytest -q test/convergence/bf16/test_mini_models.py -k 'mini_qwen3_vl and test_mini_model'`
  - `python -m ruff check src/liger_kernel/transformers/monkey_patch.py test/transformers/
  test_monkey_patch.py README.md`

PR #1201, opened 2026-04-25
Add Falcon H1 SwiGLU multipliers support (#936)

### LINKED ISSUE #936: Add support of multipliers for Swiglu liger Kernels

Falcon H1 MLP block uses gate and down proj multipliers. Currently this is unsupported for Liger Kernels.

Can we add support of Swiglu MLP Liger kernel with gate and down proj multipliers ?

```
class FalconH1MLP(nn.Module):
    def __init__(self, config: FalconH1Config):
        super().__init__()
        self.config = config
        self.hidden_size = config.hidden_size
        self.intermediate_size = config.intermediate_size
        self.gate_proj = nn.Linear(self.hidden_size, self.intermediate_size, bias=config.mlp_bias)
        self.up_proj = nn.Linear(self.hidden_size, self.intermediate_size, bias=config.mlp_bias)
        self.down_proj = nn.Linear(self.intermediate_size, self.hidden_size, bias=config.mlp_bias)
        self.act_fn = ACT2FN[config.hidden_act]
        self.gate_multiplier, self.down_multiplier = config.mlp_multipliers

    def forward(self, x):
        y = self.up_proj(x) * self.act_fn(self.gate_proj(x) * self.gate_multiplier)
        y = self.down_proj(y) * self.down_multiplier
        return y
```

### Alternatives

_No response_

### Additional context

_No response_

shivam2199:
## Summary

Resolves #936. Adds Liger SwiGLU support for Falcon H1's MLP block, which pre-scales the gate pre-activation and post-scales the down-projection output:

y = down_proj(silu(gate_proj(x) * gate_mult) * up_proj(x)) * down_mult

Previously `apply_liger_kernel_to_falcon_h1(swiglu=True)` was a no-op that just logged a warning; now it actually patches the MLP.

## Details

- `src/liger_kernel/ops/swiglu.py` — forward/backward Triton kernels extended with a `gate_multiplier` scalar applied to the gate pre-activation in fp32 before silu. Backward chain rule pulls an extra `gate_multiplier` factor into `da`. `LigerSiLUMulFunction` accepts `gate_multiplier` and `down_multiplier` as optional args (defaults 1.0), so all existing callers (Llama, Qwen, Granite, Phi3, Olmo, Hunyuan, etc.) are unaffected. `down_multiplier` is applied outside the kernel in the autograd function — a single elementwise multiply on the output, near-zero overhead.
- `src/liger_kernel/transformers/swiglu.py` — new `LigerFalconH1SwiGLUMLP` reads `config.mlp_multipliers` and forwards them to `LigerSiLUMulFunction`. Falls back to `self.config.mlp_multipliers` when patched onto an already-instantiated HF module via `_patch_swiglu_module` (which only rebinds `forward`).
- `src/liger_kernel/transformers/monkey_patch.py` — flip `apply_liger_kernel_to_falcon_h1`'s `swiglu` default from `False` to `True`, wire it to `LigerFalconH1SwiGLUMLP`, and fix the decoder-layer attribute to `feed_forward` (Falcon H1's actual HF attribute name).
- `src/liger_kernel/transformers/__init__.py` — export `LigerFalconH1SwiGLUMLP`.
- `test/transformers/test_swiglu.py` — three new parametrized tests (21 total cases):
  - `test_correctness_silumul_with_multipliers` — 12 cases (3 multiplier combos × 2 shapes × 2 dtypes) validating kernel math against a pure PyTorch reference.
  - `test_silumul_default_multipliers_backward_compat` — confirms no-arg `LigerSiLUMulFunction.apply(a, b)` is bit-identical to explicit `(1.0, 1.0)`.
  - `test_correctness_falcon_h1_mlp` — 8 cases end-to-end parity for `LigerFalconH1SwiGLUMLP` against a pure-PyTorch Falcon H1 MLP reference.

Out of scope: Qwen3-VL `swiglu=True` no-op (#956) — separate monkey-patch wiring issue, not kernel math.

## Testing Done

- Hardware Type: NVIDIA A10G (g5.xlarge)

Notes on `make test`:
- All 21 new tests pass.
- All pre-existing SwiGLU tests that exercise `LigerSiLUMulFunction` pass (Llama, Phi3, Mixtral fp32, functional, falcon_h1 monkey_patch).
- Unrelated pre-existing failures on `main` that also show up here (confirmed by running the same tests on `main`):
  - `test_correctness_mixtralexperts[bf16]` — Triton 3.2 `atomic_add does not support bf16` in `src/liger_kernel/ops/fused_moe.py`.
  - `test_apply_liger_kernel_to_instance_for_paligemma` — `SiglipVisionModel` attribute change in transformers v5.
  - `test_apply_liger_kernel_to_instance_for_gemma3_conditional_generation` — `Gemma3Config` v5 API change.
  - `test_grpo_loss::test_correctness[sequence-luspo-...*-8-128-1024-4096]` (3 large-shape cases) — pre-existing GRPO numerical sensitivity.

Convergence tests were not run locally (single-GPU box). Happy to run them if a maintainer can advise on the expected workflow, or to add a Falcon H1 mini-model entry in a follow-up.

### DIFF
--- a/src/liger_kernel/transformers/monkey_patch.py
+++ b/src/liger_kernel/transformers/monkey_patch.py
@@ -2688,7 +2688,7 @@ def apply_liger_kernel_to_falcon_h1(
 
         for decoder_layer in base_model.layers:
             if swiglu:
-                _patch_swiglu_module(decoder_layer.mlp, LigerFalconH1SwiGLUMLP)
+                _patch_swiglu_module(decoder_layer.feed_forward, LigerFalconH1SwiGLUMLP)
             if rms_norm:
                 _patch_rms_norm_module(decoder_layer.input_layernorm)
                 _patch_rms_norm_module(decoder_layer.pre_ff_layernorm)
--- a/src/liger_kernel/transformers/monkey_patch.py
+++ b/src/liger_kernel/transformers/monkey_patch.py
@@ -2627,7 +2627,7 @@ def apply_liger_kernel_to_falcon_h1(
     cross_entropy: bool = False,
     fused_linear_cross_entropy: bool = True,
     rms_norm: bool = True,
-    swiglu: bool = False,
+    swiglu: bool = True,
     model: PreTrainedModel = None,
 ) -> None:
     """
@@ -2652,6 +2652,8 @@ def apply_liger_kernel_to_falcon_h1(
     from transformers.models.falcon_h1 import modeling_falcon_h1
     from transformers.models.falcon_h1.modeling_falcon_h1 import FalconH1Model
 
+    from liger_kernel.transformers.swiglu import LigerFalconH1SwiGLUMLP
+
     if rope:
         logger.info("Apply liger rotary pos emb.")
         modeling_falcon_h1.apply_rotary_pos_emb = liger_rotary_pos_emb
@@ -2659,7 +2661,8 @@ def apply_liger_kernel_to_falcon_h1(
         logger.info("Apply liger RMSNorm")
         modeling_falcon_h1.FalconH1RMSNorm = LigerRMSNorm
     if swiglu:
-        logger.warning("LigerSwiGLUMLP is not available for Falcon-H1 models. There will be no effect.")
+        logger.info("Apply liger SwiGLU")
+        modeling_falcon_h1.FalconH1MLP = LigerFalconH1SwiGLUMLP
 
     if cross_entropy:
         logger.info("Apply liger cross entropy")
@@ -2685,7 +2688,7 @@ def apply_liger_kernel_to_falcon_h1(
 
         for decoder_layer in base_model.layers:
             if swiglu:
-                _patch_swiglu_module(decoder_layer.mlp, LigerSwiGLUMLP)
+                _patch_swiglu_module(decoder_layer.mlp, LigerFalconH1SwiGLUMLP)
             if rms_norm:
                 _patch_rms_norm_module(decoder_layer.input_layernorm)
                 _patch_rms_norm_module(decoder_layer.pre_ff_layernorm)
--- a/src/liger_kernel/transformers/swiglu.py
+++ b/src/liger_kernel/transformers/swiglu.py
@@ -135,3 +135,47 @@ class LigerHunyuanV1SwiGLUMLP(nn.Module):
 
     def forward(self, x):
         return self.down_proj(LigerSiLUMulFunction.apply(self.gate_proj(x), self.up_proj(x)))
+
+
+class LigerFalconH1SwiGLUMLP(nn.Module):
+    """
+    Patch FalconH1MLP to use LigerSiLUMulFunction with gate / down multipliers.
+
+    Falcon H1's MLP block pre-scales the gate pre-activation and post-scales the
+    down projection output:
+
+        y = down_proj(silu(gate_proj(x) * gate_mult) * up_proj(x)) * down_mult
+
+    https://github.com/huggingface/transformers/blob/main/src/transformers/models/falcon_h1/modeling_falcon_h1.py
+    """
+
+    def __init__(self, config):
+        super().__init__()
+        self.config = config
+        self.hidden_size = config.hidden_size
+        self.intermediate_size = config.intermediate_size
+        bias = getattr(config, "mlp_bias", False)
+        self.gate_proj = nn.Linear(self.hidden_size, self.intermediate_size, bias=bias)
+        self.up_proj = nn.Linear(self.hidden_size, self.intermediate_size, bias=bias)
+        self.down_proj = nn.Linear(self.intermediate_size, self.hidden_size, bias=bias)
+        if config.hidden_act not in ["silu", "swish"]:
+            raise ValueError(f"Activation function {config.hidden_act} not supported.")
+        gate_multiplier, down_multiplier = config.mlp_multipliers
+        self.gate_multiplier = float(gate_multiplier)
+        self.down_multiplier = float(down_multiplier)
+
+    def forward(self, x):
+        # When patched onto an already-instantiated HF FalconH1MLP via _patch_swiglu_module,
+        # only `forward` is rebound — read multipliers from the instance if present, else config.
+        gate_multiplier = getattr(self, "gate_multiplier", None)
+        down_multiplier = getattr(self, "down_multiplier", None)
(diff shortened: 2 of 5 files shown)


shivam2199:
@puneeshkhanna Can you check if this resolves #936 

@Tcc0403 @vaibhavjindal @lowdy1 Please review it and let me know if any specific validations are needed.

src/liger_kernel/transformers/swiglu.py
```diff
@@ -135,3 +135,47 @@ def __init__(self, config, layer_idx=None, is_shared_mlp=False):
...
+        self.gate_multiplier = float(gate_multiplier)
+        self.down_multiplier = float(down_multiplier)
+
+    def forward(self, x):
+        # When patched onto an already-instantiated HF FalconH1MLP via _patch_swiglu_module,
+        # only `forward` is rebound — read multipliers from the instance if present, else config.
+        gate_multiplier = getattr(self, "gate_multiplier", None)
+        down_multiplier = getattr(self, "down_multiplier", None)
+        if gate_multiplier is None or down_multiplier is None:
+            gate_multiplier, down_multiplier = self.config.mlp_multipliers
```
Tcc0403: Wouldn't the instantiated MLP also hold these attributes?

src/liger_kernel/transformers/monkey_patch.py
```diff
@@ -2685,7 +2688,7 @@ def apply_liger_kernel_to_falcon_h1(
 
         for decoder_layer in base_model.layers:
             if swiglu:
-                _patch_swiglu_module(decoder_layer.mlp, LigerSwiGLUMLP)
+                _patch_swiglu_module(decoder_layer.feed_forward, LigerFalconH1SwiGLUMLP)
```
Tcc0403, line 2837: Did FalconH1 have breaking changes on attribute names? We still need hf transformers v4 support.

test/transformers/test_swiglu.py
```diff
@@ -416,6 +417,180 @@ def test_correctness_functional(bsz, seq_len, size, dtype, atol, rtol):
...
+    y_liger = LigerSiLUMulFunction.apply(a2, b2, gate_multiplier, down_multiplier)
+
+    assert torch.allclose(y_ref, y_liger, atol=atol, rtol=rtol)
+
+    grad = torch.randn_like(y_ref)
+    y_ref.backward(grad.clone())
+    y_liger.backward(grad.clone())
+
+    assert torch.allclose(a1.grad, a2.grad, atol=atol, rtol=rtol)
+    assert torch.allclose(b1.grad, b2.grad, atol=atol, rtol=rtol)
```
Tcc0403: Use `torch.testing.assert_close` for consistency

test/transformers/test_swiglu.py
```diff
@@ -416,6 +417,180 @@ def test_correctness_functional(bsz, seq_len, size, dtype, atol, rtol):
...
+    assert torch.allclose(y1, y2, atol=atol, rtol=rtol)
+
+    dy = torch.randn_like(y1)
+    y1.backward(dy.clone(), retain_graph=True)
+    y2.backward(dy.clone(), retain_graph=True)
+
+    assert torch.allclose(ref_mlp.gate_proj.weight.grad, liger_mlp.gate_proj.weight.grad, atol=atol, rtol=rtol)
+    assert torch.allclose(ref_mlp.up_proj.weight.grad, liger_mlp.up_proj.weight.grad, atol=atol, rtol=rtol)
+    assert torch.allclose(ref_mlp.down_proj.weight.grad, liger_mlp.down_proj.weight.grad, atol=atol, rtol=rtol)
+    assert torch.allclose(x1.grad, x2.grad, atol=atol, rtol=rtol)
```
Tcc0403: Ditto

Tcc0403 (commented):
Thanks for your contribution. I left some comments about compatibility issue and test suite convention.

src/liger_kernel/transformers/monkey_patch.py
shivam2199: Checked across v4.53.0 (first Falcon H1 release) → v4.57.6 → v5.x — no rename. FalconH1DecoderLayer has stored its MLP as self.feed_forward since landing. The original decoder_layer.mlp reference in the monkey-patch was latent dead code: it only executed under swiglu=True, which was False by default and gated with a "not available" warning, so the attribute error never fired. Flipping the default to True is what exposed it. No version guard needed.

src/liger_kernel/transformers/swiglu.py
shivam2199: You're right — I was being overly defensive. HF's FalconH1MLP.__init__ always sets self.gate_multiplier, self.down_multiplier = config.mlp_multipliers (since v4.53 when Falcon H1 landed), so whether we construct the module ourselves or patch onto an already-instantiated HF module via _patch_swiglu_module, the attributes are guaranteed present. Dropped the fallback and the comment.

test/transformers/test_swiglu.py
shivam2199: switched all torch.allclose asserts in the new tests to torch.testing.assert_close. Thanks for catching the inconsistency.

test/transformers/test_swiglu.py
shivam2199: switched all torch.allclose asserts in the new tests to torch.testing.assert_close. Thanks for catching the inconsistency.

shivam2199:
> 

@Tcc0403 Thanks for the review and addressed review feedback in commit 2a1b801:

- Simplified LigerFalconH1SwiGLUMLP.forward — HF's FalconH1MLP.__init__ always sets the multipliers on self, so the getattr fallback was dead defensive code.
- Checked Falcon H1 transformers compat: no attribute rename across v4.53 → v5.x, so decoder_layer.feed_forward works for all supported HF versions.
- Switched new tests to torch.testing.assert_close for consistency with the rest of the suite.
- Re-ran the 21 new tests on A10G — all pass. Ready for another look when you have a chance.

Tcc0403 (approved):
LGTM, @Mecoli1219 @vaibhavjindal for testing

shivam2199:
@Mecoli1219 I see that the test cases are getting queued but not getting executed - can we take a look there? It would be great if you could rerun the tests.

Mecoli1219 (approved):
Tests passed on H100. Got some unrelated error in convergence tests. We'll need to fix it later.

PR #1203, opened 2026-04-27
[Gemma 4] Add multimodal support (apply_liger_kernel_to_gemma4 for Gemma4ForConditionalGeneration)

dvdimitrov13:
## Summary

Follow-up to #1196 (the text path) — adds `apply_liger_kernel_to_gemma4` for `Gemma4ForConditionalGeneration` (multimodal class; includes E2B / E4B / E4B-it which are loaded by `AutoModelForCausalLM` as the multimodal class even when only text is being trained).

Closes the multimodal half of #1186.

### Why

Gemma 4's text vocab is 262,144. Without FLCE the (B, T, V) bf16 logits tensor is ~17 GB at T=8192 (and ~34 GB once the loss path upcasts to fp32 for cross-entropy), which OOMs even 96 GB cards on `Gemma4ForConditionalGeneration` SFT — the OOM that originally motivated #1186. Routing loss through `LigerForCausalLMLoss` materializes only the loss scalar.

### Shape

A single unified entry point that dispatches on class, per @Mecoli1219's preference in #1186:

- `apply_liger_kernel_to_gemma4(model=Gemma4ForConditionalGeneration_instance)` — multimodal path. Class-level RMSNorm + GeGLU swaps via `apply_liger_kernel_to_gemma4_text`, FLCE forward via the new `multimodal_forward`, recurses into `model.model.language_model` for instance-level patches.
- `apply_liger_kernel_to_gemma4(model=Gemma4ForCausalLM_instance)` — routes to `apply_liger_kernel_to_gemma4_text` for backwards compatibility, so the same entry point works for either shape.
- Registry: adds `"gemma4": apply_liger_kernel_to_gemma4` alongside the existing `"gemma4_text"`.

### Drive-by fixes

Two `isinstance(model, tuple_with_None_filter)` sites (one in our new dispatcher, one in the existing `apply_liger_kernel_to_gemma4_text`) raised `TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union` when called under `with patch("transformers.models.gemma4.modeling_gemma4")`: `getattr(MagicMock_module, "Gemma4TextForCausalLM", None)` returns a MagicMock (not `None`) because `MagicMock` auto-creates attributes, so the `cls is not None` filter let it slip into the `isinstance` tuple. The text-path version was dormant — its existing test passes a `Gemma4ForCausalLM` which short-circuits the `isinstance` match before reaching the bad entry — but the multimodal recursive call into the text path passes a `Gemma4TextModel`, so no early match. Both sites now use `isinstance(cls, type)` as the filter.

### Out of scope (deferred)

- **Vision / audio tower kernel swaps.** Gemma 4 loads both via `AutoModel.from_config(config.{vision,audio}_config)`, so the module classes are polymorphic. Out of scope here; FLCE on the LM head is what unblocks training OOM.
- **`Gemma4MultimodalEmbedder` / projector norms.** Analogous to gemma3's `mm_soft_emb_norm` patching. Skipped for the same minimal-surface reason.
- **PLE (Per-Layer Embeddings).** State passes through the inner forward unchanged; verified end-to-end on `google/gemma-4-E4B-it`.
- **MoE experts (`Gemma4TextExperts`).** Guarded out via the same `enable_moe_block` check used by the text path in #1196.
- **Multimodal mini convergence test (`mini_gemma4`).** Would need new `test/resources/fake_configs/Google/Gemma4/...` scaffolding for an image / audio processor — PR #1196 followed the same pattern (only added `mini_gemma4_text` to the non-multimodal convergence files). Happy to add it as a follow-up if you'd prefer it bundled here — let me know.

### Hardware

- **Hardware Type: RTX 5090 (Blackwell, sm_120, 32 GB), Vast.ai instance**
- CUDA 13.0, NVIDIA driver 590.48.01
- Python 3.10.12, torch 2.11.0+cu130, transformers 5.7.0.dev0 (built from `huggingface/transformers@main` — gemma4 requires ≥ 5.5.0)
- bf16 throughout for end-to-end numerics

### End-to-end numerical equivalence on real `google/gemma-4-E4B-it`

Verified before authoring this PR with our internal `verify_patch_equivalence.py` (same shape as #1196's verification):

| Axis | Threshold | Measured |
| --- | --- | --- |
| Fused-CE vs reference loss diff (seq=512) | `< 5e-3` | `0.0016` |
| End-to-end top-1 token agreement (seq=256, patched vs stock SDPA) | `> 99 %` | **>99 %** |
| End-to-end loss diff (patched vs stock SDPA) | `< 5e-3` | `0.0016` |
| Logit-distribution KL (informational) | — | `~3.5e-2` |
| SDPA EFFICIENT_ATTENTION vs MATH cos-sim (seq=4096) | `> 0.9999` | **>0.9999** |

### Liger-Kernel test gates

- [x] `make checkstyle` — `All checks passed!`, 267 files already formatted
- [x] `make test` — see log
- [x] `make test-convergence` — see logs (per file)

#### `make test`

`3131 passed, 903 skipped, 12 xfailed, 3 failed` in 35:42.

Our two new unit tests pass cleanly:

```
PASSED  test/transformers/test_monkey_patch.py::test_apply_liger_kernel_to_instance_for_gemma4_text                       (0.70s)
PASSED  test/transformers/test_monkey_patch.py::test_apply_liger_kernel_to_instance_for_gemma4_conditional_generation     (0.22s)
```

All six `LigerGEGLUMLPForGemma4` edge-case tests added by #1196 also pass.

The 3 unrelated failures are pre-existing on the parent branch (untouched by this PR) and reproduce on `main`:

```
FAILED  test/transformers/test_grpo_loss.py::test_grpo_loss_with_bias_correction_kl[...]
FAILED  test/transformers/test_monkey_patch.py::test_apply_liger_kernel_to_instance_for_paligemma
FAILED  test/transformers/test_monkey_patch.py::test_apply_liger_kernel_to_instance_for_gemma3_conditional_generation
```

#### `make test-convergence` (per file)

| File | Result |
| --- | --- |
| `fp32/test_mini_models.py` | 32 passed, 3 skipped |
| `fp32/test_mini_models_multimodal.py` | 8 passed, 4 skipped, 1 xfailed, **1 failed** (`mini_qwen2_vl`) |
| `fp32/test_mini_models_with_logits.py` | 29 passed, 3 skipped, 1 xpassed |
| `bf16/test_mini_models.py` | 33 passed, **2 failed** (`mini_llama4`, `mini_gemma4_text`) |
| `bf16/test_mini_models_multimodal.py` | 11 passed, 1 skipped, **2 failed** (`mini_qwen2_vl`, `mini_llama4`) |
| `bf16/test_mini_models_with_logits.py` | 30 passed, 1 skipped, **2 failed** (`mini_llama4`, `mini_qwen3_moe`) |

`mini_gemma4_text` passes in `bf16/test_mini_models_with_logits.py` and `fp32/test_mini_models.py` (PR #1196's text path). It fails only in `bf16/test_mini_models.py` with the same Blackwell bf16 logprob drift @eqy and reviewers flagged on #1196 (review comment `r4321013177`); this is in PR #1196's territory, not introduced by this multimodal patch. The other failures (`mini_llama4`, `mini_qwen2_vl`, `mini_qwen3_moe`) are also in models we don't touch and are pre-existing test instability on consumer Blackwell.

cc @Mecoli1219 @lardinator @ruilin-gif

---

🤖 Drafted with [Claude Code](https://claude.com/claude-code) (Claude Opus 4.7), reviewed and posted by me.

### DIFF
--- a/src/liger_kernel/transformers/monkey_patch.py
+++ b/src/liger_kernel/transformers/monkey_patch.py
@@ -1392,6 +1392,7 @@ def apply_liger_kernel_to_gemma4(
     rope: bool = False,
     cross_entropy: bool = False,
     fused_linear_cross_entropy: bool = True,
+    layer_norm: bool = False,
     rms_norm: bool = True,
     geglu: bool = True,
     model: PreTrainedModel = None,
--- a/src/liger_kernel/transformers/monkey_patch.py
+++ b/src/liger_kernel/transformers/monkey_patch.py
@@ -1355,7 +1355,11 @@ def apply_liger_kernel_to_gemma4_text(
         # The model instance already exists, so we need to additionally patch the
         # instance variables that reference already-instantiated modules
 
-        causal_lm_types = tuple(cls for cls in (Gemma4ForCausalLM, Gemma4TextForCausalLM) if cls is not None)
+        # `isinstance(cls, type)` filter (rather than `cls is not None`) so we
+        # also drop the MagicMock the test harness substitutes for
+        # `Gemma4TextForCausalLM` under `with patch("...modeling_gemma4")` —
+        # an `isinstance` check against a non-class entry raises TypeError.
+        causal_lm_types = tuple(cls for cls in (Gemma4ForCausalLM, Gemma4TextForCausalLM) if isinstance(cls, type))
         if isinstance(model, causal_lm_types) or isinstance(model, Gemma4TextModel):
             # get the base model from the model instance
             base_model = model.model if isinstance(model, causal_lm_types) else model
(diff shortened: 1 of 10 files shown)


Mecoli1219:
Hi @dvdimitrov13 . Do you have any update on this pr?

dvdimitrov13:
Apologies for the silence @Mecoli1219 - relatively new to open source contribution, and I misread the draft convention as "waiting for maintainer pre-review before final polish", which I now realise isn't how it works on external repos. Rebased onto current main as a single squashed commit (the previous 31 commits had drifted off the merge base from before #1196 landed), conflicts resolved, marked ready for review.

test/transformers/test_monkey_patch.py
```diff
@@ -1957,6 +1957,89 @@ def test_apply_liger_kernel_to_instance_for_gemma4_text():
...
+            num_kv_shared_layers=0,
+            use_double_wide_mlp=False,
+            enable_moe_block=False,
+            hidden_size_per_layer_input=0,
+        )
+        # Vision/audio configs left as None — Gemma4Model wraps both towers in
+        # `if config.<m>_config is not None`, so a None-towers model still
+        # constructs as Gemma4ForConditionalGeneration and exercises the
+        # multimodal forward we're patching. The towers themselves are
+        # polymorphic (AutoModel.from_config) and not in this PR's scope.
```
Mecoli1219, line 1988: Gemma4 ships dedicated [Gemma4VisionModel](https://huggingface.co/docs/transformers/model_doc/gemma4#transformers.Gemma4VisionModel) & [Gemma4AudioModel](https://huggingface.co/docs/transformers/model_doc/gemma4#transformers.Gemma4VisionModel) classes (concrete, not polymorphic like gemma3's `SigLIP` tower). RMSNorm and the GeGLU/SwiGLU-style MLPs in those towers should be a near-direct port from the text patches — worth doing in a follow-up.

test/utils.py
```diff
@@ -505,6 +505,21 @@ def revert_liger_kernel_to_gemma4_text(model_config: MiniModelConfig):
     print("Liger kernel patches have been reverted.")
 
 
+def revert_liger_kernel_to_gemma4(model_config: MiniModelConfig):
```
Mecoli1219, line 508: Can we have a convergence test for Gemma4 multimodal model?

src/liger_kernel/transformers/monkey_patch.py
```diff
@@ -1384,6 +1388,130 @@ def _maybe_patch_scaled_norm(module):
...
+    # path so this entry point works as a single user-facing API.
+    if model is not None:
+        # `isinstance(cls, type)` filter (rather than `cls is not None`) so we
+        # also drop the MagicMock the test harness substitutes for
+        # `Gemma4TextForCausalLM` under `with patch("...modeling_gemma4")` —
+        # an `isinstance` check against a non-class entry raises TypeError.
+        text_classes = tuple(
+            cls for cls in (Gemma4ForCausalLM, Gemma4TextForCausalLM, Gemma4TextModel) if isinstance(cls, type)
+        )
+        if isinstance(model, text_classes):
```
Mecoli1219: Will this path ever happen with `_apply_liger_kernel`? If not, we can throw error like how gemma3 did.

Mecoli1219 (commented):
Thanks @dvdimitrov13, and welcome to the open-source community! Nice first contribution!

Left some inline comments. One bigger-picture thought: we should plan to support Gemma4's vision and audio towers (Gemma4VisionModel / Gemma4AudioModel) in a follow-up. Happy to scope that as a separate PR. Just wanted to flag it so we don't lose track.

src/liger_kernel/transformers/monkey_patch.py
dvdimitrov13: Confirmed - the routing branch is dead via `_apply_liger_kernel`. Text variants dispatch through the `"gemma4_text"` registry entry directly, never reaching `"gemma4"`. Matched gemma3's pattern in d0f7516: `TypeError` on a non-`ConditionalGeneration` instance.

The `isinstance(cls, type)` filter in `apply_liger_kernel_to_gemma4_text` stays - the recursive call from the multimodal path still hits it with a `Gemma4TextModel` under `unittest.mock.patch`.

test/utils.py
dvdimitrov13: Added in f23039f. Wanted to be transparent about what came up during validation.

**What's in the commit**

- New `test/resources/fake_configs/Google/Gemma4/gemma-4-e4b-it/tokenizer_config.json`. The chat template emits `<image_soft_token>` (not `<start_of_image>` like gemma3): `Gemma4Processor.__call__` expands `image_token` placeholders to `<boi><image_token>*n<eoi>`, a different pattern from gemma3's boi-token replacement.
- `mini_gemma4` entries in bf16 and fp32 `test_mini_models_multimodal.py`. Vision config has `patch_size=16` (matching `Gemma4ImageProcessor`'s hardcoded default - the processor doesn't accept a patch_size kwarg). `audio_config=None`, so audio coverage stays in the vision/audio tower follow-up.
- `apply_liger_kernel_to_gemma4` now accepts `layer_norm: bool = False` as a no-op kwarg. The convergence framework passes `layer_norm=True` by default for any model not in its exclusion list, and Gemma4 vision uses RMSNorm. Accept-and-no-op is consistent with the deferred vision/audio tower scope.

**Validation, asking for guidance**

I spun up an RTX 3090 (Ampere) Vast.ai instance to validate locally. After fixing the items above, the test runs end-to-end, but the numerical comparison fails. So does the baseline `mini_gemma3` in both bf16 and fp32 on the same env. My read is this is the same env-sensitivity the original PR ran into with `mini_gemma4_text` on Blackwell bf16.

Env I used: PyTorch 2.5.1 (cuda12.4 image), transformers 5.7.0 from PyPI. I couldn't pull transformers main because it imports `CPUOffloadPolicy` from `torch.distributed.fsdp` which requires torch >= 2.6.

A couple of paths forward:

1. Land as-is and let your CI environment validate (assuming it's the configuration where `mini_gemma3` passes today).
2. Happy to retry on a different GPU + torch + transformers combination if you have a known-green spec.

I'd appreciate your call.

test/transformers/test_monkey_patch.py
dvdimitrov13: Will open a follow-up PR after this lands. Scope as I see it: RMSNorm + GeGLU/SwiGLU on Gemma4VisionModel and Gemma4AudioModel, the multimodal projector norms (analogous to gemma3's `mm_soft_emb_norm`), and the audio convergence test scaffolding I deferred from this PR.

Happy to open a tracking issue first if you'd prefer that over a draft PR.

Mecoli1219 (approved):
Hi @dvdimitrov13, I tested locally on H100. There are some numerical drifts that cause failures in the convergence test, but they're within the range we see for other multimodal models in bf16. I think it's fine to merge as-is.

Thanks again for the contribution! If you'd like to take on the vision/audio tower follow-up, feel free to open a PR and ping me when it's ready. Looking forward to your future contributions!

Issue #1197, opened 2026-04-21
[RFC, NPU] Liger-Kernel x NPU Kernel Contribution Skill


zheliuyu:
## Motivation

I'm not sure which project to submit this SKILL document to at the moment. So I'm creating this issue for now to communicate with the community.

I'll create a PR later showing a working example of migrating an NPU kernel with this SKILL.

**The following is the specific content of the SKILL document.**

```
name: liger-kernel-npu-kernel
description: >-
  Migrates and optimizes Liger-Kernel operators for NPU with
  API-compatible symbol replacement, UB-capacity-aware tiling, vector-core-
  aligned launch, grid-stride scheduling, 2D block memory access, precision-
  path control, on-device fused dataflow, and parity/performance validation.
```

# Liger-Kernel x NPU Kernel Contribution Skill

## 1. Scope and Objective

This skill is for:

- Porting `src/liger_kernel/ops/<op>.py` to `src/liger_kernel/ops/backends/_ascend/ops/<op>.py`.
- Applying NPU-focused optimizations, not just making code run.
- Maintaining the NPU export and replacement path so upper layer imports resolve to NPU implementations.
- Delivering reproducible numerical and performance evidence for ongoing migration work.

A completed task must satisfy all of the following:

1. API compatibility.
2. NPU-specific optimization landed.
3. Forward and backward parity.
4. Measured performance result or clear bottleneck analysis.

---

### 2.1 Key Paths

- Default operators: `src/liger_kernel/ops/`
- NPU operators: `src/liger_kernel/ops/backends/_ascend/ops/`
- Vendor registration: `src/liger_kernel/ops/backends/_ascend/__init__.py`
- Vendor registry: `src/liger_kernel/ops/backends/registry.py`
- Runtime replacement entry: `src/liger_kernel/ops/__init__.py`

### 2.2 Runtime Replacement Steps

1. `liger_kernel.ops` loads and calls `_replace_with_vendor_ops()`.
2. `infer_device()` resolves device type (`npu`).
3. `get_vendor_for_device("npu")` maps to vendor `ascend`.
4. `liger_kernel.ops.backends._ascend.ops` is imported dynamically.
5. Only symbols listed in the NPU module `__all__` override same-name symbols in `liger_kernel.ops`.

### 2.3 Import Semantics (Always Verify)

- `from liger_kernel.ops import ...`: vendor replacement applies.
- `from liger_kernel.ops.<module> import ...`: usually stays on default implementation.

[...]

PR #1206, opened 2026-04-28
[NPU] optimize cross_entropy for ASCEND NPU

sunyi0505:
# Summary
Refactors Ascend `cross_entropy` from a single fused forward+backward kernel to dedicated forward and backward Triton kernels, with device-side reduction stats and an optimized no-weight backward path.

# Motivation
- Avoid overwriting logits with gradients in forward and remove `debug_barrier` coupling between backward-style writes and loss computation.
- Reduce host sync from `.item()` before kernel launch by passing precomputed inverse scales via a small `ce_stats` buffer on device.
- Improve locality by assigning contiguous row ranges per program instead of interleaved grid-stride scheduling.

# Key changes
- **Split kernels**: `liger_cross_entropy_forward_kernel` + `liger_cross_entropy_backward_kernel` / `liger_cross_entropy_backward_kernel_no_weight`.
- **`ce_stats`**: `[inv_n_scale, inv_sum_weight_scale, weight_sum]` built without `.item()` for launch-friendly execution on NPU.
- **Optional LSE skip**: For fp32 logits and the simple CE path (no weight, smoothing, softcap, z-loss), skip a dedicated LSE tensor and derive LSE in the no-weight backward kernel from per-row loss + target logit where applicable.
- **Tuning**: Tiered block sizes / `get_no_weight_fast_path_block_size`; `eviction_policy="evict_first"` on logits loads; hoist `ls_eps` to host.
- **Autograd**: Save `ce_stats`, conditional `loss_1d` vs `lse`, `derive_lse_from_loss`; apply `grad_output` inside backward kernels.

# Testing
- Existing Ascend / transformers cross-entropy tests should cover numerical parity for enabled dtypes and feature flags.

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
@@ -7,137 +7,290 @@ import triton.language as tl
 from triton.language.math import tanh
 
 from liger_kernel.ops.backends._ascend.ub_manager import compute_default_tiling_strategy
-from liger_kernel.ops.utils import element_mul_kernel
 from liger_kernel.ops.utils import get_npu_core_count
 
 
 @triton.jit
-def liger_cross_entropy_kernel(
+def liger_cross_entropy_forward_kernel(
     X_ptr,
     X_stride,
     Y_ptr,
-    Y_stride,
     weight_ptr,
     loss_ptr,
     z_loss_ptr,
-    loss_stride,
+    lse_ptr,
     token_accuracy_ptr,
     token_accuracy_stride,
     predicted_tokens_ptr,
     predicted_tokens_stride,
     n_cols,
     n_rows,
-    n_non_ignore,
-    sum_non_ignore_weight,
-    weight_sum,
+    ce_stats_ptr,
     ignore_index,
+    ls_eps,
     lse_square_scale: tl.constexpr,
     label_smoothing: tl.constexpr,
-    reduction: tl.constexpr,  # set it as constexpr since reduction is always known at compile time
+    reduction: tl.constexpr,
     softcap,
     RETURN_Z_LOSS: tl.constexpr,
+    RETURN_LSE: tl.constexpr,
     RETURN_TOKEN_ACCURACY: tl.constexpr,
     RETURN_PREDICTED_TOKENS: tl.constexpr,
     BLOCK_SIZE: tl.constexpr,
     HAS_WEIGHT: tl.constexpr,
     HAS_SOFTCAPPING: tl.constexpr,
-    HAS_GRADIENTS: tl.constexpr,
 ):
     """
-    This kernel computes both cross entropy loss and the gradient of the input.
-    We only consider hard label + mean reduction for now. Please refer to https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html for the math.
+    Triton kernel: cross-entropy forward (per-row loss, optional z-loss, LSE, accuracy, argmax).
 
     Parameters:
-    X_ptr: Pointer to input tensor.
-    X_stride (int): The stride of the input tensor.
-    Y_ptr: Pointer to target tensor.
-    Y_stride (int): The stride of the target tensor.
-    weight_ptr: Pointer to weight tensor.
-    loss_ptr: Pointer to tensor to store the loss.
-    z_loss_ptr: Pointer to tensor to store the z loss. No operation if RETURN_Z_LOSS is 0.
-    loss_stride (int): The stride of the loss tensor.
-    token_accuracy_ptr: Pointer to tensor to store the per-token accuracy. No operation if RETURN_TOKEN_ACCURACY is 0.
-    token_accuracy_stride (int): The stride of the token accuracy tensor.
-    n_cols (int): The number of columns in the input tensor.
-    n_rows (int): The total number of rows to process.
-    n_non_ignore (float): The number of non-ignored elements in the batch.
-    sum_non_ignore_weight (float): The sum of non-ignored target's weights in the batch.
-    weight_sum (float): The sum of weight tensor.
-    ignore_index (int): The index to ignore in the target.
-    label_smoothing (float): The amount of smoothing when computing the loss, where 0.0 means no smoothing.
-    lse_square_scale (float): The scaler of (logsumexp(_input)) ^ 2 adding to the loss for the stability of training.
-    reduction (str): The string for the reduction to apply
-    softcap (float): The upper threshold for scaling logits to the range (-softcap, +softcap).
-    RETURN_Z_LOSS (int): The boolean value to decide whether to store z loss to z_loss_ptr or not. It must be 0 or 1.
-    RETURN_TOKEN_ACCURACY (int): The boolean value to decide whether to store per-token accuracy to token_accuracy_ptr or not. It must be 0 or 1.
-    BLOCK_SIZE (int): The block size for Triton operations.
-    HAS_WEIGHT (bool): The boolean value to determine whether assigning weight to each of the classes.
-    HAS_SOFTCAPPING (bool): The boolean value to determine whether applying soft-capping or not.
-    HAS_GRADIENTS (bool): The boolean value to determine whether calculating gradients in forward pass.
+        X_ptr: Logits pointer; logical shape ``(n_rows, n_cols)``, row-major with stride ``X_stride``.
+        X_stride (int64): Stride between consecutive rows of logits (columns contiguous).
+        Y_ptr: Target class indices per row; ignored rows compare equal to ``ignore_index``.
+        weight_ptr: Per-class weights ``[n_cols]`` when ``HAS_WEIGHT``; unused otherwise.
+        loss_ptr: Output per-row (unreduced) loss including optional z-loss term.
+        z_loss_ptr: Per-row z-loss ``lse_square_scale * lse**2`` when ``RETURN_Z_LOSS``; unused otherwise.
+        lse_ptr: Per-row log-sum-exp when ``RETURN_LSE``; caller may pass an unused tensor binding when false.
+        token_accuracy_ptr: Per-row float ``1.0`` if argmax matches target else ``0.0`` when ``RETURN_TOKEN_ACCURACY``.
+        token_accuracy_stride (int64): Row stride of ``token_accuracy_ptr`` (or ``0`` when disabled).
+        predicted_tokens_ptr: Per-row argmax index (``int64``) when ``RETURN_PREDICTED_TOKENS``; ``-1`` for ignored rows.
+        predicted_tokens_stride (int64): Row stride of ``predicted_tokens_ptr`` (or ``0`` when disabled).
+        n_cols (int): Vocabulary size (number of classes).
+        n_rows (int): Number of rows (batch * sequence length).
+        ce_stats_ptr: Float32 vector ``[inv_n_scale, inv_sum_weight_scale, weight_sum]`` for mean reduction and smoothing.
+        ignore_index (int): Label value to skip (no loss contribution; optional outputs zeroed or ``-1``).
+        ls_eps (float): ``label_smoothing / n_cols`` used in the weighted smoothing path when ``label_smoothing > 0``.
+        lse_square_scale (constexpr float): Coefficient on ``lse**2`` added into per-row loss (z-loss stabilizer).
+        label_smoothing (constexpr float): Label smoothing amount in ``[0, 1)``.
+        reduction (constexpr str): ``"mean"`` scales by stats from ``ce_stats_ptr``; ``"sum"`` / ``"none"`` use scale ``1``.
+        softcap (float): If ``HAS_SOFTCAPPING``, logits are ``softcap * tanh(x / softcap)``.
+        RETURN_Z_LOSS (constexpr bool): Write z-loss rows to ``z_loss_ptr``.
+        RETURN_LSE (constexpr bool): Write per-row LSE to ``lse_ptr``.
+        RETURN_TOKEN_ACCURACY (constexpr bool): Write per-row correctness to ``token_accuracy_ptr``.
+        RETURN_PREDICTED_TOKENS (constexpr bool): Write per-row argmax to ``predicted_tokens_ptr``.
+        BLOCK_SIZE (constexpr int): Block size along the vocabulary dimension.
+        HAS_WEIGHT (constexpr bool): Load ``weight_ptr[target]`` and per-class weights for smoothing when true.
+        HAS_SOFTCAPPING (constexpr bool): Apply tanh soft-capping with ``softcap`` when true.
+
+    Returns:
+        None. Effects are writes to ``loss_ptr`` and optional output pointers listed above.
     """
 
-    # Grid-Stride Loop: each program processes multiple rows
     pid = tl.program_id(0)
     num_progs = tl.num_programs(0)
 
-    start_row = pid
-    stride = num_progs
+    # Contiguous row ranges per program improve MTE3 store locality and reduce
+    # scalar address arithmetic versus interleaving (pid, pid+num_progs, ...).
+    row_chunk = (n_rows + num_progs - 1) // num_progs
+    row_start = pid * row_chunk
+    row_end = tl.minimum(row_start + row_chunk, n_rows)
+    n_local = row_end - row_start
 
-    for row_idx in range(start_row, n_rows, stride):
-        # https://github.com/triton-lang/triton/issues/1058
-        # If B*T*V is too large, program_id * stride will overflow out of int32, so we convert to int64
-        program_id = row_idx.to(tl.int64)
+    inv_n_scale = tl.load(ce_stats_ptr + 0)
+    inv_sum_weight_scale = tl.load(ce_stats_ptr + 1)
+    weight_sum = tl.load(ce_stats_ptr + 2)
 
-        # 1. Load Y_ptr first because if the target is ignore_index, we can return right away
-        Y_ptr_offset = program_id * Y_stride
-        y = tl.load(Y_ptr + Y_ptr_offset)
+    # One int64 row base per program; advance row pointers with small integer adds
+    # instead of row_idx.to(int64) * stride each iteration.
+    row_start_int64 = row_start.to(tl.int64)
+    target_row_ptr = Y_ptr + row_start_int64
+    logits_row_ptr = X_ptr + row_start_int64 * X_stride
+    loss_row_ptr = loss_ptr + row_start_int64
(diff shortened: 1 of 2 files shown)


sunyi0505:
1. **Graph Split: Forward / Backward Separation**  
   - **Old version**: A single `liger_cross_entropy_kernel` uses `HAS_GRADIENTS` to both write the loss and write gradients back in-place into `X_ptr` inside the same kernel, relying on `tl.debug_barrier()` to ensure read/write order.  
   - **New version**:  
     - `liger_cross_entropy_forward_kernel` only performs forward computation (loss, optional z-loss, accuracy, predicted tokens, optional LSE).  
     - `liger_cross_entropy_backward_kernel` and `liger_cross_entropy_backward_kernel_no_weight` are dedicated to computing gradients, writing them into a separate `dX` without overwriting the logits.  

2. **Grid & Memory Access: Contiguous Row Chunks vs Interleaved Stride**  
   - **Old version**: Uses `pid` interleaved traversal of rows with `stride num_progs` (grid-stride loop).  
   - **New version**: Uses contiguous row intervals by `row_chunk = ceil(n_rows / num_progs)`. The comments indicate this improves MTE memory locality on Ascend and reduces repeated address calculations like `row_idx * stride`. Inner loops move along rows using pointer arithmetic (`+1`).  

3. **Statistics `ce_stats`: Avoid `.item()` Before Launch**  
   - **Old version**: Passes `n_non_ignore`, `sum_non_ignore_weight`, `weight_sum` via `.item()` to the kernel, which easily introduces host synchronization on NPU.  
   - **New version**: `_make_ce_stats_buffer` constructs `[inv_n_scale, inv_sum_weight_scale, weight_sum]` on the device. The kernel loads `ce_stats_ptr + ...` via `tl.load()`, without synchronously fetching scalars from Python.  

4. **No-Weight Common Path: Dedicated Backward Kernel + Optional LSE Omission**  
   - The new version uses `backward_kernel_no_weight` when there is no weight / softcap / label smoothing / z-loss, and uses `get_no_weight_fast_path_block_size` (larger block size for small vocab) to optimize bandwidth.  
   - `_skip_lse_buffer_for_backward`: Under the simplified conditions above with fp32 logits, the forward pass can skip allocating and writing a separate fp32 LSE buffer. The backward pass restores LSE from `loss_row` and `x_y` via `DERIVE_LSE_FROM_LOSS`, reducing memory usage and forward stores (bf16/fp16 logits still retain LSE to avoid numerical issues and test mismatches).  

5. **Forward No Longer Uses Logits as Gradient Buffer**  
   - **Old version**: Ignored classes would zero out the entire row in `X` to achieve gradient semantics.  
   - **New version**: The forward pass only writes loss, etc. Gradients are only written to `grad_input` in the backward kernel, making the semantics clearer and facilitating the above LSE omission and fast path.  

6. **Constants & Minor Optimizations**  
   - `ls_eps`: `label_smoothing / V` is computed in Python and passed in; the kernel no longer divides by `n_cols` each row.  
   - `eviction_policy="evict_first"`: Added eviction policy for large-block logits loads (targeting Ascend).  
   - `get_optimal_block_size`: When no gradients are needed in the forward pass, uses segmented defaults based on vocab size (e.g., ≤32k → 1024, etc.), then falls back to UB tiling. The default fallback is increased from 2048 to 4096 (when no tiling is used).  

7. **Other Behavior / API**  
   - Token accuracy uses tensor operations like `(argmax_idx == y).to(tl.float32)`. Argmax across block merging is also adjusted.  
   - Removes dependency on `element_mul_kernel` in backward: scalar/vector `grad_output` is multiplied inside the backward kernel.  
   - Autograd saves `ce_stats`, `derive_lse_from_loss`, and saves `loss_1d` instead of LSE when needed.


foward:
image
backward:
image
full:
image
no_grad-forward:
image
ut:
image

sunyi0505:
@Tcc0403 This PR is ready for review.

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -7,324 +7,448 @@
...
-                    argmax_idx = tl.where(is_new_max, current_block_argmax_idx, argmax_idx)
+                    argmax_idx = argmax_idx + is_new_max.to(tl.int64) * (current_block_argmax_idx - argmax_idx)
 
                 if label_smoothing > 0:
-                    # scale X beforehand to avoid overflow
                     if HAS_WEIGHT:
-                        weight_block = tl.load(weight_ptr + X_offsets, mask=X_offsets < n_cols)
-                        scaled_x_sum += tl.sum(tl.where(X_offsets < n_cols, -eps * X_block * weight_block, 0.0))
+                        weight_block = tl.load(weight_ptr + X_offsets, mask=X_offsets < n_cols, other=0.0)
+                        scaled_x_sum += tl.sum(tl.where(X_offsets < n_cols, -ls_eps * X_block * weight_block, 0.0))
```
Tcc0403: Since out of bound weight is set to 0 already, we should be able to skip tl.where when doing tl.sum

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -7,324 +7,448 @@
...
+    Yp = Y_ptr + rs
+    Xrow = X_ptr + rs * X_stride
+    lossp = loss_ptr + rs
+    lsep = lse_ptr + rs
+    if RETURN_Z_LOSS:
+        zlossp = z_loss_ptr + rs
+    if RETURN_TOKEN_ACCURACY:
+        taccp = token_accuracy_ptr + rs * token_accuracy_stride
+    if RETURN_PREDICTED_TOKENS:
+        predp = predicted_tokens_ptr + rs * predicted_tokens_stride
```
Tcc0403: Keep variables full names instead of abbreviations and try using the original names if possible. It will be easier to maintain and have better readability.

For example, rs->row_start_int64, predp->predicted_token_ptrs, etc

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -7,324 +7,448 @@
...
+    Built without .item() to avoid NPU host sync before Triton launch.
+    """
+    device = target.device
+    dtype = torch.float32
+    if target_mask is None:
+        target_mask = target != ignore_index
+    sum_n = target_mask.sum(dtype=dtype)
+
+    if reduction == "mean":
+        inv_n = torch.ones((), dtype=dtype, device=device) / sum_n.clamp(min=1.0)
```
Tcc0403: https://docs.pytorch.org/docs/stable/generated/torch.Tensor.reciprocal.html#torch.Tensor.reciprocal
Use the inplace version of this instead?

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -7,324 +7,448 @@
...
+    else:
+        inv_n = torch.ones((), dtype=dtype, device=device)
+
+    if weight is not None:
+        non_ignore_targets = target.masked_select(target_mask)
+        sum_w = torch.gather(weight, dim=0, index=non_ignore_targets).sum(dtype=dtype)
+        w_sum = weight.sum(dtype=dtype)
+        if reduction == "mean":
+            inv_sum_w = torch.ones((), dtype=dtype, device=device) / sum_w.clamp(min=1e-12)
```
Tcc0403: Ditto

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -351,11 +476,13 @@ def cross_entropy_forward(
...
     n_rows = BT
 
-    BLOCK_SIZE = get_optimal_block_size(V, has_gradients=_input.requires_grad)
+    BLOCK_SIZE = get_optimal_block_size(V, has_gradients=False)
 
     # unreduced loss
     loss_1d = torch.zeros(n_rows, dtype=_input.dtype, device=_input.device)
     z_loss_1d = torch.zeros(n_rows, dtype=_input.dtype, device=_input.device) if return_z_loss else None
+    # When save_lse is False, forward does not write LSE; reuse loss_1d as the dummy lse_ptr only.
+    lse_1d = torch.empty(n_rows, dtype=torch.float32, device=_input.device) if save_lse else loss_1d
```
Tcc0403: Can we just pass None instead of dummy ptr?

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -412,21 +531,22 @@ def cross_entropy_forward(
         else 0,  # always 1 if predicted tokens is enabled
         n_cols=V,
         n_rows=n_rows,
-        n_non_ignore=n_non_ignore,
-        sum_non_ignore_weight=sum_non_ignore_weight,
+        weight_size0=weight.shape[0] if weight is not None else 0,
```
Tcc0403: Aren't weight.ndim=1 and weight.numel()=V always be True if weight is present? Why couldn't we reuse n_cols for boundary checking?

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -7,324 +7,448 @@
...
+    ce_stats_ptr,
     ignore_index,
+    ls_eps,
     lse_square_scale: tl.constexpr,
     label_smoothing: tl.constexpr,
-    reduction: tl.constexpr,  # set it as constexpr since reduction is always known at compile time
+    reduction: tl.constexpr,
     softcap,
     RETURN_Z_LOSS: tl.constexpr,
+    SAVE_LSE: tl.constexpr,
```
Tcc0403: Name it RETURN_LSE for consistency

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -7,324 +7,448 @@
...
     RETURN_TOKEN_ACCURACY: tl.constexpr,
     RETURN_PREDICTED_TOKENS: tl.constexpr,
     BLOCK_SIZE: tl.constexpr,
     HAS_WEIGHT: tl.constexpr,
+    WEIGHT_IS_FLOATING: tl.constexpr,
```
Tcc0403: What does it do?

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -7,324 +7,448 @@
...
-    label_smoothing (float): The amount of smoothing when computing the loss, where 0.0 means no smoothing.
-    lse_square_scale (float): The scaler of (logsumexp(_input)) ^ 2 adding to the loss for the stability of training.
-    reduction (str): The string for the reduction to apply
-    softcap (float): The upper threshold for scaling logits to the range (-softcap, +softcap).
-    RETURN_Z_LOSS (int): The boolean value to decide whether to store z loss to z_loss_ptr or not. It must be 0 or 1.
-    RETURN_TOKEN_ACCURACY (int): The boolean value to decide whether to store per-token accuracy to token_accuracy_ptr or not. It must be 0 or 1.
-    BLOCK_SIZE (int): The block size for Triton operations.
-    HAS_WEIGHT (bool): The boolean value to determine whether assigning weight to each of the classes.
-    HAS_SOFTCAPPING (bool): The boolean value to determine whether applying soft-capping or not.
-    HAS_GRADIENTS (bool): The boolean value to determine whether calculating gradients in forward pass.
```
Tcc0403, line 76: Keep it and add new arguments

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -7,324 +7,448 @@
...
+    dX_ptr,
+    dX_stride,
+    n_cols,
+    n_rows,
+    ce_stats_ptr,
+    ignore_index,
+    reduction: tl.constexpr,
+    BLOCK_SIZE: tl.constexpr,
+    HAS_GRAD_OUTPUT_VECTOR: tl.constexpr,
+    DERIVE_LSE_FROM_LOSS: tl.constexpr,
```
Tcc0403: HAS_LSE

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -7,324 +7,448 @@
...
+        ).cast(tl.float32)
+
+        final_scale = grad_scale * scale_factor
+
+        if y == ignore_index:
+            for i in range(0, n_cols, BLOCK_SIZE):
+                X_offsets = i + tl.arange(0, BLOCK_SIZE)
+                tl.store(dX_ptr + dX_ptr_offset + X_offsets, 0.0, mask=X_offsets < n_cols)
+        else:
+            if DERIVE_LSE_FROM_LOSS:
```
Tcc0403:
```
if not HAS_LSE:  # derive from loss if not present
```

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -7,324 +7,448 @@
...
+    grad_output_stride,
+    dX_ptr,
+    dX_stride,
+    n_cols,
+    n_rows,
+    ce_stats_ptr,
+    ignore_index,
+    reduction: tl.constexpr,
+    BLOCK_SIZE: tl.constexpr,
+    HAS_GRAD_OUTPUT_VECTOR: tl.constexpr,
```
Tcc0403: Isn't it equivalent to reduction=="none"?

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -7,324 +7,448 @@
...
+    ce_stats_ptr,
+    ignore_index,
+    lse_square_scale: tl.constexpr,
+    label_smoothing: tl.constexpr,
+    reduction: tl.constexpr,
+    softcap,
+    BLOCK_SIZE: tl.constexpr,
+    HAS_WEIGHT: tl.constexpr,
+    HAS_SOFTCAPPING: tl.constexpr,
+    HAS_GRAD_OUTPUT_VECTOR: tl.constexpr,
```
Tcc0403: Ditto

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -512,8 +703,19 @@ def forward(
         tuple: A tuple with the computed losses, accuracy, and predicted tokens: (loss, z_loss, token_accuracy, predicted_tokens). z_loss, token_accuracy, and predicted_tokens are None if not requested.
         """
         input_requires_grad = _input.requires_grad
-
-        loss, z_loss, token_accuracy, predicted_tokens, _input = cross_entropy_forward(
+        skip_lse_buf = _skip_lse_buffer_for_backward(weight, label_smoothing, softcap, lse_square_scale, _input.dtype)
+        save_lse_to_global = input_requires_grad and not skip_lse_buf
```
Tcc0403: Merge input_requires_grad condition into the helper function, rename `save_lse_to_global` to `return_lse`

src/liger_kernel/ops/backends/_ascend/ops/fused_linear_cross_entropy.py
Tcc0403, line 1: Mirror ce change request

zheliuyu:
The execution method for `benchmark_cross_entropy` has changed recently. Could you update the code and re-run the performance test?

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
sunyi0505: Triton requires a tensor pointer; when return_lse is False the kernel never reads/writes LSE (``RETURN_LSE`` false), so we pass ``loss_1d`` as the unused binding (Triton does not accept None pointers).

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
sunyi0505: HAS_GRAD_OUTPUT_VECTOR: When False, grad_output is a 0‑dimensional scalar (common with reduction="mean" or "sum").When True, grad_output is a 1‑dimensional vector (common with reduction="none", one gradient per row).
Inside the kernel, this determines which branch the code takes:
True: Read per row: grad_output[row_idx]
False: Read the same scalar: grad_output[0] (broadcast to all rows)

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
sunyi0505: Align with PyTorch weight convention, but it is not used inside the kernel, so remove it temporarily.

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
Tcc0403: Thanks for the explanation. If I understand correctly, we can reuse reduction for branching conditions. We can comment `grad_output is a vector if reduction!="none"` next to it.

sunyi0505: The branching here is not for choosing whether grad_output is a per-row vector or a scalar; instead, it selects which backward kernel to use.
has_grad_output_vector only indicates whether the incoming upstream grad_output is one scalar per token, or a single scalar for the entire loss. This is orthogonal to whether class weights, label smoothing, softcap, or z-loss (lse_square_scale) are enabled.
Only plain cross-entropy — with no class weights, no softcap, no label smoothing, and no z-loss term — is eligible for this optimized lightweight kernel. Once any of these features are enabled, the backward formulation changes entirely, and we must fall back to the general liger_cross_entropy_backward_kernel.
If we instead branch externally on reduction:
When reduction="mean" and class weights are present: has_grad_output_vector is False, causing an incorrect branch dispatch. Entering the no-weight kernel in this case produces wrong gradients.
When reduction="none" and weight / label smoothing are enabled: has_grad_output_vector is True. Routing solely based on this flag into the no-weight kernel also yields incorrect gradients.
Therefore, reduction cannot replace use_no_weight_fast_path: the former does not imply whether the backward pass can be implemented with plain cross-entropy logic.

sunyi0505: Thanks for your suggestions. I've already fixed it.

Tcc0403 (approved):
LGTM

zheliuyu:
The cross_entropy kernel optimization for shape 1024 is based on this SKILL document: https://github.com/linkedin/Liger-Kernel/issues/1197.

PR #1183, opened 2026-04-03
[NPU] Add NPU Fused MoE kernel

zheliuyu:
## Motivation

This pr ports `fused_moe.py` and `fused_moe_kernels.py` to an NPU-affine implementation while preserving the original math. The computational definition is unchanged: forward remains `W1 (gate/up) -> SwiGLU -> W2 -> token-weighted gather`, and backward still follows `dA' = dO @ W2^T` to produce `d_pre_act / dS / dW2 / dX / dW1`.
The main changes are execution-strategy optimizations for NPU.

## Note: Use the Skill

For this fused_moe kernel migration, we followed the skill document from https://github.com/linkedin/Liger-Kernel/issues/1197.

## Testing Done

- Hardware Type: Ascend 910B2
- [x] run `make test` to ensure correctness
- [x] run `make checkstyle` to ensure code style
- [x] run `make test-convergence` to ensure convergence

🤖 Generated with: [cursor](https://cursor.com/).

PR #1209, opened 2026-04-30
[fix] Distinguish the import path for compute_routing_metadata based on device type

zheliuyu:
## Summary

As Title. 

Fix: https://github.com/linkedin/Liger-Kernel/pull/1183#discussion_r3163137729 

- Hardware Type: npu


zheliuyu:
A temporary workaround is to distinguish the import path for `compute_routing_metadata` based on `infer_device`. Thanks for the heads-up — we'll follow up with a more robust solution for cases where different devices need to import different functions. Please review. @Mecoli1219

Mecoli1219 (approved):

Issue #1200, opened 2026-04-24
[Benchmark] Benchmark probe issue, memory probe and estimation may be inaccurate for some kernels


lowdy1:
## **Issue with Current Probe Design**

In the current design, `estimate_kernel_peak_memory` probes a kernel’s peak memory usage and returns `peak_bytes`:

```python
peak_bytes = estimate_kernel_peak_memory(probe_fn=_probe)
kernel_bpt = peak_bytes // num_tokens  # if needed
```

This derives a **bytes-per-token (BPT)** estimate by assuming memory scales linearly with the number of tokens.

---

## **Limitation**

This assumption only holds for kernels whose memory footprint scales **linearly** with `num_tokens`.

However, for some kernels Scaling may be **non-linear** (e.g., quadratic)

As a result, the computed `kernel_bpt` can be inaccurate or misleading for such kernels.

---

## **Summary**

The current BPT estimation is:

* ✅ Reasonable for linear-scaling kernels
* ❌ Unreliable for kernels with non-linear or structured memory behavior

Tcc0403:
What about a new argument `scaling_method` for the estimation function. Default to "linear".

shivam2199:
@Tcc0403  Took a look at this. The linear assumption actually appears in two places, not just `estimate_kernel_peak_memory`:

1. `estimate_kernel_peak_memory` returns raw `peak_bytes`; callers divide by `num_tokens` themselves.
2. `compute_seq_len_sweep_config` inverts via `max_tokens = usable_bytes / kernel_bytes_per_token` — only valid for linear scaling. `benchmark_sparse_multi_token_attention.py:334` already works around this by dividing by `probe_L * probe_L`, but the downstream sweep math still treats the result as linear bytes-per-token.

One direction worth considering: a new helper that internalizes both the probe and the inversion, leaving the existing functions alone.

```python
def compute_seq_len_sweep_config_with_probe(
    model_cfg: ModelConfig,
    probe_fn: Callable[[], torch.Tensor],
    probe_seq_len: int,
    probe_batch_size: int = 1,
    scaling_method: Literal["linear", "quadratic_in_L"] = "linear",
    memory_utilization: float = 0.4,
    max_seq_len: Optional[int] = None,
    max_batch_size: int = 32,
) -> SeqLenSweepConfig:
    ...
```

[...]

PR #1199, opened 2026-04-24
[Benckmark] Benchmark refactor example

lowdy1:
## Summary

The current benchmark scripts contain significant boilerplate when constructing `common_configs` for `run_benchmarks`. Although `compute_model_config_sweep_config` and `compute_seq_len_sweep_config` provide the core sweep logic, each script still:

* Reimplements probe logic
* Manually builds `extra_benchmark_config`
* Duplicates `common_configs` assembly
* Defines redundant helpers like `_resolve_*` and `*_model_config` variants

This PR removes that duplication by introducing higher-level builders that standardize how benchmarks are defined.

---

### 1. Introduce higher-level sweep builders

Add two unified helper functions in `benchmark_model_configs.py`:

```python
build_model_config_sweep(...)
build_token_length_sweep(...)
```

These functions:

* Wrap existing sweep utilities (`compute_*_sweep_config`)
* Internally handle memory probing via `setup_fn + forward_fn`
* Automatically construct `extra_benchmark_config` from:

  * `model_keys` (dynamic model attributes)
  * `extra_configs` (static overrides)
* Return a fully-formed `common_configs` dict

So benchmark scripts reduce to:

```python
common_configs = build_*(...)
run_benchmarks(**common_configs)
```

---

### 2. Standardize kernel definition via `setup_fn`

Instead of manually writing `probe_fn`, all kernels now define:

```python
setup_fn: SingleBenchmarkRunInput -> Tuple[Any, ...]
forward_fn: Tuple[Any, ...] -> torch.Tensor  (optional)
```

The builders handle:

```python
setup_out = setup_fn(input)
output = forward_fn(*setup_out)
```

A default is provided:

```python
forward_fn = lambda x, layer: layer(x)
```

This removes duplicated forward/probe logic across scripts.

---

### 3. Eliminate redundant helpers

The following patterns are removed across benchmark scripts:

* `probe_fn` definitions
* `extra_config_fn`
* `_resolve_model_config_*`
* `bench_*_model_config` variants

All are now handled centrally by the builders.

---

### `build_model_config_sweep`

* Sweeps across model configurations (x-axis = model name)
* Keeps total tokens (`B * T`) approximately constant
* Uses `setup_fn + forward_fn` to estimate memory per model

```python
build_model_config_sweep(
    kernel_name,
    all_model_configs,
    setup_fn,
    model_keys,
    forward_fn=...,
    probe_provider="torch",
    extra_configs=None,
    bt=2048,
    overwrite=False,
)
```

---

[...]

PR #1218, opened 2026-05-07
[Benchmark] Add compute_seq_len_sweep_config_with_probe with linear/quadratic scaling support

shivam2199:
## Summary

Refs #1200. Addresses non-linear memory scaling in benchmark sweep config inference.

The existing `compute_seq_len_sweep_config` inverts memory via `max_tokens = usable_bytes / kernel_bytes_per_token`, which only holds for linear-scaling kernels. For O(L²) kernels (e.g. `benchmark_sparse_multi_token_attention.py`), this overestimates capacity by orders of magnitude — the existing workaround there divides by `probe_L * probe_L`, but the downstream sweep math still treats the result as linear bytes-per-token.

Per discussion on the issue (https://github.com/linkedin/Liger-Kernel/issues/1200#issuecomment-4396956347), this PR adds a new helper rather than threading `scaling_method` through the existing function — 16+ benchmark scripts call `estimate_kernel_peak_memory` today, and a wider signature change would conflict with in-flight benchmark refactors (#1199, #1180). Linear-scaling callers are unchanged; only quadratic-scaling benchmarks opt in.

### What changed

- **`benchmark/scripts/benchmark_model_configs.py`** — adds `compute_seq_len_sweep_config_with_probe(model_cfg, probe_fn, probe_seq_len, probe_batch_size=1, scaling_method="linear" | "quadratic", ...)`. Internalizes the probe call + inversion; reuses `estimate_kernel_peak_memory` for the measurement.
- **`benchmark/scripts/benchmark_sparse_multi_token_attention.py`** — switches the `token_length` sweep mode to the new helper with `scaling_method="quadratic"`, dropping the manual `peak_bytes // (probe_L * probe_L)` workaround.

`estimate_kernel_peak_memory` and `compute_seq_len_sweep_config` are untouched.

## Validation

Hardware: A10G 24GB (g5.xlarge).

Synthetic O(L²) probe (B=2, L=2048, allocates `B * L * L` floats) using `LLAMA_3_8B` config and `max_seq_len=2**20` to bypass the model cap so the raw inversion is visible:

```
quadratic: SeqLenSweepConfig(batch_size=2, seq_len=8192)
linear:    SeqLenSweepConfig(batch_size=2, seq_len=65536)
```

The 8× gap (≈17× before snap-to-power-of-2) demonstrates the inversion difference: `linear` claims a sweep at L=65536 fits, when in reality L² at that size would require multiple TBs. `quadratic` lands at a realistic L=8192. This matches the issue's premise — for non-linear-scaling kernels, the existing inversion overestimates capacity and would OOM at the predicted boundary.

## Testing Done

- [x] Synthetic O(L²) sanity check on A10G — confirms `quadratic` predicts L=8192 vs `linear` predicts L=65536 for the same probe (8× separation, scales as expected).
- [x] `benchmark_sparse_multi_token_attention.py` imports + helper resolution verified locally.
- [ ] Full sparse-attention end-to-end sweep on A10G (deferred — synthetic test already isolates the inversion math from kernel-specific noise).

cc @Tcc0403

### DIFF
--- a/benchmark/scripts/benchmark_model_configs.py
+++ b/benchmark/scripts/benchmark_model_configs.py
@@ -291,6 +291,68 @@ def estimate_kernel_peak_memory(probe_fn: Callable[[], torch.Tensor]) -> int:
     return max(1, peak_bytes)
 
 
+def compute_seq_len_sweep_config_with_probe(
+    model_cfg: ModelConfig,
+    probe_fn: Callable[[], torch.Tensor],
+    probe_seq_len: int,
+    probe_batch_size: int = 1,
+    scaling_method: Literal["linear", "quadratic"] = "linear",
+    memory_utilization: float = 0.4,
+    max_seq_len: Optional[int] = None,
+    max_batch_size: int = 32,
+) -> SeqLenSweepConfig:
+    """Compute safe sweep config from a probe, supporting non-linear seq-len scaling.
+
+    Wraps :func:`estimate_kernel_peak_memory` and inverts the memory model
+    according to *scaling_method*. Linear-scaling kernels can keep using
+    :func:`compute_seq_len_sweep_config` directly; this helper exists for
+    kernels whose memory grows non-linearly with seq_len (e.g. attention
+    kernels with O(L^2) scratch).
+
+    Args:
+        model_cfg: Model architecture config.
+        probe_fn: Callable that performs setup, runs a forward pass, and
+            returns an output tensor suitable for ``.backward()``. Same
+            contract as :func:`estimate_kernel_peak_memory`'s *probe_fn*.
+        probe_seq_len: Sequence length used inside *probe_fn*. Required so
+            the inversion can isolate the seq-len-dependent term.
+        probe_batch_size: Batch size used inside *probe_fn*. Defaults to 1.
+        scaling_method: How peak memory scales with seq_len, holding batch
+            size fixed.
+
+            - ``"linear"``: peak ~ B * L. Inversion: ``L_max = usable / (B * c_per_BL)``.
+            - ``"quadratic"``: peak ~ B * L^2. Inversion: ``L_max = sqrt(usable / (B * c_per_BL2))``.
+        memory_utilization: Fraction of total device memory to target (0 to 1).
+        max_seq_len: Hard upper bound for sequence length. Defaults to
+            ``model_cfg.max_position_embeddings``.
+        max_batch_size: Hard upper bound for batch size.
+    """
+    if scaling_method not in ("linear", "quadratic"):
+        raise ValueError(f"scaling_method must be 'linear' or 'quadratic', got {scaling_method!r}")
+
+    peak_bytes = estimate_kernel_peak_memory(probe_fn=probe_fn)
+
+    total_memory_gb = get_total_gpu_memory()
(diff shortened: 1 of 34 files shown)


shivam2199:
@Tcc0403 @Mecoli1219 Please take a look

benchmark/scripts/benchmark_model_configs.py
```diff
@@ -291,6 +291,68 @@ def estimate_kernel_peak_memory(probe_fn: Callable[[], torch.Tensor]) -> int:
...
+
+    if scaling_method == "linear":
+        c_per_BL = max(1.0, peak_bytes / (probe_batch_size * probe_seq_len))
+        max_seq_len_from_mem = max(1, int(usable_bytes / (batch_size * c_per_BL)))
+    else:
+        c_per_BL2 = max(1.0, peak_bytes / (probe_batch_size * probe_seq_len * probe_seq_len))
+        max_seq_len_from_mem = max(1, int(math.sqrt(usable_bytes / (batch_size * c_per_BL2))))
+
+    seq_len = min(max_seq_len, max_seq_len_from_mem)
+    seq_len = 2 ** int(math.log2(seq_len)) if seq_len >= 1024 else 1024
```
Tcc0403: Is it possible to just plug this part to `compute_seq_len_sweep_config`?

shivam2199: @Tcc0403 Good call. Pushed 6c204db which extracts two private helpers — `_max_seqlen_under_memory` (handles both linear and quadratic inversion) and `_snap_pow2_seqlen` — and collapses both public functions to thin orchestration over them.

`compute_seq_len_sweep_config` treats `kernel_bytes_per_token` as a unit-probe (B=L=1, linear) so the inversion math reduces to the existing `max_tokens = usable / bpt` quantity. No behavior change for the existing callers; the duplicated inversion/snap logic is gone.

benchmark/scripts/benchmark_model_configs.py
```diff
@@ -291,6 +291,88 @@ def estimate_kernel_peak_memory(probe_fn: Callable[[], torch.Tensor]) -> int:
...
+    raise ValueError(f"scaling_method must be 'linear' or 'quadratic', got {scaling_method!r}")
+
+
+def _snap_pow2_seqlen(seq_len: int, max_seq_len: int) -> int:
+    """Clamp to *max_seq_len* and snap down to nearest power of 2 (floor at 1024)."""
+    seq_len = min(max_seq_len, seq_len)
+    return 2 ** int(math.log2(seq_len)) if seq_len >= 1024 else 1024
+
+
+def compute_seq_len_sweep_config_with_probe(
```
Tcc0403: we can replace all occurrences of `compute_seq_len_sweep_config` with yours, keeping only one helper function

shivam2199:
@Tcc0403 Pushed a8a8f40. Collapsed the two public helpers into one — `compute_seq_len_sweep_config` now takes `probe_fn` + `probe_seq_len` directly with optional `probe_batch_size` and `scaling_method`. Migrated all 33 callers (net −154 lines).

The `benchmark_multi_token_attention` script's manual `peak_bytes // (probe_L * probe_L)` inversion is now first-class via `scaling_method="quadratic"`. `make checkstyle` clean.

Tcc0403 (approved):
Thank you, lgtm

PR #1222, opened 2026-05-13
[NPU] optimize fused_linear_cross_entropy

sunyi0505:
## Summary

Move gradient computation out of forward, add a plain CE fast path, and optionally reuse saved logits to improve Ascend training performance while preserving correct grad_output handling.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
@@ -204,6 +204,64 @@ def liger_cross_entropy_forward_kernel(
 
 
 @triton.jit
+def liger_cross_entropy_forward_kernel_plain(
+    X_ptr,
+    X_stride,
+    Y_ptr,
+    loss_ptr,
+    n_cols,
+    n_rows,
+    ce_stats_ptr,
+    ignore_index,
+    reduction: tl.constexpr,
+    BLOCK_SIZE: tl.constexpr,
+):
+    """
+    Plain CE forward fast path (no weight/softcap/smoothing/z-loss/metrics).
+    1 program processes 1 row to maximize parallelism (unlike the generic kernel
+    which chunks rows per program for flexibility).
+    Writes per-row loss to ``loss_ptr`` (float32 recommended).
+    """
+    row = tl.program_id(0)
+    if row >= n_rows:
+        return
+
+    row_i64 = row.to(tl.int64)
+    y = tl.load(Y_ptr + row_i64)
+
+    # Pre-load mean scaling. For sum/none we treat scale as 1.0.
+    inv_n = tl.load(ce_stats_ptr + 0).cast(tl.float32)
+
+    if y == ignore_index:
+        tl.store(loss_ptr + row_i64, 0.0)
+        return
+    logits_row_ptr = X_ptr + row_i64 * X_stride
+    m = float("-inf")
+    d = 0.0
+    # Keep x[y] as a direct load outside the scan. Folding it into the block
+    # loop adds extra selection state and tends to increase UB/register pressure
+    # on Ascend.
+    x_y = tl.load(logits_row_ptr + y).cast(tl.float32)
+    for i in range(0, n_cols, BLOCK_SIZE):
+        offs = i + tl.arange(0, BLOCK_SIZE)
+        x = tl.load(
+            logits_row_ptr + offs,
+            mask=offs < n_cols,
+            other=float("-inf"),
+            eviction_policy="evict_first",
+        ).cast(tl.float32)
+        block_max = tl.max(x)
+        m_new = tl.maximum(m, block_max)
+        d = d * tl.exp(m - m_new) + tl.sum(tl.exp(x - m_new))
+        m = m_new
+    lse = m + tl.log(d)
+    loss = lse - x_y
+    if reduction == "mean":
+        loss = loss * inv_n
+    tl.store(loss_ptr + row_i64, loss)
+
+
+@triton.jit
 def liger_cross_entropy_backward_kernel_no_weight(
     X_ptr,
     X_stride,
--- a/src/liger_kernel/ops/backends/_ascend/ops/fused_linear_cross_entropy.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/fused_linear_cross_entropy.py
@@ -1,13 +1,16 @@
+import warnings
+
 import torch
 import triton
 
 from liger_kernel.ops.backends._ascend.ops.cross_entropy import _make_ce_stats_buffer
 from liger_kernel.ops.backends._ascend.ops.cross_entropy import liger_cross_entropy_backward_kernel
+from liger_kernel.ops.backends._ascend.ops.cross_entropy import liger_cross_entropy_backward_kernel_no_weight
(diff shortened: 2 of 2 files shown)


sunyi0505:
# Main Modifications

## Introduce plain fast path for Ascend CE:

Add a new kernel: liger_cross_entropy_forward_kernel_plain. A lightweight kernel is used for vanilla cross-entropy cases without ce_weight, softcap, label smoothing, z loss, token metrics, or token scaling.

## Remove upfront gradient computation in forward pass

The original implementation precomputed and cached grad_input / grad_weight / grad_bias during the forward pass. The new implementation only computes loss and essential statistics, and offloads all actual gradient calculation to LigerFusedLinearCrossEntropyFunction.backward().

## Reimplement backward pass with chunked computation

Reuse cached forward logits if available; otherwise recompute input @ weight.T in the backward pass. The CE backward kernel produces grad_logits, and GEMM is then applied to derive grad_input, grad_weight, and grad_bias.

## Add logits storage limit via _get_logits_save_limit_bytes()

Under the plain fast path, logits are cached when their size is within device memory budget to avoid redundant GEMM in the backward pass. The limit is defined as min(4GiB, total_memory / 8) with a hard minimum of 1GiB.

## Adjust token scaling logic

The legacy implementation gathered valid targets separately. The new logic adopts safe_targets combined with masking, and caches the complete scaling_factors_full for backward usage.

## Remove the element_mul_kernel path

The old backward logic multiplied precomputed forward gradients by the upstream grad_output. The new implementation consumes grad_output directly inside the CE backward kernel.

# Motivation

## The core objective is to optimize the performance and gradient semantics of fused linear cross entropy on Ascend.

## The legacy implementation placed heavy gradient computation into the forward pass, making the forward stage undertake backward workload. It also failed to fully utilize the actual grad_output passed from autograd, especially for reduction="none" scenarios where upstream gradients are vector-valued. By moving gradient computation back to the backward pass, CE gradients are computed strictly based on the real incoming grad_output, yielding cleaner and more consistent semantics.

## Furthermore, vanilla cross-entropy is a high-frequency path in language model training. The newly added plain fast path leverages larger block size, fewer kernel branches, and reduced Python/kernel launch overhead. Logits are reused within a controlled memory budget to lower backward recomputation cost. Overall, this change achieves an Ascend-optimized trade-off between memory footprint and GEMM / kernel launch efficiency.

forward:
image

backward:
image

full:
image

no_grad_forward:
image

ut:
image

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -203,6 +203,63 @@ def liger_cross_entropy_forward_kernel(
...
+
+    if y == ignore_index:
+        tl.store(loss_ptr + row_i64, 0.0)
+        return
+    logits_row_ptr = X_ptr + row_i64 * X_stride
+    m = float("-inf")
+    d = 0.0
+    # Directly load x[y]. Attempting to "extract" it from the scanned blocks
+    # tends to increase UB/register pressure on Ascend.
+    x_y = tl.load(logits_row_ptr + y).cast(tl.float32)
```
Tcc0403: Do you mean reading after for loop might hurt the performance?

src/liger_kernel/ops/backends/_ascend/ops/fused_linear_cross_entropy.py
```diff
@@ -409,10 +415,173 @@ def backward(ctx, grad_output, grad_output2, grad_output3, grad_output4):
...
+                    BLOCK_SIZE=backward_block_size,
+                    HAS_WEIGHT=True if ce_weight is not None else False,
+                    HAS_SOFTCAPPING=True if ctx.softcap is not None else False,
+                )
+
+            if ctx.use_token_scaling and scaling_factors_full is not None:
+                grad_logits_chunk = grad_logits_chunk * scaling_factors_full[start_idx:end_idx].unsqueeze(-1)
+
+            grad_input[start_idx:end_idx] = grad_logits_chunk @ weight
+            grad_logits_weight = grad_logits_chunk.t() @ input_chunk
```
Tcc0403: nit: `grad_logits_weight `-> `grad_weight_`

src/liger_kernel/ops/backends/_ascend/ops/fused_linear_cross_entropy.py
```diff
@@ -409,10 +415,173 @@ def backward(ctx, grad_output, grad_output2, grad_output3, grad_output4):
...
+            if chunk_id == 0:
+                grad_weight.copy_(grad_logits_weight)
+            else:
+                grad_weight.add_(grad_logits_weight)
+            if grad_bias is not None:
+                grad_logits_sum = grad_logits_chunk.sum(dim=0)
```
Tcc0403: nit: `grad_logits_sum` -> `grad_bias_`

src/liger_kernel/ops/backends/_ascend/ops/fused_linear_cross_entropy.py
```diff
@@ -409,10 +415,173 @@ def backward(ctx, grad_output, grad_output2, grad_output3, grad_output4):
...
             del grad_output4  # predicted_tokens is only for metrics
-        (grad_input, grad_weight, grad_bias) = ctx.saved_tensors
-        grad_input, grad_weight, grad_bias = fused_linear_cross_entropy_backward(
-            grad_output, grad_input, grad_weight, grad_bias
+        (_input, weight, target, loss_1d, ce_stats, saved_logits) = ctx.saved_tensors
+        bias = ctx.bias
+        ce_weight = ctx.ce_weight
+        scaling_factors_full = ctx.scaling_factors_full
+
+        device = _input.device
```
Tcc0403: put the following code in the original `fused_linear_cross_entropy_backward`?

src/liger_kernel/ops/backends/_ascend/ops/fused_linear_cross_entropy.py
```diff
@@ -72,48 +70,97 @@ def fused_linear_cross_entropy_forward(
...
+            loss,
+            None,
+            None,
+            None,
+            loss_1d,
+            ce_stats,
+            None,
+            plain_fast_path,
+            logits_for_backward,
+        )
```
Tcc0403, line 182: This fallback makes sense but shouldn't we inform user to use cross entropy instead of fused lienar cross entropy?

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
sunyi0505: Load x[y] directly outside the scan loop. Avoid embedding selection logic into the block loop to reduce undefined behavior risks and register pressure on Ascend, thereby improving performance.

src/liger_kernel/ops/backends/_ascend/ops/fused_linear_cross_entropy.py
sunyi0505: Added logging in fused_linear_cross_entropy_forward:if the caller already has logits available, it is recommended to use the standalone cross-entropy operator directly.The warning is triggered only once per process to avoid redundant alerts at every training step.

Tcc0403 (approved):
lgtm

Issue #1229, opened 2026-05-19
LigerORPOTrainer doesn't work when not using FSDP


Ola-Vish:
### 🐛 Describe the bug

Hi all,
I tried using LigerORPOTrainer but it doesn't seem to work when running on a single GPU.
I tried running the orpo trainer example but it fails on `assert isinstance(wrapper_module, FullyShardedDataParallel)` .
As far as I can tell from the code, this is because  _FSDPForwardRedirection is used without checking if the model is FullyShardedDataParallel, as can be seen [here](https://github.com/linkedin/Liger-Kernel/blame/30b8486a2bd48dff97f22c5f0c88520b8825cb35/src/liger_kernel/transformers/trainer/orpo_trainer.py#L77). 

Please let me know if I understood the issue correctly, perhaps I can submit a fix 😃 

### Reproduce

Link to reproduction code - [here](https://gist.github.com/Ola-Vish/d93567a8f4a9896fa80e70d9e2da3cec)

Tried running this code (which is basically the same as the this [example](https://github.com/linkedin/Liger-Kernel/blob/main/examples/alignment/run_orpo.py) code by running "python liger_orpo_trainer.py", and what I get is the following error:

` 0%|                                                                                            | 0/100 [00:00<?, ?it/s]Traceback (most recent call last):
  File "/home/ubuntu/training_code/liger_orpo_trainer.py", line 43, in <module>
    main()
  File "/home/ubuntu/training_code/liger_orpo_trainer.py", line 39, in main
    trainer.train()
  File "/home/ubuntu/training_code/sota_venv/lib/python3.12/site-packages/transformers/trainer.py", line 1424, in train
    return inner_training_loop(
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/training_code/sota_venv/lib/python3.12/site-packages/transformers/trainer.py", line 1506, in _inner_training_loop
    self._run_epoch(
  File "/home/ubuntu/training_code/sota_venv/lib/python3.12/site-packages/transformers/trainer.py", line 1734, in _run_epoch
    tr_loss_step = self.training_step(model, inputs, num_items_in_batch)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/training_code/sota_venv/lib/python3.12/site-packages/transformers/trainer.py", line 1906, in training_step
[...]

PR #1230, opened 2026-05-19
fix LigerORPOTrainer bug in none fsdp setup

Ola-Vish:
This is my suggested solution to the bug I reported in #1229.
With my change, the [code](https://gist.github.com/Ola-Vish/d93567a8f4a9896fa80e70d9e2da3cec) that used to fail now works.

- Hardware Type: NVIDIA A100-SXM4-40GB
- [] run `make test-convergence` to ensure convergence

image
image


feiiiiii5:
@Ola-Vish Following up here too. I commented on the parent issue #1229 with options to help push this forward. The fix direction looks correct (mirroring the L51 `isinstance(model, FullyShardedDataParallel)` guard at L77). The main gap blocking merge is the missing regression test — happy to collaborate on adding one.

Ola-Vish:
@feiiiiii5 I kinda lost track of this issue, good thing you reminded me of this 😃 

I couldn't for the life of me get the tests to pass, and it had nothing to do with my branch, so I gave up. I'll be very happy if you could contributed the "make test validation" as per your first suggestion, @kashif already kindly submitted a regression test.

feiiiiii5:
@Ola-Vish Great — glad the ping was useful, and thanks @kashif for the regression test! 🎉

Reading the current diff on `bugfix/orpo_trainer` (after @kashif's merge), I can see PR #1230 now bundles three things together:

1. **Your original non-FSDP guard fix** in `orpo_trainer.py` (mirroring the L51 isinstance check at L77)
2. **@kashif's regression test** `test/transformers/test_orpo_trainer.py` (`test_concatenated_forward_without_fsdp`)
3. **@kashif's trl-compat fallback** — `ORPOTrainer`/`ORPOConfig` import from `trl.experimental.orpo` with fallback to old `trl.trainer` path (closes #1244)

This is a solid bundle. Let me run `make test` validation locally to confirm everything passes — that's the part you mentioned got stuck. I'll pull your `bugfix/orpo_trainer` branch into a local checkout, install the test deps, and run the specific test (plus the broader `test/transformers/` suite to check for regressions). Will report back with the result + a CI-status summary so we can ping a maintainer for review with concrete evidence.

Give me ~15-20 minutes — I'll post the validation result here.

feiiiiii5:
@Ola-Vish Follow-up with the local `make test` validation result.

**TL;DR — I couldn't reproduce a clean local run on macOS, but I think I found why you were stuck, and it's not your branch.**

## What I tried locally

I fetched `bugfix/orpo_trainer` into a fresh checkout and attempted `make test` with:
- macOS arm64 (Apple M2 Pro, no NVIDIA/ROCm GPU)
- `torch==2.13.0`, `transformers==5.13.1`, `trl==1.8.0`, `pytest==8.4.2`
- `liger_kernel` installed in editable mode

## Where it broke (and likely where you were stuck too)

The regression test never got to run. Collection fails at:

```
src/liger_kernel/ops/attn_res.py:23: in <module>
    import triton
E   ModuleNotFoundError: No module named 'triton'
```

`triton` is **Linux-only** — `pip install triton` returns `ERROR: No matching distribution found for triton` on macOS. `liger_kernel.ops.__init__` eagerly imports every op module (including `attn_res.py`) at package load time, so the whole `liger_kernel` import chain dies before the trainer is ever reached.

I tried faking a minimal `triton` module via `sys.modules`, but `transformers==5.13.1` then goes one step further and tries `import triton.backends.compiler` when loading `PreTrainedModel`, which the fake module can't satisfy either.

**This is almost certainly the "tests won't pass and it had nothing to do with my branch" symptom you described.** The CI matrix (`amd-ci.yml` / `intel-ci.yml` / `nvidia-ci.yml`) runs on Linux runners where `triton` installs cleanly, so the same code path that's broken on a Mac laptop is green on CI.

## Code-level trace (since the local run is blocked)

I traced `concatenated_forward` on the current `bugfix/orpo_trainer` HEAD (commit `de9befd`):

- `src/liger_kernel/transformers/trainer/orpo_trainer.py:56-65` — the `isinstance(model, FullyShardedDataParallel)` guard now correctly branches:
  - FSDP path (L57-64): goes through `_FSDPForwardRedirection` against `model._fsdp_wrapped_module.model`
  - non-FSDP path (L65-73): unwraps `DataParallel` if needed and calls `model.model(...)` directly
- `src/liger_kernel/transformers/trainer/orpo_trainer.py:92-100` — same `isinstance(model, FSDP)` guard wraps the `orpo_loss` call site. The non-FSDP path falls through to a plain `orpo_partial(lm_head, last_hidden_state, ...)` call, which is what #1229 was missing.

The `trl.experimental.orpo` import on L14 with the `trl.trainer.ORPOTrainer` fallback on L16 looks right for the trl 0.x → 1.x transition.

## Regression test (test/transformers/test_orpo_trainer.py)

Statically reviewed the 60-line test:
- `pytest.importorskip("trl")` — correct guard
- Stubs `concatenated_inputs` via `lambda *a, **k: concatenated_batch` so the test only exercises `concatenated_forward` (no dataset plumbing)
- Small `LlamaForCausalLM` (`vocab=128`, `hidden=32`, `layers=2`) — fits CPU and never hits a triton kernel
- Asserts `loss.isfinite().all()` and `len(aux_outputs) > 0` — exactly what the original FSDP-only code path would crash on for a non-FSDP model

Logic looks sound. The test would have caught #1229.

## Suggested next step

The PR is currently `draft=True` with `0 reviews` and `no checks reported on the 'bugfix/orpo_trainer' branch`. Two options:

1. **Mark the PR ready-for-review** so maintainer CI gets a chance to run on Linux (where triton installs cleanly). If the regression test passes on CI, that's the green light we need.
2. If you'd rather have a local Linux validation, the cheapest path is a one-shot Docker run: `docker run --rm -v $PWD:/work -w /work python:3.10 bash -c "pip install -e .[dev] && pytest test/transformers/test_orpo_trainer.py -v"`. Happy to script and post the output if useful.

Want me to mark the PR ready-for-review, or would you prefer to keep it draft until you've had a chance to look yourself?

PR #1227, opened 2026-05-18
[NPU, A3] Add NPU kernel support for A3 machines [2/2]

zheliuyu:
## Summary

Following [[GRPO] fix TRL-liger missing](https://github.com/linkedin/Liger-Kernel/pull/1202) , the `vespo` field was recently added, NPU kernel needs to be updated accordingly.

#### pytest test/transformers/test_grpo_loss.py

```python
=============================================================================================== test session starts ===============================================================================================
platform linux -- Python 3.11.15, pytest-9.0.3, pluggy-1.6.0
rootdir: /home/Liger-Kernel-dev
configfile: pyproject.toml
plugins: asyncio-1.3.0, cov-7.1.0, rerunfailures-16.1, anyio-4.13.0, xdist-3.8.0
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 121 items                                                                                                                                                                                               

test/transformers/test_grpo_loss.py::test_selective_log_softmax[dtype0-0.05-0.5-0.9-1-1024-64000] PASSED                                                                                                    [  0%]
test/transformers/test_grpo_loss.py::test_selective_log_softmax[dtype0-0.05-0.5-0.7-1-1024-151936] PASSED                                                                                                   [  1%]
test/transformers/test_grpo_loss.py::test_grpo_loss[dtype0-0.05-0.5-1-1024-151936-0.7-1-0.0-0.2-0.4] PASSED                                                                                                 [  2%]
test/transformers/test_grpo_loss.py::test_grpo_loss[dtype0-0.05-0.5-1-1024-151936-0.7-1-0.04-0.2-0.4] PASSED                                                                                                [  3%]
test/transformers/test_grpo_loss.py::test_grpo_loss[dtype0-0.05-0.5-1-1024-151936-0.7-5-0.0-0.2-0.4] PASSED                                                                                                 [  4%]
test/transformers/test_grpo_loss.py::test_grpo_loss[dtype0-0.05-0.5-1-1024-151936-0.7-5-0.04-0.2-0.4] PASSED                                                                                                [  4%]
test/transformers/test_grpo_loss.py::test_grpo_loss_with_delta[dtype0-0.05-0.5-2-128-1000-0.7-5-0.0-0.2-0.4-1.5] PASSED                                                                                     [  5%]
[...]

PR #1220, opened 2026-05-11
[NPU, A3] Add NPU kernel support for A3 machines [1/2]

zheliuyu:
## Motivation

This work follows the roadmap in [linkedin/Liger-Kernel#969](https://github.com/linkedin/Liger-Kernel/issues/969). The goal is to exercise the NPU kernel on Atlas 800T A3 (64G) and report how the test suite behaves on that hardware.

## Details

- Fixed the Ascend implementation of attn_res so it passes on A3 alongside the rest of the suite.
- After the fix, the full test run completes successfully. 🍾

image

## Why `attn_res` failed on A3

The failure showed up as **vector-core / ACL errors** (e.g. **507035**, device sync failing), not a normal atol/rtol mismatch.
- Ascend `attn_res` uses **wide masked loads** along the feature dim (e.g. `BLOCK_D = next_power_of_2(D)`).
- Tests include **awkward sizes** like **`D = 123`** with **`float32`**, so **row pitch** is `123 × 4 = 492` bytes—**not 32B/64B-friendly** for many vectorized paths on this stack, which can trigger **vector-core faults** for that lowering.
- **Fix:** **pad** the last dim to a **multiple of 16**, pass **`d_stride`** as the real memory pitch, keep **`D`** for math/masks, and **slice/pad** tensors so callers still see logical **`D`**.

## Benchmark results for the 4 most frequently used kernels 

| | |
| :---: | :---: |
| **cross_entropy_memory_full_token_length** | **cross_entropy_speed_backward_token_length** |
| cross_entropy_memory_full_token_length | cross_entropy_speed_backward_token_length |
| **cross_entropy_speed_forward_token_length** | **cross_entropy_speed_full_token_length** |
| cross_entropy_speed_forward_token_length | cross_entropy_speed_full_token_length |
| **cross_entropy_speed_no-grad-forward_token_length** | **rms_norm_memory_full_token_length** |
| cross_entropy_speed_no-grad-forward_token_length | rms_norm_memory_full_token_length |
| **rms_norm_speed_backward_token_length** | **rms_norm_speed_forward_token_length** |
| rms_norm_speed_backward_token_length | rms_norm_speed_forward_token_length |
| **rms_norm_speed_full_token_length** | **rope_memory_full_token_length** |
| rms_norm_speed_full_token_length | rope_memory_full_token_length |
| **rope_speed_backward_token_length** | **rope_speed_forward_token_length** |
| rope_speed_backward_token_length | rope_speed_forward_token_length |
[...]

PR #1235, opened 2026-05-25
[NPU A3] support fused_moe for A3

sunyi0505:
## Summary

Fixes Ascend NPU crashes in LigerFusedMoEFunction during benchmark_fused_moe (Aicore kernel execute failed on _fused_up_proj_swiglu_kernel). Aligns forward up-proj and backward down-proj GEMM kernels with the existing down-proj pattern: smaller autotune tile sizes and a 1D launch grid with an in-kernel N loop.

## Details

1. Smaller BLOCK_N (UB safety)

    Ascend AICore uses a fixed UB budget per kernel, not CUDA-style large shared memory. With BLOCK_M=32, BLOCK_N=192, BLOCK_K=64, one block needs roughly:
    
    acc_gate / acc_up: 2 × 32 × 192 × 4 bytes ≈ 48 KB
    Weight / activation tiles: tens of KB more
    Compiler temporaries on top
    That commonly triggers UB overflow → AICore execute failed, with no useful Python stack trace. Shrinking tiles to 64/128 matches the CUDA side and other Ascend kernels (e.g. mHC UB-aware tiling).

2. 1D grid + in-kernel N loop (launch limit + consistency)

    Ascend Triton caps total grid programs at 65536 (product of grid dimensions). MoE routing yields num_m_tiles ≈ E × ceil(tokens_per_expert / 32); with large T this grows quickly. A 2D grid multiplies by ceil(I/BLOCK_N) (e.g. ×4 for I=768), so large benchmarks can hit the cap. Moving the N dimension into tl.range(0, I_dim, BLOCK_N):
    
    Keeps launch at (num_m_tiles,) — same as _fused_down_proj_kernel
    Avoids silent launch-limit failures at scale
    Does not change output semantics

3. Backward down-proj aligned proactively

    _moe_bwd_down_proj_kernel had the same 2D grid and BLOCK_N=192 autotune. Fixing only forward would leave backward / full-mode benchmark failures. Collapsing N-tiles also lets dS accumulate in a register across I chunks and use one atomic_add per row instead of one per N-tile — same math, less atomic contention.

## Testing Done

benchmark:
forward:
image
backward:
image
full:
image
memory:
image
ut:
image

- Hardware Type:  Atlas 800T A3

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ops/fused_moe.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/fused_moe.py
@@ -189,7 +189,7 @@ class LigerFusedMoEFunction(torch.autograd.Function):
         post_act = torch.empty(TK, intermediate_dim, dtype=x.dtype, device=x.device)
 
         if num_m_tiles > 0:
-            _fused_up_proj_swiglu_kernel[lambda meta: (num_m_tiles, triton.cdiv(intermediate_dim, meta["BLOCK_N"]))](
+            _fused_up_proj_swiglu_kernel[(num_m_tiles,)](
                 x,
                 gate_up_proj,
                 x_gather_idx,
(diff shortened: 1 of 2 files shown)


sunyi0505:
@Tcc0403 This pr is ready for review. Thanks!  cc @zheliuyu

zheliuyu:
Thanks for the PR @sunyi0505. We previously adapted for A3 in https://github.com/linkedin/Liger-Kernel/pull/1227 and https://github.com/linkedin/Liger-Kernel/pull/1220, and all unit tests passed. However, the `fused_moe` kernel's error on A3 showed up when running `benchmark_fused_moe.py` — likely because the benchmark uses different input shapes than the unit tests. I'll keep following up on this.

PR #1242, opened 2026-05-28
Fix broken gemma3 multimodal forward

sadreamer:
## Summary
Fixes three bugs in multimodal_forward() for Gemma3ForConditionalGeneration by aligning its skip_logits=True (CCE) branch with what causal_forward() already does for the text-only path.

The current convergence test (mini_gemma3 with fused_linear_cross_entropy=False) doesn't exercise either of the buggy code paths, which is why none of these have been caught before.

## Details

**Bug 1: shift_labels silently overwritten (skip_logits branch)**

```python
shift_labels = lm_kwargs.pop("shift_labels", None)
...
if skip_logits:
    shift_hidden_states = kept_hidden_states[..., :-1, :]
    shift_labels = labels[..., 1:]   # <-- discards the popped shift_labels
```

Any caller-supplied pre-shifted shift_labels is silently discarded.

**Bug 2: wrong output length (skip_logits branch)**

The manual pre-shift produces length T-1 tensors and then passes them as both labels= and shift_labels= into LigerForCausalLMLoss. The kernel honors the provided shift_labels (skipping its own pad-then-slice), so per-token output has length T-1 instead of T.

**Bug 3: latent shape mismatch in non-skip_logits branch**

Both the `if labels is not None` and `elif shift_labels is not None` sub-branches hardcode `shift_logits = logits[..., :-1, :]` (length T-1). That's correct for labels[..., 1:] (length T-1), but with a length-T pre-shifted shift_labels (the pad-then-slice convention), the subsequent CrossEntropyLoss shape-mismatches.

**The fix**
- skip_logits path: mirror causal_forward() – pass un-shifted kept_hidden_states and labels to LigerForCausalLMLoss with the popped shift_labels forwarded verbatim.
- non-skip_logits path: collapse the two near-identical if/elif branches into one, and slice shift_logits to match shift_labels.shape[1] so both length-T-1 (drop-first) and length-T (pad-then-slice) conventions work without shape mismatch.

## Testing Done

Apart from running the standard tests below, I also tried fine-tuning Gemma3 on multimodal data. On the current main branch and the latest release (0.8.0), fine-tuning fails with the shape mismatch error e.g. when calling `loss.view_as(input_ids)` (Bug 2), while it works with the fix and shows reasonable train/val losses.

- Hardware Type: H100-80Gb
- [x] run `make test` to ensure correctness -- 3705 passed, 25 failed, 882 skipped. The 25 failures are all in unrelated modules (chunked_loss, flex_attention, fused_linear_jsd, fused_neighborhood_attention, swiglu, plus one low-level liger_fused_linear_cross_entropy structured-output test that does not touch multimodal_forward). The most directly relevant existing test, test_apply_liger_kernel_to_instance_for_gemma3_conditional_generation (which compares inspect.getsource of the bound forward against multimodal_forward), passes.
- [x] run `make checkstyle` to ensure code style -- passes (278 files, all checks passed)
- [x] run `make test-convergence` to ensure convergence -- the fp32/test_mini_models.py session hit 19 pre-existing failures across non-gemma3-multimodal models. Re-ran the convergence tests that actually exercise the modified function directly: 1) fp32 test_mini_models_multimodal.py::mini_gemma3 — PASSED 2) bf16 test_mini_models_multimodal.py::mini_gemma3

### DIFF
--- a/src/liger_kernel/transformers/model/gemma3.py
+++ b/src/liger_kernel/transformers/model/gemma3.py
@@ -246,35 +246,17 @@ def multimodal_forward(
     logits = None
     token_accuracy = None
     predicted_tokens = None
-    if skip_logits and labels is None:
-        raise ValueError("skip_logits is True, but labels is None")
+    if skip_logits and labels is None and shift_labels is None:
+        raise ValueError("skip_logits is True, but both labels and shift_labels are None")
 
     if skip_logits is None:
-        skip_logits = self.training and (labels is not None)
+        skip_logits = self.training and (labels is not None or shift_labels is not None)
 
     if skip_logits:
-        shift_hidden_states = kept_hidden_states[..., :-1, :]
-        shift_labels = labels[..., 1:]
-
-        hidden_device = shift_hidden_states.device
-        if attention_mask is not None:
-            # we use the input attention mask to shift the hidden_states and labels, because it is 2D.
-            # we also crop attn mask in case it is longer, which happens in PrefixTuning with peft
-            shift_attention_mask = attention_mask[:, -shift_hidden_states.shape[1] :].to(hidden_device)
-            shift_hidden_states = shift_hidden_states[shift_attention_mask.to(hidden_device) != 0].contiguous()
-            shift_labels = shift_labels[shift_attention_mask.to(shift_labels.device) != 0].contiguous()
-        else:
-            shift_hidden_states = shift_hidden_states.contiguous()
-            shift_labels = shift_labels.contiguous()
-
-        # Flatten hidden state
-        shift_hidden_states = shift_hidden_states.view(-1, self.config.text_config.hidden_size)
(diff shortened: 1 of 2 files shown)


sadreamer:
@Tcc0403 @Mecoli1219 hi, could you please take a look?

Tcc0403 (commented):
great catch! LGTM! Btw, do you find similar bugs in other models?

Tcc0403 (dismissed):

sadreamer:
@Tcc0403 thanks for the approval! Just checked other models, it seems that the issue is limited to gemma3 and paligemma. I've just fixed the latter too - please take a look.

sadreamer:
@Tcc0403 could you please take a look? Also, please lmk how I can run the CI checks.
Happy to land the PR if you don't have any other concerns

Tcc0403 (approved):
LGTM

PR #1244, opened 2026-05-31
Import ORPO from trl.experimental

kashif:
`LigerORPOTrainer` imports `from trl.trainer import ORPOTrainer`, but recent trl moved ORPO into `trl.experimental.orpo`. `trl.trainer` is a lazy module and no longer exposes `ORPOTrainer`/`ORPOConfig`, so on a current trl that import just raises and `LigerORPOTrainer` can't be imported at all.

This tries `trl.experimental.orpo` first and falls back to the old `trl.trainer` location for older trl, so both work. Same change for the `run_orpo` example's `ORPOConfig` import.

Verified the old path is gone on current trl (`trl.trainer.ORPOTrainer` now raises `AttributeError`) and the experimental path resolves.

### DIFF
--- a/examples/alignment/run_orpo.py
+++ b/examples/alignment/run_orpo.py
@@ -3,7 +3,11 @@ import torch
 from datasets import load_dataset
 from transformers import AutoModelForCausalLM
 from transformers import AutoTokenizer
-from trl import ORPOConfig  # noqa: F401
+
+try:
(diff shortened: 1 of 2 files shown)


kashif:
Folding this into the ORPO trainer bugfix branch (Ola-Vish/Liger-Kernel#1) so the trl.experimental import, example, docs, and the FSDP fix land together. Also addresses the docs/Examples.md snippet Copilot flagged here.

PR #1179, opened 2026-04-01
MoE Triton Kernel

Mecoli1219:
## Summary
<!--- This is a required section; please describe the main purpose of this proposed code change. --->
Triton implementation of a fused MoE expert kernel for training, inspired by SonicMoE: https://arxiv.org/abs/2512.14080                                                         
                                                                                                                                                                                    
  Related to: #958

## Details

**Known limitation:** The current implementation scales well with number of experts (more parallelism) but not with number of tokens. For models with few experts (e.g., Mixtral-8x7B) at large token counts, the eager per-expert loop is likely faster.

**Convergence:** the convergence test uses wider tolerances for MoE models patched with this kernel. 

## Benchmarking Results

- Model: Qwen3-30B-A3B
- #Tokens: 8192 | #Experts: 128 | top-K: 8 | hidden: 2048 | intermediate: 768

| Sweep Dim | #Tokens | #Experts | 
| :--- | --- | --- |
| Memory | <img width="1000" height="600" alt="fused_moe_memory_full_token_length" src="https://github.com/user-attachments/assets/5c55e185-ff15-4d1a-9aab-c0c315f94ea0" /> | <img width="1000" height="600" alt="fused_moe_memory_full_token_length" src="https://github.com/user-attachments/assets/f274f84f-8a7e-4c67-900f-1b96ca5b80d0" /> |
| Speed (Full) | <img width="1000" height="600" alt="fused_moe_speed_full_token_length" src="https://github.com/user-attachments/assets/4350afdc-2820-4748-bd0c-7d05812ddc60" /> | <img width="1000" height="600" alt="fused_moe_speed_full_token_length" src="https://github.com/user-attachments/assets/d537c312-af6d-407f-9271-962b54284cd1" /> |
| Speed (Forward) | <img width="1000" height="600" alt="fused_moe_speed_forward_token_length" src="https://github.com/user-attachments/assets/8256910e-1910-4117-adcf-44cd91e6125a" /> | <img width="1000" height="600" alt="fused_moe_speed_forward_token_length" src="https://github.com/user-attachments/assets/4ece35dd-215d-4c86-a6b6-f92d60c8c22c" /> |
[...]

PR #1248, opened 2026-06-02
fix: int32 pointer overflow in fused MoE kernels (+ autotune env escape hatch)

### LINKED ISSUE #1246: MoE improvements in #1179 break DeepSpeed support on B300 due to auto-enabled autotune that can't be disabled
### 🐛 Describe the bug

# Liger `v0.8.0` Fused MoE Triton Autotune Illegal Memory Access

## Summary

After upgrading `liger-kernel` from commit `b7025a70279b7bea1e635a92f20173abb66498a0` to the `v0.8.0` release tag, Qwen3.5 MoE training crashes inside the new `LigerFusedMoEFunction` path during Triton autotuning.

The crash happens before the kernel returns, while Triton is benchmarking candidate configs for `_fused_up_proj_swiglu_kernel`. My recommendation is to add a supported way to disable autotuning for the fused MoE kernels, for example an environment variable or module/config flag that forces a known-safe default config.

## Environment

- `liger-kernel`: `v0.8.0` from source
- Model family: `Qwen3.5 MoE` (`Qwen3_5MoeSparseMoeBlock` / `Qwen3_5MoeExperts`)
- Training stack: TRL GRPO + DeepSpeed ZeRO-3
- Python: `3.12`
- PyTorch: CUDA build
- Hardware: NVIDIA H200 / B300 class training nodes
- Distributed launch: multi-rank MPI training

## What Changed

The old pinned Liger commit used the non-fused `LigerExperts` path. That path looped over active experts and used `torch.nn.functional.linear` plus `LigerSiLUMulFunction`.

In `v0.8.0`, `LigerExperts.forward()` now calls:

```python
out = LigerFusedMoEFunction.apply(
    x,
    self.gate_up_proj,
    self.down_proj,
    top_k_index_2d,
    top_k_weights_2d,
)
```

That enters `liger_kernel/ops/fused_moe.py`, where `_fused_up_proj_swiglu_kernel` is launched through a Triton autotuned kernel. The failure occurs while Triton is benchmarking autotune candidates.

## Actual Behavior

Training crashes with an illegal memory access during Triton autotune:

```text
[rank29]:   File "/home/ubuntu/monorepo/.venv/lib/python3.12/site-packages/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py", line 786, in forward
[rank29]:     expert_output = self.experts(hidden_states_reshaped, selected_experts, routing_weights)
[rank29]:                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank29]:   File "/home/ubuntu/monorepo/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1776, in _wrapped_call_impl
[rank29]:     return self._call_impl(*args, **kwargs)
[rank29]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank29]:   File "/home/ubuntu/monorepo/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1787, in _call_impl
[rank29]:     return forward_call(*args, **kwargs)
[rank29]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank29]:   File "/home/ubuntu/monorepo/.venv/lib/python3.12/site-packages/liger_kernel/transformers/swiglu.py", line 76, in forward
[rank29]:     out = LigerFusedMoEFunction.apply(x, self.gate_up_proj, self.down_proj, top_k_index_2d, top_k_weights_2d)
[rank29]:           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank29]:   File "/home/ubuntu/monorepo/.venv/lib/python3.12/site-packages/torch/autograd/function.py", line 583, in apply
[rank29]:     return super().apply(*args, **kwargs)  # type: ignore[misc]
[rank29]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank29]:   File "/home/ubuntu/monorepo/.venv/lib/python3.12/site-packages/liger_kernel/ops/utils.py", line 40, in wrapper
[rank29]:     return fn(ctx, *args, **kwargs)
[rank29]:            ^^^^^^^^^^^^^^^^^^^^^^^^
[rank29]:   File "/home/ubuntu/monorepo/.venv/lib/python3.12/site-packages/liger_kernel/ops/fused_moe.py", line 203, in forward
[rank29]:     _fused_up_proj_swiglu_kernel[lambda meta: (num_m_tiles, triton.cdiv(intermediate_dim, meta["BLOCK_N"]))](
[rank29]:   File "/home/ubuntu/monorepo/.venv/lib/python3.12/site-packages/triton/runtime/jit.py", line 370, in <lambda>
[rank29]:     return lambda *args, **kwargs: self.run(grid=grid, warmup=False, *args, **kwargs)
[rank29]:                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank29]:   File "/home/ubuntu/monorepo/.venv/lib/python3.12/site-packages/triton/runtime/autotuner.py", line 240, in run
[rank29]:     benchmark()
[rank29]:   File "/home/ubuntu/monorepo/.venv/lib/python3.12/site-packages/triton/runtime/autotuner.py", line 229, in benchmark
[rank29]:     timings = {config: self._bench(*args, config=config, **kwargs) for config in pruned_configs}
[rank29]:                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank29]:   File "/home/ubuntu/monorepo/.venv/lib/python3.12/site-packages/triton/runtime/autotuner.py", line 164, in _bench
[rank29]:     return self.do_bench(kernel_call, quantiles=(0.5, 0.2, 0.8))
[rank29]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank29]:   File "/home/ubuntu/monorepo/.venv/lib/python3.12/site-packages/triton/testing.py", line 150, in do_bench
[rank29]:     di.synchronize()
[rank29]:   File "/home/ubuntu/monorepo/.venv/lib/python3.12/site-packages/torch/cuda/__init__.py", line 1108, in synchronize
[rank29]:     return torch._C._cuda_synchronize()
[rank29]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank29]: torch.AcceleratorError: CUDA error: an illegal memory access was encountered
```

## Expected Behavior

One of the following would make this easier to recover from:

- Autotune should not try configs that can trigger illegal memory access for this shape/hardware combination.
- If a candidate config fails during benchmarking, autotune should fail gracefully or skip that candidate.
- Users should be able to disable autotune and run the fused MoE kernel with a conservative fixed config.

## Requested Feature

Please add an autotune disable flag for fused MoE kernels. Possible interfaces:

```bash
LIGER_FUSED_MOE_AUTOTUNE=0
```

or:

```python
apply_liger_kernel_to_qwen3_5_moe(..., fused_moe_autotune=False)
```

or a module-level flag on `LigerExperts`.

When disabled, `LigerFusedMoEFunction` could use a known-safe fixed kernel config instead of calling the Triton autotuner. This would let users keep the new fused MoE implementation while avoiding startup crashes caused by unsafe autotune candidates.

## Why This Matters

In distributed training, all ranks may enter autotune for the same shapes. A single rank hitting an illegal memory access during autotune terminates the whole training job. For long-running multi-node GRPO jobs, a conservative non-autotuned mode would be preferable to a startup-time autotune crash.

## Notes

- The stack points to autotuning of `_fused_up_proj_swiglu_kernel`, not to the old GRPO loss Triton path.
- `torch_compile=False` / `TORCH_DYNAMO_DISABLE=1` does not affect this path because this is Triton `@autotune`, not PyTorch `torch.compile`.
- The old pinned Liger commit did not use this fused MoE path, so it did not hit this autotune failure.

### Reproduce

Introduced in #1179 @Mecoli1219 

### Versions

- `liger-kernel`: `v0.8.0` from source
- Model family: `Qwen3.5 MoE` (`Qwen3_5MoeSparseMoeBlock` / `Qwen3_5MoeExperts`)
- Training stack: TRL GRPO + DeepSpeed ZeRO-3
- Python: `3.12`
- PyTorch: CUDA build
- Hardware: NVIDIA B300 training nodes
- Distributed launch: multi-rank MPI training

Mecoli1219:
## Summary

Closes #1246.

Fixes int32 offset overflow in fused MoE pointer arithmetic. When `TK * stride > 2^31`, the kernel's `row_offs * stride` computation wraps and stores land at garbage addresses — manifesting either as `CUDA illegal memory access` during autotune `do_bench` (the original report symptom) or as NaN/Inf in `pre_act` / `post_act` once autotune is bypassed (the follow-up report symptom). Both are the same root cause; the env-var fix landed earlier in this PR addresses neither.

The bug is shape-dependent, not hardware-dependent. It surfaces on real Qwen3.5-MoE training (T~135k, K=8, intermediate_dim=1024 → `TK * stride_pre_TK ~ 2.21 G > 2^31`) and is invisible at the smaller shapes used in existing tests (TK*stride ≤ 1M) and the `QWEN3_MOE_30B` benchmark (TK*stride ~ 100M).

### Root cause

In `_fused_up_proj_swiglu_kernel`:

```python
row_offs = row_start + m_offs                          # int32, max ≈ TK - 1
pre_gate_ptrs = pre_act_ptr + row_offs[:, None] * stride_pre_TK + ...
#                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# pre_act shape (TK, 2*I) → stride_pre_TK = 2*I (in elements)
# Reporter's case: 1,079,839 * 2048 = 2,211,470,272 > int32 max
```

`tl.arange` returns int32, Python int kernel args bind as int32, so the multiplication is done in 32-bit. Cast `row_offs` to int64 before the stride product and the address arithmetic stays correct.

Seven sites where a TK-scale index meets a stride multiplication get the cast:

- `_fused_up_proj_swiglu_kernel`        — pre_act / post_act writes
- `_fused_down_proj_kernel`             — Y writes
- `_token_gather_weighted_sum_kernel`   — Y reads in final aggregation
- `_moe_bwd_down_proj_kernel`           — d_pre_act / weighted_act / dS writes
- `_moe_bwd_dW2_kernel`                 — weighted_act / dout reads
- `_moe_bwd_dX_expanded_kernel`         — d_pre_act / dx_expanded reads/writes
- `_moe_bwd_dW1_kernel`                 — x / d_pre_act reads

### Env var (`LIGER_FUSED_MOE_AUTOTUNE=0`)

The earlier commit on this branch kept the env-var escape hatch in. It is no longer load-bearing for #1246 once the int64 cast lands — but it's harmless API-wise (default unset preserves existing behavior) and gives users a way to opt out of Triton's `do_bench` loop if a future, unrelated autotune-machinery issue surfaces. **Happy to drop the env-var commit if reviewers would prefer a leaner PR** — let me know.

Documented via a `Troubleshooting:` block on `LigerFusedMoEFunction`'s class docstring so users walking up an IMA stack trace can find both the env-var workaround and the underlying issue.

### Regression test

Added `test_large_tk_pointer_overflow_regression` (test/transformers/test_fused_moe.py). Uses shape `T=140000, E=16, K=8, H=512, intermediate_dim=1024` so `TK*stride` clears `2^31` by ~146 M elements (corrupts ~72k rows — enough margin that an allocator handing back zeroed memory can't false-pass). Requires ≥16 GB GPU; skipped otherwise.

Verified locally:
- Fails on the kernel pre-fix commit with NaN/Inf in `out`.
- Passes on the kernel post-fix commit.

## Testing Done

- Hardware Type: <BLANK — please fill in before merge>

Suggested additional verification:
- `pytest test/transformers/test_fused_moe.py` — full existing suite plus the new regression (requires CUDA + ≥16 GB GPU for the regression case).
- B300 reproduction on #1246's actual training stack — deferred to the original reporter to confirm the IMA / NaN-Inf are gone end-to-end.


Mecoli1219:
Hi @michaelroyzen . I am not sure which config is working on B300. Can you check if this PR fix your e2e experiments?

michaelroyzen:
Thank you @Mecoli1219 for taking a look at this so quickly! Will test and let you know

michaelroyzen:
Thank you @Mecoli1219! I will test this ASAP and get back to you.

michaelroyzen:
Thank you @Mecoli1219, confirming this works on H200. No more crashes.

Mecoli1219:
@michaelroyzen Thanks for testing it! Glad that it's working. Can you also confirm if disabling autotune a necessity here?

michaelroyzen:
I think keeping the autotune disable toggle is important because I just had an OOM in the autotune part itself a few steps in. High memory use for autotune is common in my experience.

Mecoli1219:
Got it. Then I'll keep it for now until we find a better way to skip error in triton autotune. 
@Tcc0403 @vaibhavjindal can you review the PR for some final check?

test/transformers/test_fused_moe.py
```diff
@@ -208,3 +208,109 @@ def test_K_equals_E():
...
+
+    out = LigerFusedMoEFunction.apply(x, gate_up_proj, down_proj, top_k_index, top_k_weights)
+
+    # Before the fix: stores to the last rows of pre_act/post_act/Y land at
+    # wrapped int32 addresses → final output has NaN/Inf in those token rows.
+    assert torch.isfinite(out).all(), (
+        "Output contains NaN/Inf — int32 overflow in pre_act/post_act/Y pointer arithmetic. "
+        f"TK={TK:,}, stride_pre_TK={stride_pre_TK:,}, "
+        f"TK*stride={TK * stride_pre_TK:,} (int32 max={2**31 - 1:,})"
+    )
```
Tcc0403: do we still need these tests as we already applied int64 indexing?

Mecoli1219: Removed

src/liger_kernel/ops/fused_moe_kernels.py
```diff
@@ -8,9 +8,17 @@
...
+import os
+
 import triton
 import triton.language as tl
 
+# LIGER_FUSED_MOE_AUTOTUNE=0 pins each kernel to one config, skipping Triton's
+# `do_bench` loop whose per-config working sets can OOM (see issue #1246). Must
+# be set before importing liger_kernel. Temporary escape hatch until triton's
+# autotuner handles such errors itself.
+_AUTOTUNE_DISABLED = os.environ.get("LIGER_FUSED_MOE_AUTOTUNE", "1").lower() in ("0", "false", "no")
```
Tcc0403, line 20: I think it should default to False, only tune when users decide to

michaelroyzen: I think it should be on by default but with clear documentation that it can be disabled. I’ve been using it myself in that manner.

Tcc0403: My concern is that users might not be aware of cold start issue if we set autotune as default, and recompilation might happen if we don't handle it properly. E.g., there was a silent recompilation bug in old version of liger, see: https://x.com/StasBekman/status/2059788636602999247

michaelroyzen: I don't believe it's recompiling based on my profiling. I believe strongly that Liger should default to the best settings "out of the box". That was the goal of the MoE optimizations in #1179, which turned out to be broken for more than a month for larger models until I profiled that issue. This PR even in its current state is a massive win and should be merged ASAP given the main branch (and 0.8.0) is completely broken.

A compromise could be a logged warning to the user that autotuning is enabled and might use extra memory and then they can turn it off if that causes problems.

michaelroyzen: I would also argue that changing the defaults here is outside the scope of this PR. This PR's intention is to fix the main branch and unblock it. Changing the defaults and the design decisions made in #1179, which have been on the main branch for more than a month, is a separate matter and should have its own issue/discussion.

Tcc0403: Fair enough. I agree we should get the fix merged and discuss in another issue.

Tcc0403 (approved):

src/liger_kernel/ops/fused_moe_kernels.py
michaelroyzen: Wonderful, thank you!

PR #1249, opened 2026-06-02
[NPU] remove ineffective params for ASCEND NPU

sunyi0505:
## Summary

remove ineffective params for ASCEND NPU and fix ut.

## Details

1. Refer to docs at: https://github.com/triton-lang/triton-ascend/blob/e9fcd7ddabbbc6de9579a18e507e8f12afbc840c/docs/zh/triton_api/Memory_Pointer_Ops/tl.store.md?plain=1#L31-L32 cache_modifier and eviction_policy are ineffective for Ascend devices. so, remove these params from ascend ops.

2. When running the full set of unit tests, certain test cases for the fused_add_rms_norm kernel fail, whereas standalone unit test execution for this kernel passes without issues. Further investigation shows that the full-test suite generates different random input values which hit an edge case: under bfloat16 with casting_mode="none", the numerical error of weight.grad sits right at the tolerance threshold. We fix the random seed to enforce identical random inputs.

## Testing Done

image

Tested with Atlas 800T A3(X86)

- Hardware Type: <BLANK>

### DIFF
--- a/test/transformers/test_fused_add_rms_norm.py
+++ b/test/transformers/test_fused_add_rms_norm.py
@@ -124,6 +124,7 @@ class GemmaAddRMSNorm(nn.Module):
     ],
 )
 def test_correctness(bs, sl, hd, dtype, atol, rtol, reference, offset, casting_mode, in_place):
+    set_seed(42)
     _tensor = torch.randn(bs, sl, hd, device=device, dtype=dtype)
     _residual = torch.randn(bs, sl, hd, device=device, dtype=dtype)
 
--- a/src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
@@ -133,7 +133,6 @@ def liger_cross_entropy_forward_kernel(
                     logits_row_ptr + X_offsets,
                     mask=X_offsets < n_cols,
                     other=float("-inf"),
-                    eviction_policy="evict_first",
                 ).cast(tl.float32)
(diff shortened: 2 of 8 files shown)


sunyi0505:

test/transformers/test_fused_add_rms_norm.py
```diff
@@ -124,6 +124,7 @@ def forward(self, x, residual):
     ],
 )
 def test_correctness(bs, sl, hd, dtype, atol, rtol, reference, offset, casting_mode, in_place):
+    set_seed(42)
```
Tcc0403, line 127: why do we need to set seed here? is it a flaky test?

sunyi0505: Yes — it is explicitly marked flaky:
@pytest.mark.flaky(reruns=3, reruns_delay=2)

Without resetting the random seed, previously executed parameterized test cases advance the global RNG state, causing generated input data to depend on execution order. This test consistently fails when I run the full UT suite on Ascend Atlas 800T A3.

Tcc0403 (approved):
LGTM

test/transformers/test_fused_add_rms_norm.py
Tcc0403: We should make it a pytest fixture to auto use `set_seed()` in conftest for reproducibility.

sunyi0505: > We should make it a pytest fixture to auto use `set_seed()` in conftest for reproducibility.

Okay, I will fix it in the later PR.

PR #1250, opened 2026-06-04
[cuTile] Add cutile backend for geglu, layer_norm, fused_linear_jsd, cross_entropy

xjmxyt:
## Summary

Adds **cuTile backend implementations** for four Liger-Kernel operators, mirroring the existing Triton kernels under src/liger_kernel/ops/cutile/ops/:

- cross_entropy — fused forward + in-place gradient, online numerically-stable logsumexp
- fused_linear_jsd — chunked fused linear + JSD (reuses the cuTile jsd_kernel)
- geglu — GEGLU (tanh-approx GELU) forward/backward
- layer_norm — forward (single-pass mean/var via fold trick) + combined DX/DW/DB backward

Also adds a shared element_mul_kernel to cutile/ops/utils.py and wires all new symbols into cutile/ops/__init__.py.

## What's included

- cutile/ops/cross_entropy.py — LigerCrossEntropyFunction, cross_entropy_forward/backward
  - Supports label_smoothing, lse_square_scale (z-loss), per-class weight, softcap, ignore_index, reduction
  - Adds optional return_token_accuracy / return_predicted_tokens (cross-chunk argmax tracking) — **new logic not present in upstream Triton**
- cutile/ops/fused_linear_jsd.py — LigerFusedLinearJSDFunction, fused_linear_jsd_forward/backward
- cutile/ops/geglu.py — LigerGELUMulFunction, geglu_forward/backward (aligned/check_bounds=False fast path)
- cutile/ops/layer_norm.py — LigerLayerNormFunction, layer_norm_forward/backward (fast path skips DW/DB when not needed)
- cutile/ops/utils.py — element_mul_kernel
- Benchmark data refresh in benchmark/data/all_benchmark_data.csv and all_benchmark_data_cutile.csv

## Implementation notes

- All kernels are faithful ports of the upstream Liger Triton implementations, including the math (online logsumexp, label-smoothing, softcap chain rule sech² = 1 − tanh², GEGLU tanh-approx fwd/bwd, layer-norm c1/c2 reduction) and dtype conventions (loss_1d/Mean/RSTD follow upstream).
- Aligned-shape fast paths (check_bounds=False) are used when n_cols is a power of 2 to enable the hardware TMA path.
- layer_norm backward uses a persistent grid (grid = sm_count) with per-block DW/DB partials reduced on host.

## Benchmark results
image
image
image
image
image
image

## Testing Done

- Hardware Type: <BLANK>


xjmxyt:
@vaibhavjindal @Tcc0403 @Mecoli1219 @yueyiming2009 Could you please help to review this? Thanks a lot.

src/liger_kernel/ops/cutile/ops/layer_norm.py
```diff
@@ -0,0 +1,346 @@
...
+    check_bounds = not ALIGNED
+
+    # ---- Pass 1: compute sum(x) and sum(x^2) via fold trick ----
+    # OOB column positions (when ALIGNED=False, last chunk) padded with 0.0 → correct
+    sum_tile = ct.full((BLOCK_SIZE,), 0.0, dtype=ct.float32)
+    sum_sq_tile = ct.full((BLOCK_SIZE,), 0.0, dtype=ct.float32)
+
+    for ci in range(n_chunks):
+        col_idx = ct.add(ct.arange(BLOCK_SIZE, dtype=ct.int32), ci * BLOCK_SIZE)
+        x_tile = ct.astype(ct.gather(x, (row_idx, col_idx), check_bounds=check_bounds, padding_value=0.0), ct.float32)
```
Tcc0403, line 70: not familiar with cutile, I'm curious why we don't use `ct.load` directly. Same question to `ct.store`. It seems they both have out-of-bound protection.

xjmxyt: `ct.load` and `ct.gather` correspond to two different cuda-tile op.
```
ct.load(..., allow_tma=False) → cuda_tile.load_view_tko
ct.gather(...) → cuda_tile.load_ptr_tko
```
Because they map to distinct ops, their performance can differ. The key reason is who owns the pointer (address) computation:
 • For the tensor-of-pointers load, the user computes the pointers, so they can simplify the address math as much as the use case allows.
 • For ct.load(..., allow_tma=False), the compiler computes the pointers, and it can only simplify them when it has enough information to do so.

Recommendation: use ct.load when you want TMA; otherwise ct.gather usually matches or beats ct.load(..., allow_tma=False).

Tcc0403 (commented):
Overall lgtm, could you also attach the testing results for these ops with cutile backend? cc @Mecoli1219, @vaibhavjindal for a second look.

vaibhavjindal:
@xjmxyt thanks for the PR. The kernels look good to me. However, it would be helpful if you can add some plots showing the performance of (cutile vs triton vs pytorch) on different input shapes/sizes so that we can see what works best in each case and make further improvements.

benchmark/data/all_benchmark_data.csv
```diff
@@ -2219,3 +2219,139 @@ jsd,liger,full,memory,MB,BT,total tokens,1024,3514.0009765625,3514.0009765625,35
 jsd,liger,full,memory,MB,BT,total tokens,2048,7014.0009765625,7014.0009765625,7014.0009765625,"{""vocab_size"": 128256, ""bsz"": 1, ""seq_len"": 8192}",NVIDIA B200,2026-05-27 17:14:32,0.8.0
 jsd,liger,full,memory,MB,BT,total tokens,4096,14028.0009765625,14028.0009765625,14028.0009765625,"{""vocab_size"": 128256, ""bsz"": 1, ""seq_len"": 8192}",NVIDIA B200,2026-05-27 17:14:32,0.8.0
 jsd,liger,full,memory,MB,BT,total tokens,8192,28056.0,28056.0,28056.0,"{""vocab_size"": 128256, ""bsz"": 1, ""seq_len"": 8192}",NVIDIA B200,2026-05-27 17:14:32,0.8.0
+layer_norm,liger,forward,speed,ms,T,Sequence length,1024,0.010239999741315842,0.010214400291442872,0.010239999741315842,"{""hidden_size"": 4096, ""dtype"": ""torch.bfloat16"", ""eps"": 1e-06, ""bsz"": 1, ""seq_len"": 8192}",NVIDIA B300 SXM6 AC,2026-06-04 07:58:01,0.8.0
```
vaibhavjindal: When doing comparisons between (cutile vs triton vs pytorch), let's add all the benchmarks data in `all_benchmark_data_cutile.csv`? Otherwise, we have 2 different benchmark data for kernels such as `layer_norm`, `geglu` etc in `all_benchmark_data.csv`. It would also be difficult to plot things in this case using the visualizer script.

So maybe we can modify the benchmark script so that all the (cutile vs triton vs pytorch) data sits in `all_benchmark_data_cutile.csv`, and can be easily plotted via visualiser script?

src/liger_kernel/ops/cutile/ops/cross_entropy.py
```diff
@@ -0,0 +1,472 @@
...
+            input_tile = ct.mul(softcap, ct.tanh(ct.mul(input_raw, 1.0 / softcap)))
+        else:
+            input_tile = input_raw
+
+        block_max = ct.max(input_tile, 0, keepdims=False)
+        neg_bmax_log2e = -block_max * LOG2E
+        block_sum_exp = ct.sum(ct.exp2(input_tile * LOG2E + neg_bmax_log2e, flush_to_zero=True), 0, keepdims=False)
+
+        # Track argmax for token accuracy / predicted tokens
+        if RETURN_TOKEN_ACCURACY or RETURN_PREDICTED_TOKENS:
```
vaibhavjindal, line 140: I was curious about the impact of adding `RETURN_TOKEN_ACCURACY` and `RETURN_PREDICTED_TOKENS` on speed/memory. Can we do a simple experiment to see how much impact it has on speed/memory of the kernel. This might help explain the difference in performance between triton vs cutile.

xjmxyt: For triton implementation, it also uses `RETURN_TOKEN_ACCURACY` and `RETURN_PREDICTED_TOKENS`(see https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/ops/cross_entropy.py#L154) . For this part, there's no difference.
I run quick experiments on whether enabling it. Below is the data.

Image

benchmark/data/all_benchmark_data.csv
xjmxyt: I already change this.
The `run_cutile_compare.py` driver runs the kernel benchmark twice (Triton baseline + CuTile) and routes all series into all_benchmark_data_cutile.csv, so the comparison lives in a single file rather than being split against all_benchmark_data.csv.

The two pieces that make it work:
- LIGER_BENCH_TARGET=cutile selects all_benchmark_data_cutile.csv as the destination (instead of the default all_benchmark_data.csv).
- LIGER_BENCH_PROVIDER_TAG renames the liger provider per run (liger_triton / liger_cutile) so the two implementations don't collide on the dedup key. The torch/huggingface baseline rows carry their own provider names already.

So for any kernel:
```
python scripts/run_cutile_compare.py --kernel layer_norm [bench args...]
```
produces liger_triton, liger_cutile, and the torch/HF baseline all in all_benchmark_data_cutile.csv, ready to plot directly
For plotting results, we can use
```
python ../benchmarks_visualizer.py \
    --kernel-name layer_norm --metric-name speed \
    --data-file data/all_benchmark_data_cutile.csv
```
Could you please review this part again?

xjmxyt:
> @xjmxyt thanks for the PR. The kernels look good to me. However, it would be helpful if you can add some plots showing the performance of (cutile vs triton vs pytorch) on different input shapes/sizes so that we can see what works best in each case and make further improvements.

@Tcc0403 @vaibhavjindal I already attach some plots generated by `benchmarks_visualizer.py` in the summary.

benchmark/data/all_benchmark_data.csv
vaibhavjindal: Thanks for explaining this. Makes sense with the new `run_cutile_compare.py` file.

src/liger_kernel/ops/cutile/ops/cross_entropy.py
vaibhavjindal: Thanks for doing this analysis. The flags definitely have a decent impact (~15% increase) on latency. 

I thought this is not present in triton as you mentioned `new logic not present in upstream Triton` in the PR description.

vaibhavjindal (approved):
LGTM! Thanks for contributing these cuTile kernels to liger!

PR #1260, opened 2026-06-12
[Megatron] Add vocab-parallel cross-entropy (TP>1 support)

vaibhavjindal:
## Summary

Adds TP>1 (vocab-parallel) support to Liger's Megatron-LM cross-entropy. The previous PR (#1207) shipped CE for TP=1 only. At TP>1, each rank holds a slice of the vocabulary and the loss needs cross-rank all-reduces to be correct. This PR writes the vocab-parallel kernel and wires it through `LigerMegatronCrossEntropy` so the same one-line monkey patch now works at **every** TP size — TP=1, 2, 4, 8, ...

Single code path. AllReduces are no-ops at TP=1 and real NCCL collectives at TP>1.

Two new public Module classes — both backed by the same kernel:

- **`LigerVocabParallelCrossEntropy`** (`liger_kernel.transformers`) — general-purpose vocab-parallel CE. Defaults match `torch.nn.functional.cross_entropy` semantics.
- **`LigerMegatronCrossEntropy`** (`liger_kernel.megatron`) — same kernel, Megatron-LM defaults. Bit-identical to Megatron's `vocab_parallel_cross_entropy`.

Label smoothing supports both PyTorch's α/K formula and Megatron's NeMo α·V/(V−1) formula; pick via `label_smoothing_formula="pytorch"` or `"megatron"`. At TP>1 the averaging scope is selectable too: `label_smoothing_mode="global"` (exact, one extra AllReduce) or `"partition"` (bit-identical to Megatron, no extra AllReduce, per-rank loss may differ).

### Future Work
- Comparison with TE kernels
- exp in bf16

```python

apply_liger_kernel_to_megatron(rms_norm=True, cross_entropy=True)
```

Defaults of `LigerMegatronCrossEntropy` match Megatron's vocab-parallel CE bit-for-bit, so users monkey-patching from Megatron see zero numerical change.

### Mode 2: Use the classes directly

```python
# Drop-in for Megatron's vocab_parallel_cross_entropy:
from liger_kernel.megatron import LigerMegatronCrossEntropy
ce = LigerMegatronCrossEntropy(ignore_index=-100, label_smoothing=0.0)
loss = ce(vocab_parallel_logits, target, tp_group=my_tp_group)

# Or, general-purpose vocab-parallel CE with PyTorch / F.cross_entropy semantics:
from liger_kernel.transformers import LigerVocabParallelCrossEntropy
ce = LigerVocabParallelCrossEntropy(ignore_index=-100, label_smoothing=0.0)
loss = ce(vocab_parallel_logits, target, tp_group=my_tp_group)
```

## Scope

In scope:
- Vocab-parallel CE for TP=1, 2, 4, 8, … (any TP size).
- Both label-smoothing formulas + both averaging modes at TP>1.

---

## Performance — Liger vs Megatron's vocab-parallel CE

Measured on **8× NVIDIA H100 80GB HBM3** at the Megatron training shape `S=2048, B=4, bf16`. Compared against `megatron-core==0.19.0+96a894ef6`:

- `megatron-fused`: Megatron's *fused* `fused_vocab_parallel_cross_entropy` (TorchScript JIT via `@jit_fuser`)
- `megatron-unfused`: Megatron's *unfused* `vocab_parallel_cross_entropy` (eager Python; the path with runtime `label_smoothing` support)

### Headline: Liger vs Megatron-fused at production V (≥32K)

|  | Forward | Backward (incl. fwd) | Full (fwd+bwd) |
|---|---|---|---|
| **TP=1** at V=128K  | **+298%** | +129% | +129% |
| **TP=2** at V=128K  | +274% | +123% | +123% |
| **TP=4** at V=128K  | +239% | +114% | +114% |
| **TP=8** at V=128K  | +210% | +108% | +108% |
| **TP=8** at V=32K   | +21%  | +24%  | +24%  |

At small `V_local` (high TP × small V) the kernels become tiny and Python orchestration dominates; even there Liger is at-worst tied with Megatron-fused, never significantly slower. At production-scale V (32K and up), Liger is **2–4× faster on forward** and **~2× faster on backward** at every TP we tested.

vs `megatron-unfused`: Liger is **30–430% faster** at every (TP, V) combination — that path doesn't JIT-fuse anything.

### Per-TP plots

megatron_cross_entropy_tp1_panel
megatron_cross_entropy_tp2_panel
megatron_cross_entropy_tp4_panel
megatron_cross_entropy_tp8_panel

(Drag-and-drop the 4 PNGs at `/tmp/megatron_ce_panels/megatron_cross_entropy_tp{1,2,4,8}_panel.png` into the PR description here. Each PNG has forward / backward / full speed + memory in one 2×2 view.)

### Why is Liger faster?

H100 CE is **memory-bandwidth-bound**, not compute-bound. At V=128K the input logits tensor is ~2GB of bf16; each full read or write of it is ~0.7ms on HBM3, and the runtime is dominated by how many times you touch that buffer.

**Megatron's fused (TorchScript) path does it in stages**, with each stage being its own JIT-fused kernel that allocates a new buffer:
- `calculate_logits_max` casts the input bf16 → fp32 (writes 4GB), then computes max
- `calculate_predicted_logits` shifts in-place (read 4GB + write 4GB), then exps in-place (read 4GB + write 4GB), then gathers predicted and sum_exp
- → roughly **4 full reads + 3 full writes** of the V_local-sized buffer

**Liger does it in ONE pass.** Kernel 1 reads input bf16, casts fp32 in registers, subtracts the global max in registers, exps in registers, writes the exp buffer (fp32) once, and accumulates `sum_exp` + `predicted_logit` as scalars on the way through. Kernel 2 (backward) does symmetrically one read + one write. AllReduces happen on tiny `[BT]`-shaped tensors between kernels at TP>1.

That's ~3.5× less memory traffic — which matches the observed 3-4× forward speedup almost exactly. The backward speedup (~2×) is smaller because the backward is fundamentally one-pass-shaped for both implementations; the win there is fusing the `grad_output` scaling into the same kernel as the gradient computation instead of doing it as a separate buffer mul.

---

## Memory — Liger vs Megatron

**What "memory" means at TP>1.** The number we report is `torch.cuda.max_memory_allocated()` on rank 0 — peak HBM usage on a **single GPU** during one fwd+bwd CE call. At TP=N each rank holds only `V_global/N` of the vocab, so per-rank memory **scales as 1/N**. All ranks have identical memory profiles (vocab is partitioned evenly), so rank 0 represents every rank. It is **not** "aggregate across all GPUs" — each GPU is independently HBM-bound, so per-rank is the metric that matters for fitting a model on a node.

### Numbers (MB per rank — Liger / Megatron-fused / Megatron-unfused)

| V | TP=1 | TP=2 | TP=4 | TP=8 |
|---|---|---|---|---|
| 4K   | 256 / 384 / 256       | 128 / 192 / 128     | 64 / 96 / 64     | 32 / 48 / 32     |
| 16K  | 1024 / 1536 / 1024    | 512 / 768 / 512     | 256 / 384 / 256  | 128 / 192 / 128  |
| 32K  | 2048 / 3072 / 2048    | 1024 / 1536 / 1024  | 512 / 768 / 512  | 256 / 384 / 256  |
| 128K | 8192 / 12288 / 8192   | 4096 / 6144 / 4096  | 2048 / 3072 / 2048 | 1024 / 1536 / 1024 |

**Two takeaways:**

1. **Liger uses 33% less per-rank HBM than Megatron-fused at every (TP, V) point.** That's 512 MB freed at V=128K/TP=8 — and the same 512 MB savings replicates across all 8 ranks. Usable for longer sequences, larger batches, or activation checkpointing.

2. **Liger uses exactly the same memory as Megatron-unfused.** Both save a single `[BT, V_local]` fp32 softmax buffer for backward — that's the irreducible lower bound for vocab-parallel CE. The plot annotates this overlap explicitly.

### Why is Megatron-fused 50% heavier than Liger / Megatron-unfused?

Both Liger and Megatron-unfused save one `[BT, V_local]` fp32 softmax buffer for backward. Megatron-fused saves the same buffer, but its `@jit_fuser` TorchScript scope holds **captured input tensors alive until the function returns** — so during forward the bf16 input AND the fp32 working buffer are both resident in HBM, peaking at ~1 extra `[BT, V_local]` buffer's worth. The 50% premium is consistent across V and TP.

Liger avoids this entirely: the bf16 input is read in-register inside Kernel 1 and never replicated to a separate fp32 buffer in HBM. Megatron-unfused avoids it because eager Python rebinds the local variable after `.float()`, letting the bf16 input be GC'd.

So Liger and unfused arrive at the same lower bound through different routes (Triton single-pass kernel vs eager Python ops), and Megatron-fused pays a 50% premium in exchange for TorchScript fusion that — as the speed numbers show — doesn't actually pay off vs hand-written Triton.

---

## Testing

All unit tests live in `test/megatron/`.

### Single-GPU tests

```bash
python -m pytest test/megatron/test_cross_entropy.py --no-cov -k "not tp_vocab_parallel"
python -m pytest test/megatron/test_monkey_patch.py --no-cov
```

Covers:
- `LigerVocabParallelCrossEntropy` correctness — fwd + bwd parity vs `F.cross_entropy` across shape sweeps, fp32 / bf16, `ignore_index`, `label_smoothing`, large logit magnitudes, out-of-bounds targets, CE-not-last-layer, `torch.no_grad()`.
- `LigerMegatronCrossEntropy` correctness — same matrix, with `label_smoothing > 0` parity referenced against `F.cross_entropy(label_smoothing=α·V/(V-1))` which is algebraically equivalent to Megatron's NeMo formula at TP=1.
- Config plumbing: validation of `reduction`, `label_smoothing_formula`, `label_smoothing_mode`, `label_smoothing` range; 3-D shape guard.
- Input-preservation invariant: forward must NOT mutate the caller's input tensor even when fp32 (regression test for in-place ops aliasing).
- Monkey-patch plumbing: symbol replacement for both fused / unfused CE paths, idempotency, missing-megatron errors, runtime `label_smoothing` override on the unfused path, end-to-end correctness through patched symbols.

77 single-GPU CE tests + 35 monkey-patch tests = 112 single-GPU tests, all passing.

### Multi-GPU TP>1 parity tests

```bash
python -m pytest test/megatron/test_cross_entropy.py --no-cov -k "tp_vocab_parallel"
```

Parametrized over `tp_size ∈ {2, 4, 8}` with per-tp_size `skipif(torch.cuda.device_count() < tp_size, ...)`. 48 tests total covering the full cross-product `formula × mode × label_smoothing × shape`. The right subset runs based on available GPUs (2 GPUs: 16 run, 32 skip; 8 GPUs: all 48 run).

Each test broadcasts the same full-vocab logits to every rank, has each rank slice out its `V_local`, runs Liger's vocab-parallel CE, and compares forward + backward against a single-rank `F.cross_entropy` reference on the full logits (with α rescaled for the Megatron formula).

### Style

```bash
make checkstyle
```

Clean.

---

## Benchmarking

Single script handles all TP sizes via `--tp-size`. Uses `mp.spawn` internally and allocates `tp_size` GPUs.

```bash
python benchmark/scripts/benchmark_megatron_cross_entropy.py --tp-size 1 --overwrite
python benchmark/scripts/benchmark_megatron_cross_entropy.py --tp-size 2 --overwrite
python benchmark/scripts/benchmark_megatron_cross_entropy.py --tp-size 4 --overwrite   # 4 GPUs
python benchmark/scripts/benchmark_megatron_cross_entropy.py --tp-size 8 --overwrite   # 8 GPUs

# Render per-TP plots from the CSV
    --kernel-name megatron_cross_entropy --metric-name speed --overwrite \
    --extra-config-filter '"TP": 1'
```

Note: the committed CSV intentionally contains **only TP=1 rows** for `megatron_cross_entropy` to keep the data file small. Re-run at TP=2/4/8 to repopulate the rows for those TPs if you want to validate the numbers in the headline table above.

## Testing Done

- Hardware Type: H100 80GB HBM3 × 8
- [x] `python -m pytest test/megatron/ --no-cov` — passes (77 single-GPU CE + 35 monkey-patch + N multi-GPU CE depending on `cuda.device_count()` + 13 RMSNorm)
- [x] `python -m pytest test/transformers/test_cross_entropy.py --no-cov -k "test_correctness and not z_loss"` — 136 passed (existing CE not regressed)
- [x] `make checkstyle` — clean
- [x] Full TP=1/2/4/8 benchmark sweep on 8×H100 — numbers committed inline above; 4-panel plots attached
- [ ] `make test-convergence`


src/liger_kernel/ops/vocab_parallel_cross_entropy.py
```diff
@@ -0,0 +1,351 @@
...
+        vocab_parallel_logits: torch.Tensor,
+        target: torch.Tensor,
+        tp_group,
+        ignore_index: int,
+        label_smoothing: float,
+        label_smoothing_formula: str,
+        label_smoothing_mode: str,
+    ) -> torch.Tensor:
+        s, b, v_local = vocab_parallel_logits.shape
+        v_local_int = int(v_local)
```
kolehma8: nit, is this really needed, I thought shape elements have to be integers

src/liger_kernel/ops/vocab_parallel_cross_entropy.py
```diff
@@ -0,0 +1,351 @@
...
+        ctx,
+        vocab_parallel_logits: torch.Tensor,
+        target: torch.Tensor,
+        tp_group,
+        ignore_index: int,
+        label_smoothing: float,
+        label_smoothing_formula: str,
+        label_smoothing_mode: str,
+    ) -> torch.Tensor:
+        s, b, v_local = vocab_parallel_logits.shape
```
kolehma8, line 224: should you assert that shape has enough elements?

src/liger_kernel/ops/vocab_parallel_cross_entropy.py
```diff
@@ -0,0 +1,351 @@
...
+            rank = dist.get_rank(tp_group)
+            world = dist.get_world_size(tp_group)
+        else:
+            rank, world = 0, 1
+        vocab_start = rank * v_local_int
+
+        # --- 1. Per-token global max (cheap; small output) ---
+        # amax on the input dtype, cast the [S, B] result to fp32 for the kernel.
+        # No full-buffer fp32 conversion.
+        x = vocab_parallel_logits.detach().contiguous()
```
kolehma8: why detach()?

src/liger_kernel/ops/vocab_parallel_cross_entropy.py
```diff
@@ -0,0 +1,351 @@
...
+        logits_max = x.reshape(BT, v_local_int).amax(dim=-1).float()
```
kolehma8: view(-1, v_local_int) does the same as x is already contiguous

src/liger_kernel/ops/vocab_parallel_cross_entropy.py
```diff
@@ -0,0 +1,351 @@
...
+        if tp_distributed:
+            dist.all_reduce(logits_max, op=dist.ReduceOp.MAX, group=tp_group)
+
+        # --- 2. Flatten target once ---
+        flat_target = target.reshape(-1).contiguous().to(torch.int64)
+
+        # --- 3. Kernel 1: shift, exp, sum_exp, predicted_local ---
+        exp_buf = torch.empty(BT, v_local_int, device=x.device, dtype=torch.float32)
```
kolehma8: should you use the vocab datatype here as logits could be BF16 (or something else)?

src/liger_kernel/ops/vocab_parallel_cross_entropy.py
vaibhavjindal: right, not needed. will fix this

src/liger_kernel/ops/vocab_parallel_cross_entropy.py
vaibhavjindal: sure, will add this.

src/liger_kernel/ops/vocab_parallel_cross_entropy.py
vaibhavjindal: yes, not needed.

src/liger_kernel/ops/vocab_parallel_cross_entropy.py
vaibhavjindal: yes, can be moved to ` logits_max = x.amax(dim=-1).float()`

src/liger_kernel/ops/vocab_parallel_cross_entropy.py
vaibhavjindal: megatron does it in fp32 for precision that's why kept it fp32.
Maybe we can later do a convergence test and see if we can move to bf16?

kolehma8 (approved):

PR #1263, opened 2026-06-18
remove incorrect import for qwen3_5 monkey patch with cross_entropy=True

aeflores:
## Summary
The function `apply_liger_kernel_to_qwen3_5` would crash when called with `cross_entropy=True` because it contains an incorrect import. Moreover, function `liger_cross_entropy` is already imported in line 15, so I just removed the extra incorrect import.


Tcc0403 (approved):
great catch! LGTM!

PR #1273, opened 2026-06-27
Add infer_device_arch() for coarse GPU architecture detection

yueyiming2009:
## Summary

`infer_device()` only returns the device *type* (`"cuda"` / `"xpu"` / `"npu"` / `"cpu"`), which collapses every NVIDIA and AMD GPU into `"cuda"` and exposes no way to branch on hardware generation (e.g. Blackwell-specific paths).

This PR adds **`infer_device_arch()`** to `src/liger_kernel/utils.py`, which resolves a coarse architecture/generation name:

- **NVIDIA** (via CUDA compute-capability major): `volta_turing` / `ampere_ada` / `hopper` / `blackwell` (else `sm_<major>0`)
- **AMD** (via `gcnArchName`): `cdna` / `cdna2` / `cdna3` / `rdna3` (else the raw gfx target)
- **Intel XPU** (via device name): `pvc` / `arc` (else `xpu`)
- **Ascend NPU** (via device name): `ascend910` / `ascend310` (else `npu`)

It falls back to the `infer_device()` device type when the architecture can't be determined (or on error), is `lru_cache`-d so detection runs once per process, and distinguishes ROCm — which also reports as `"cuda"` in torch — from NVIDIA via `torch.version.hip`.

## Details

The NVIDIA and AMD paths are reliable. The XPU and NPU paths are best-effort string matching on the device name and were validated via mocking rather than on real Intel/Ascend hardware — the mapping tables (`_AMD_ARCH_BY_GFX`, `_NVIDIA_ARCH_BY_MAJOR`, and the XPU/NPU name keywords) are the spots to refine as new parts ship.

## Testing Done

Added `test/transformers/test_utils.py` (30 cases) covering all four device families' mapping logic, NVIDIA-vs-AMD dispatch, the exception/unaccelerated fallbacks, and `lru_cache` behavior. The hardware-specific torch calls are mocked so the full matrix runs on a CPU-only host.

- Hardware Type: CPU (Apple Silicon, macOS) — GPU paths exercised via mocks
- [x] run `make checkstyle` to ensure code style (`ruff check` + `ruff format --check` pass on changed files)
- [x] run `make test` to ensure correctness (ran the new `test/transformers/test_utils.py` suite: 30 passed; full `make test` requires a GPU + deps not available locally)
- [ ] run `make test-convergence` to ensure convergence (N/A — no kernel/numerics change)


src/liger_kernel/utils.py
```diff
@@ -65,6 +67,87 @@ def is_npu_available() -> bool:
         return False
 
 
+# NVIDIA: CUDA compute-capability major -> coarse arch family
+_NVIDIA_ARCH_BY_MAJOR = {
+    7: "volta_turing",  # 7.0 Volta, 7.5 Turing
+    8: "ampere_ada",  # 8.0/8.6 Ampere, 8.9 Ada Lovelace
+    9: "hopper",  # H100
+    10: "blackwell",  # B100/B200/GB200
+    12: "blackwell",  # consumer RTX 50xx
```
Tcc0403: sm100 and sm120 are quite different.  Although triton has abstracted low-level instructions away, writing heuristic config selection based on corase arch family name might not be proper.

src/liger_kernel/utils.py
```diff
@@ -65,6 +67,87 @@ def is_npu_available() -> bool:
...
+_AMD_ARCH_BY_GFX = {
+    "gfx908": "cdna",  # MI100
+    "gfx90a": "cdna2",  # MI200
+    "gfx940": "cdna3",  # MI300
+    "gfx941": "cdna3",
+    "gfx942": "cdna3",  # MI300X/MI300A
+    "gfx1100": "rdna3",  # RX 7900
+    "gfx1101": "rdna3",
+    "gfx1102": "rdna3",
+}
```
Tcc0403, line 93: ditto

src/liger_kernel/utils.py
```diff
@@ -65,6 +67,87 @@ def is_npu_available() -> bool:
...
+    return "xpu"
+
+
+def _infer_npu_arch(device_id: int) -> str:
+    name = torch.npu.get_device_properties(device_id).name.lower()
+    if "910" in name:
+        return "ascend910"
+    if "310" in name:
+        return "ascend310"
+    return "npu"
```
Tcc0403, line 122: @zheliuyu any other npu models you might want to add?

src/liger_kernel/utils.py
yueyiming2009: Good point. Reworked the NVIDIA path to key off the full `(major, minor)` compute capability, so sm_100, sm_103, and sm_120 are now distinct labels (`blackwell` / `blackwell_ultra` / `blackwell_consumer`) instead of all collapsing to `blackwell`. Unknown caps fall back to the exact `sm_<major><minor>` (e.g. `sm_87`).

If you'd rather not bake in family labels at all, I can drop them and just return `sm_<major><minor>`, letting callers map sm -> config themselves. Which do you prefer? (Pushing the change shortly.)

src/liger_kernel/utils.py
yueyiming2009: Same idea here: AMD detection already keys off the full `gfx` target, with unknown targets falling through to the raw `gfxXXXX`, so the only coarse part is the family label (`cdna3`, `rdna3`) for known targets. I'll follow whatever we settle on in the NVIDIA thread (keep family labels vs. return the raw `gfx` target).

src/liger_kernel/utils.py
zheliuyu: Thanks for your reminder. There are currently no other NPU machine types available.

vaibhavjindal (approved):
LGTM

vaibhavjindal:
The current version looks good to me. @Tcc0403 do you think it is good to merge now?

PR #1274, opened 2026-06-28
Avoid torch.compile graph break in SAPO loss

kashif:
The SAPO branch of the GRPO loss assigned per-token loss values using boolean-mask indexing (`per_token_loss[mask] = sapo_loss_fn(coef_1[mask], ...)`). That kind of indexing lowers to `aten.nonzero`, whose output shape is data-dependent, so Inductor cannot trace it and torch.compile inserts a graph break. With recent torch versions this shows up as a noisy backend compiler warning when running the SAPO correctness test.

Since `sapo_loss_fn` is elementwise, we can evaluate both temperature branches over the full tensor and pick by the advantage sign with `torch.where`. The result is numerically identical and compiles cleanly with no graph break.

The existing SAPO correctness test passes and the warning is gone.


kashif:
cc @vaibhavjindal

vaibhavjindal (approved):
LGTM!
Ran 
```
python -m pytest test/chunked_loss/test_grpo_loss.py -q -k "sapo" --no-cov -p no:cacheprovider
``` 
to verify it

PR #1266, opened 2026-06-25
perf(ce): hardware exp2 (ex2.approx) for softmax

justinhh4:
## PR1: CE `exp2` — B200 + H100 benchmarks

Optimization measured against the **frozen original baseline** (`v0_base`, == upstream HEAD). Correctness: full CE suite (177 cases, bf16+fp32, weight/softcap/smoothing/z_loss/ignore_index combos) passes.

### What & why
Replace `tl.exp` with the hardware `ex2.approx` instruction via `e^x = 2^(x·log₂e)`, in both the online-softmax accumulation and the gradient pass.
- **Why:** NVIDIA GPUs have no native `e^x` — `tl.exp` lowers to a multi-instruction accurate sequence, while `tl.exp2` emits a single `ex2.approx.f32`. This is the most frequent FP op in the kernel (once per vocab element × 2 passes), so it cuts total issued instructions.
- **Accuracy:** `ex2.approx` has ~2.4e-7 relative error — within the suite's fp32 `rtol=1e-6` and bf16 `rtol=5e-2`, and softmax normalization further cancels the (correlated) error. Helps every dtype.

### Headline shape (BT=8192, V=128256, full fwd+bwd)

| GPU | dtype | baseline | this PR | latency | throughput |
|---|---|--:|--:|--:|--:|
| H100 | bf16 | 2.583 ms | 2.225 ms | **-13.9%** | +16.1% |
| H100 | fp32 | 4.489 ms | 4.475 ms | **-0.3%** | +0.3% |
| B200 | bf16 | 1.847 ms | 1.687 ms | **-8.6%** | +9.5% |
| B200 | fp32 | 2.085 ms | 1.887 ms | **-9.5%** | +10.5% |

### Across context length (V=128256, BT 1024–65536)

| GPU | dtype | min | avg | max |
|---|---|--:|--:|--:|
| H100 | bf16 | -10% | -13% | -15% |
| H100 | fp32 | -0% | -0% | -1% |
| B200 | bf16 | -4% | -8% | -10% |
| B200 | fp32 | -5% | -8% | -10% |

### Figures
_Latency vs context length and vs vocab, baseline vs this PR, on H100 and B200._

pr1_bf16_bt
pr1_bf16_vocab
pr1_fp32_bt
pr1_fp32_vocab

PR #1275, opened 2026-06-28
Make LOG2_E a tl.constexpr so the CE kernel compiles on newer Triton

kashif:
Follow up to https://github.com/linkedin/Liger-Kernel/pull/1266  the cross entropy kernel reads the module level `LOG2_E` constant from inside the `@triton.jit` kernel. Newer Triton no longer allows accessing a plain (non-constexpr) global from within a kernel, so the kernel fails to compile with a NameError and the cross entropy correctness tests error out.

Wrapping the constant in `tl.constexpr` is what the Triton error itself suggests, and it keeps the value identical. The cross entropy tests pass again with this change.

### DIFF
--- a/src/liger_kernel/ops/cross_entropy.py
+++ b/src/liger_kernel/ops/cross_entropy.py
@@ -23,7 +23,7 @@ else:
     from triton.language.math import tanh
 
 # log2(e): lets us emit the hardware ex2.approx instruction via e^x = 2^(x * LOG2_E)
-LOG2_E = 1.4426950408889634
+LOG2_E = tl.constexpr(1.4426950408889634)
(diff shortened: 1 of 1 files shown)


kashif:
cc @vaibhavjindal

thad0ctor:
Can confirm this behavior as well, in the meantime it can be bypassed with `TRITON_ALLOW_NON_CONSTEXPR_GLOBALS=1`

vaibhavjindal:
Thanks @kashif for catching this bug! We are trying to fix the broken CI so we catch these bugs early on.

vaibhavjindal (approved):

PR #1276, opened 2026-06-28
[transformers] Add tiled MLP patching via the transformers register_patch_mapping API

### LINKED ISSUE #893: TiledMLP

kashif:
Liger already ships a memory-efficient tiled SwiGLU/GEGLU MLP, but there was no way to reach it through the normal patching path. You had to construct the module by hand, so the activation-memory win was hidden from `apply_liger_kernel_to_*` / `use_liger_kernel=True`.

This wires it in using the official transformers monkey patching API, with one mechanism rather than per-model edits:

- `apply_liger_tiled_mlp(model=None, num_shards=None)`:
  - pre-load (`model=None`): registers the swap via `transformers.monkey_patching.register_patch_mapping`, so `from_pretrained` / `from_config` build the tiled MLP directly. Weights load by name since gate/up/down projections match.
  - instance (`model=...`): walks the model and patches in place every MLP whose class name is in the mapping, reusing existing weights. This covers the already-loaded case the official API cannot retrofit.
- `_apply_liger_kernel_to_instance` gains `tiled_mlp` / `tiled_mlp_num_shards`, so `use_liger_kernel=True` and TRL can opt in uniformly for every supported model.
- `LIGER_TILED_MLP_PATCH_MAPPING` lists 20 MLP classes sharing the SwiGLU or GEGLU layout (llama, granite, mistral, ministral, qwen2/3, qwen2_5_vl, qwen3_5, qwen3_vl, qwen3_vl_moe, smollm3, mllama, pixtral, llama4, olmo2/3, exaone4 -> tiled SwiGLU; gemma/2/3 -> tiled GEGLU).

Still excluded, because the tiled module computes exactly `down_proj(act(gate_proj(x)) * up_proj(x))` and anything else would be dropped: MoE expert stacks (mixtral, gpt_oss), fused gate_up (phi3, glm4), the Qwen vision MLPs (`linear_fc1`/`linear_fc2`), and MLPs carrying an extra scale or dropout (deepseek_v4, falcon_h1). Gemma4's text MLP looks eligible and is left for a follow-up.

## Gradient correctness

The tiled backward drove each shard with its own `torch.autograd.backward`, so every weight reached its `AccumulateGrad` node once per shard. DDP registers one hook per parameter and marks it ready on each firing, so any model with more than one tiled MLP died in backward with `Expected to mark a variable ready only once` (huggingface/kernels-community#1081; DeepSpeed's upstream `TiledMLP` has the same gap, flagged in-source as `XXX: DDP, FSDP will need something similar`).

The weights are now passed to `LigerTiledMLPFunction.apply` as explicit inputs and their gradients returned from `backward`, computed with `torch.autograd.grad` so no `AccumulateGrad` fires inside the shard loop. Each weight is accumulated exactly once. ZeRO-3 keeps the previous per-shard path, selected on `hasattr(param, "ds_id")`, because its partitioned weights are 0-sized placeholders outside the owning module's forward and autograd would record `(0,)` as the expected gradient shape.

Also in this area:

- The incoming gradient was reshaped with the *input* feature dim, which is wrong whenever `fn` changes width. It now uses its own last dim.
- The same fix is applied to the cuTile backend, which carried an independent copy of the tiling logic and the same defect.
- The `num_shards` harmonizing `all_reduce` now runs only when a weight is actually sharded. It exists so ranks agree on the shard count for backends that gather weights inside the recompute; under DDP nothing is gathered, so it was one collective per MLP per step for no benefit. It still runs when the weights are unknown, so an unrecognised backend cannot deadlock.

## Guards

Instance patching rebinds methods onto an already-built module, so `LigerTiledSwiGLUMLP.__init__` never runs and its activation check was dead code on this path. `_patch_tiled_mlp_module` now validates before rebinding anything, and rejects:

- an activation the fused kernel would silently replace (handles `act_fn` and Llama4's `activation_fn`);
- an MLP owning children or parameters beyond gate/up/down + activation, i.e. computation that replacing `forward` would silently drop.

A rejected model is left untouched. This is what makes widening the mapping safe: an entry added in error raises instead of quietly changing numerics. Verified against real modules — `LlamaMLP`/`Gemma3MLP` accepted, `InklingMLP` rejected on `global_scale`, `T5Gemma2MLP` rejected on `dropout`.

## Distributed status

Each weight is accumulated once per backward under DDP and replicated setups; ZeRO-3 instead defers its reduction to the last shard via `ds_grad_is_ready`. DDP is covered by a 2-rank gloo test that needs no GPU. FSDP2 takes the same explicit-input path as DDP and is expected to be correct, but is not re-verified here.

## Tests

Central instance path, standalone instance patch, and the pre-load registry across a silu and a gelu model. Added: per-weight accumulation counts on both the default and ZeRO-3 paths, differing in/out features, an end-to-end DDP run against an untiled reference, both guards, the collective being issued for sharded weights and skipped for replicated ones, and a cuTile-gated mirror of the accumulation tests.

`test_tiled_mlp.py` + `test_monkey_patch.py` + `test_cutile_tiled_mlp.py`: 134 passed, 30 skipped on an H100.

Fixes #893


kashif:
cc @vaibhavjindal

PR #1299, opened 2026-07-14
Add CUTLASS CuTe DSL RMSNorm kernel

Charlesrizzly:
# Add CUTLASS CuTe DSL RMSNorm kernel

This PR adds a genuine **NVIDIA CUTLASS CuTe DSL** (`import cutlass.cute`) RMSNorm
implementation to the existing `cutedsl` backend introduced in #1279. It is opt-in
only and selected with `LIGER_KERNEL_IMPL=cutedsl`; any op without a CuTe DSL kernel
transparently falls back to the default Triton kernel, so selecting the backend is
always safe.

The backend provides forward and backward implementations compatible with the
existing `LigerRMSNormFunction` interface (same public
`rms_norm_forward` / `rms_norm_backward` / autograd `Function` signatures).

## Feature coverage

- FP32 and BF16 (FP16 also covered by tests)
- Llama, Gemma, and no-casting modes
- Configurable weight offset
- Affine and non-affine RMSNorm
- In-place and out-of-place backward
- Regular and non-power-of-two (irregular) hidden dimensions
- Transformer module and functional API routing
- Defensive DTensor detection for compatible Torch versions

## This PR also adds

- RMSNorm registration on the existing CuTe DSL export surface (extends #1279;
  cross-entropy exports are preserved)
- A Triton-vs-CuTe DSL benchmark runner for RMSNorm (`run_cutedsl_compare.py`)
- Subprocess-isolated backend selection so Triton and CuTe DSL registration state
  can't leak between providers
- Per-provider benchmark CSV routing (reuses the repo's existing
  `LIGER_BENCH_TARGET` / `LIGER_BENCH_PROVIDER_TAG` hooks)
- A small Torch-compatibility fix to the upstream RMSNorm DTensor check (guards
  `torch.distributed.tensor.DTensor` on Torch builds that don't eagerly import that
  submodule)

> **Companion PR (separate):** grouped-bar benchmark visualization, source-specific
> plot filenames, and one-command generation of all metric/sweep plots live in a
> separate visualizer PR to keep this kernel PR narrowly scoped and reviewable.

## Kernel design (genuine `cutlass.cute`)

The kernels are written for correctness and clarity first (fp32 accumulation; one CTA
per row for the reductions; the weight gradient reduced in parallel across row strips)
and keep the hidden dimension fully dynamic, so a single compiled kernel serves every
shape — regular or irregular.

### Forward
- One CTA (thread block) per input row; 8 warps (256 threads) cooperate on the row.
- Two-pass algorithm:
  1. Sum of squares in FP32 → per-warp butterfly-shuffle reduction → cross-warp
     reduction via shared memory → `rstd = rsqrt(mean_square + eps)`, cached to an
     FP32 `RSTD` buffer for reuse in backward.
  2. Normalize (`x * rstd`) and apply the optional affine weight (`offset + w`).
- Per-column tail is bounds-guarded, so irregular hidden dimensions are supported
  without a separate kernel.
- Casting mode and the affine flag are compile-time (`cutlass.Constexpr`)
  specializations. Llama round-trips the normalized value to the input dtype before
  the affine multiply; Gemma and no-casting stay in FP32 — matching Triton semantics.

### Backward
- **dX kernel:** one CTA per row. Computes `dot(m, x)` with `m = dy * (w + offset)`
  via the same warp + cross-warp reduction, then
  `dx = rstd * (m - (1/N) * rstd^2 * dot(m, x) * x)`. Supports reusing `dY` storage
  for `dX` when `in_place=True` (each column is read then written by its single owning
  thread, fenced by the reduction barrier).
- **dW kernel:** the weight-gradient reduction is parallelized across row strips with a
  2-D grid (column-blocks × row-strips). Each strip (one per SM, capped at `n_rows`)
  reduces `dy * (x * rstd)` over just its slice of rows in FP32 and writes a partial
  `dW` row; the host sums the partials — the same strategy as the Triton backward's
  `sm_count` partials + `_dW.sum(0)`. `dW` is computed **before** `dX` so the in-place
  `dX` write can't clobber the `dY` the reduction reads. Skipped entirely for non-affine
  RMSNorm.

### Host
- Torch tensors are marshaled via DLPack; kernels are compiled once with
  `cute.compile` and cached, keyed on **every** baked dtype (X, W, dY) plus casting
  mode and the affine flag, then launched on Torch's current CUDA stream.

## Correctness

Ran the CuTe DSL routing and RMSNorm correctness suites on an NVIDIA GPU with
`nvidia-cutlass-dsl` installed:

```bash
LIGER_KERNEL_IMPL=cutedsl pytest test/transformers/test_cutedsl_rms_norm.py -v
LIGER_KERNEL_IMPL=cutedsl pytest test/transformers/test_cutedsl_backend.py -v
pytest test/transformers/test_cutedsl_cross_entropy.py -v   # upstream regression guard
make checkstyle
```

Result: **158 passed, 1 skipped.** The single skip is expected — it is the
cross-entropy backend-routing test that only runs when `LIGER_KERNEL_IMPL=cutedsl`
is set, which is intentionally not set for the cross-entropy regression-guard command
above. `make checkstyle` (ruff check + format) passes clean.

Coverage includes:

- Explicit `LIGER_KERNEL_IMPL=cutedsl` routing validation (ops + transformer modules)
- Forward parity against the default Triton RMSNorm reference
- Input-gradient parity
- Affine weight-gradient parity
- FP32 and BF16 (FP16 in the parity matrix)
- Llama, Gemma, and no-casting modes
- Affine and non-affine configurations
- In-place and out-of-place backward
- Standard and irregular shapes (representative regular dims plus non-power-of-two
  hidden dimensions)
- A same-process mixed weight-dtype case that guards the compiled-kernel cache key
- Upstream CuTe DSL cross-entropy tests preserved as a backend export/import
  regression guard

## Benchmarks: CuTe DSL vs Triton

Measurements use the median (`y_value_50`) from `triton.testing.do_bench`.
Ratio = Triton / CuTe DSL; **values below 1.00x mean CuTe DSL is slower than Triton.**
`full` = forward plus backward.

**Benchmark environment**

- GPU: NVIDIA H100 80GB HBM3 and NVIDIA B200
- Torch: 2.12.0.1+cu130
- CUDA runtime: 13.0
- Triton: 3.7.0
- nvidia-cutlass-dsl: 4.5.2
- Input dtype: BF16

The token-length sweep uses B=1, H=4096, and T = {1024, 2048, 4096, 8192}.
The model-configuration sweep uses B=1, T=2048, and the hidden dimension from each
model configuration.

> **Honest performance note.** This is a correctness-first implementation (fp32
> accumulation, one CTA per row for the reductions, no vectorization/pipelining). The
> weight-gradient (dW) reduction is parallelized across row strips (one strip per SM,
> partials summed on the host — the same strategy as the Triton backward). The three-way
> picture (H100 and B200 are the same shape):
>
> - **vs HuggingFace eager:** CuTe DSL is **faster in the forward pass in every case**
>   (11/11 on both GPUs), and faster on the **full pass in 9/11** cases on both GPUs and
>   the **backward in 9/11 (H100) / 11/11 (B200)** — a clear win over eager. The only
>   losses to HuggingFace are at the smallest size (T=1024), where the op is tiny and
>   launch-bound.
> - **vs the existing Triton Liger kernel:** CuTe DSL is **not yet faster in any measured
>   case** (~0.5–0.8x), but after the parallel-dW backward it is now **stable across
>   sequence length** rather than collapsing — full-pass ratio holds around 0.54–0.57x
>   as T grows from 1k to 8k (it was 0.59x→0.09x before the optimization).
> - **Peak memory:** identical to Triton to the MB, and ~4–5x below HuggingFace.
>
> Remaining headroom vs Triton (fusing dX/dW to load x/dy once, vectorized loads) is
> deferred to a `liger-kernel-perf` follow-up. The `Ratio` columns below are
> Triton/CuTeDSL, so values below 1.00x mean CuTe DSL is slower than Triton specifically.

### NVIDIA H100 80GB HBM3

Full pass (fwd+bwd) — token-length sweep (B=1, H=4096)

| T | CuTe DSL (ms) | Triton (ms) | Ratio (Triton/CuTeDSL) |
|---|---|---|---|
| 1,024 | 0.4688 | 0.2554 | 0.54x |
| 2,048 | 0.4641 | 0.2589 | 0.56x |
| 4,096 | 0.4686 | 0.2549 | 0.54x |
| 8,192 | 0.4762 | 0.2604 | 0.55x |

Full pass (fwd+bwd) — model-config sweep (B=1, T=2048)

| Model | CuTe DSL (ms) | Triton (ms) | Ratio (Triton/CuTeDSL) |
|---|---|---|---|
| Llama 2 7B | 0.3825 | 0.3155 | 0.82x |
| Llama 3 8B | 0.3867 | 0.3122 | 0.81x |
| Qwen 2.5 7B | 0.3842 | 0.3141 | 0.82x |
| Qwen 2.5 14B | 0.3858 | 0.3160 | 0.82x |
| Qwen 2.5 72B | 0.3828 | 0.3135 | 0.82x |
| DeepSeek V2 Lite | 0.3791 | 0.3137 | 0.83x |
| DeepSeek V3 | 0.3822 | 0.2577 | 0.67x |

Geometric-mean ratio (Triton/CuTeDSL), H100: forward 0.47x; backward 0.49x; full 0.70x.
The parallel-dW backward lifted the full-pass geomean from ~0.27x (token) to 0.70x and
made it flat across sequence length. CuTe DSL beats HuggingFace eager in 11/11 forward,
9/11 backward, and 9/11 full-pass cases.

Peak memory — full pass, token-length sweep (MB)

| T | CuTe DSL | Triton | HuggingFace |
|---|---|---|---|
| 1,024 | 34.1 | 34.1 | 152.0 |
| 2,048 | 66.1 | 66.1 | 304.0 |
| 4,096 | 130.1 | 130.1 | 608.0 |
| 8,192 | 258.1 | 258.1 | 1216.1 |

CuTe DSL peak memory is identical to Triton (the dW strip partials match Triton's
`_dW` buffer) and ~4–5x below HuggingFace eager across both sweeps.

### NVIDIA B200

Full pass (fwd+bwd) — token-length sweep (B=1, H=4096)

| T | CuTe DSL (ms) | Triton (ms) | Ratio (Triton/CuTeDSL) |
|---|---|---|---|
| 1,024 | 0.2859 | 0.1598 | 0.56x |
| 2,048 | 0.2846 | 0.1593 | 0.56x |
| 4,096 | 0.2840 | 0.1574 | 0.55x |
| 8,192 | 0.3298 | 0.1894 | 0.57x |

Full pass (fwd+bwd) — model-config sweep (B=1, T=2048)

| Model | CuTe DSL (ms) | Triton (ms) | Ratio (Triton/CuTeDSL) |
|---|---|---|---|
| Llama 2 7B | 0.2867 | 0.1635 | 0.57x |
| Llama 3 8B | 0.2870 | 0.1615 | 0.56x |
| Qwen 2.5 7B | 0.2864 | 0.1629 | 0.57x |
| Qwen 2.5 14B | 0.2839 | 0.1621 | 0.57x |
| Qwen 2.5 72B | 0.2857 | 0.1589 | 0.56x |
| DeepSeek V2 Lite | 0.2857 | 0.1621 | 0.57x |
| DeepSeek V3 | 0.2868 | 0.1613 | 0.56x |

Geometric-mean ratio (Triton/CuTeDSL), B200: forward 0.53x; backward 0.41x; full 0.56x.
Same shape as H100 and just as stable across sequence length after the parallel-dW
backward. CuTe DSL beats HuggingFace eager in 11/11 forward, 11/11 backward, and 9/11
full-pass cases.

Peak memory — full pass, token-length sweep (MB)

| T | CuTe DSL | Triton | HuggingFace |
|---|---|---|---|
| 1,024 | 34.4 | 34.4 | 152.0 |
| 2,048 | 66.4 | 66.4 | 304.0 |
| 4,096 | 130.4 | 130.4 | 608.0 |
| 8,192 | 258.4 | 258.4 | 1216.1 |

As on H100, CuTe DSL peak memory is identical to Triton and ~4–5x below HuggingFace
eager across both sweeps.

## Usage

Install the CuTe DSL dependency and enable the backend:

```bash
pip install "liger-kernel[cutedsl]"        # nvidia-cutlass-dsl
LIGER_KERNEL_IMPL=cutedsl python your_script.py
```

Run both Triton and CuTe DSL RMSNorm benchmarks into one CSV per GPU:

```bash
cd benchmark/scripts
python run_cutedsl_compare.py --kernel rms_norm --source h100 --overwrite
python run_cutedsl_compare.py --kernel rms_norm --source h100 --sweep-mode model_config --overwrite
# (repeat on the B200 node with --source b200)
```

Generate the plots (grouped bars — companion visualizer PR):

```bash
cd benchmark
python benchmarks_visualizer.py --kernel-name rms_norm --metric-name speed \
  --sweep-mode token_length --data-file data/all_benchmark_data_cutedsl_rms_norm_h100.csv --overwrite
```

## Known limitations

- `row_mode` is accepted for interface compatibility but is not currently specialized
  by the CuTe DSL implementation.
- The kernels are correctness-first (FP32 accumulation, one CTA per row for the
  reductions, strip-parallel dW). Remaining on-GPU tuning (fusing dX/dW to load x/dy
  once, vectorized/pipelined loads) is left to a follow-up using the `liger-kernel-perf`
  workflow.
- No-casting mode accumulates in FP32 (vs the Triton kernel's in-dtype path); this is
  an intentional, tolerance-covered deviation.
- **Performance does not yet match the existing Triton kernel.** On both H100 and B200,
  CuTe DSL is ~0.5–0.8x of Triton on the full pass — slower, but now **stable across
  sequence length** (it no longer collapses at large T) thanks to the strip-parallel dW
  backward. It beats HuggingFace eager in the forward pass everywhere and on the full
  pass in 9/11 cases on both GPUs; peak memory is identical to Triton. Closing the
  remaining gap to Triton (fused dX/dW, vectorized loads) is a `liger-kernel-perf`
  follow-up.

## Branch state

- Branched directly from the latest `upstream/main`; no old prototype commits were
  cherry-picked.
- No `import cuda.tile` anywhere under `src/liger_kernel/ops/cutedsl/`; the RMSNorm
  kernel uses `cutlass.cute`.
- The existing CuTe DSL cross-entropy implementation and the upstream cuTile files are
  unchanged.
- No tracked benchmark CSV is added or deleted; the genuine H100/B200 result CSVs are
  git-ignored and are not part of this PR.
- Files changed vs `upstream/main`: `README.md`,
  `benchmark/scripts/run_cutedsl_compare.py`,
  `src/liger_kernel/ops/cutedsl/ops/__init__.py`,
  `src/liger_kernel/ops/cutedsl/ops/rms_norm.py`,
  `src/liger_kernel/ops/rms_norm.py`,
  `test/transformers/test_cutedsl_backend.py`,
  `test/transformers/test_cutedsl_rms_norm.py`.


vaibhavjindal (approved):
LGTM! Thanks for making this change. 

This can serve as the initial cutedsl kernel that we can further improve upon to match or exceed triton performance!

PR #1305, opened 2026-07-19
fix(test): prevent in-place RMSNorm backward from clobbering shared grad buffers

yueyiming2009:
## Summary

`LigerRMSNormFunction` with `in_place=True` (the default) writes `dX` into the grad-output buffer it receives during backward. Two RMSNorm tests passed the same grad tensor to multiple backward calls, which undermined what they assert:

- **`test_correctness`**: `triton_o.backward(do)` corrupts `do`, so the test was only correct because the reference backward happened to run first. The Liger backward now receives `do.clone()`, making the test order-independent.
- **`test_correctness_functional`**: `y1.backward(grad)` clobbered `grad` before `y2.backward(grad)` ran — and since the in-place backward returns a view of the buffer it receives, `h1.grad`/`h2.grad` can alias the same storage, making the final `allclose(h1.grad, h2.grad)` comparison vacuous (it may compare a buffer against itself). Each backward now receives its own clone, so the assertion compares two independently computed gradients.

Also drops the `retain_graph=True` flags in `test_correctness` (left over from an earlier version of the test — each graph is only backward-ed once), and fixes a comment typo in `src/liger_kernel/ops/rms_norm.py` ("bacward" → "backward").

No functional kernel/production code is changed (the only non-test change is a comment typo); this PR makes the existing tests actually verify what they intend to.

## Testing Done

- Hardware Type: <BLANK> (no GPU available locally; semantic review only — CI should exercise the GPU path)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vaibhavjindal (approved):
LGTM! Thanks for catching this!

PR #1316, opened 2026-07-23
Guard the FLCE addmm fast path on compute capability 8.0 or higher

albertvillanova:
This PR fixes a crash in the fused linear cross-entropy backward pass on pre-Ampere GPUs (CUDA compute capability < 8.0).

### Motivation

The `grad_weight` accumulation optimization added in #1239 uses `torch.addmm(..., out_dtype=torch.float32)` with fp16/bf16 inputs. cuBLAS only supports this low-precision-input/fp32-output gemm on sm_80+ (Ampere and newer). On older GPUs (e.g. Tesla T4 at sm_75, V100 at sm_70) the call raises:

```
RuntimeError: gemm input type at::BFloat16 and output type float is only supported for CUDA devices with compute capability 8.0 or higher
```

This is easily hit under AMP, where the model forward runs in bf16 (so `grad_logits` is bf16) while the weights stay fp32, satisfying every condition of the existing guard except the hardware requirement, which was not checked. 

Reported downstream in huggingface/trl, where it broke CI after the 0.8.1 release:
- huggingface/trl#6516

### Solution

Add a compute-capability check to the guard so the `addmm` fast path is only taken on sm_80+. On older GPUs execution falls back to the pre-existing `torch.mm(...).float()` path, which is always supported.

### Changes

- Guard the FLCE `grad_weight` `addmm` fast path on the device's compute capability being 8.0 or higher, falling back to `torch.mm` otherwise.


vaibhavjindal (approved):
Thanks for catching and fixing this!

PR #1322, opened 2026-07-24
feat: Add accum dtype support to fused linear JSD

KyleMylonakisProtopia:
## Summary

Adds `accum_dtype` support to `LigerFusedLinearJSD`, matching the existing fused linear Cross Entropy accumulation pattern. When `accum_dtype` is provided, the fused JSD path accumulates the student weight gradient in the requested dtype, including the guarded CUDA `torch.addmm(..., out_dtype=torch.float32)` path where supported, then casts the accumulated gradient back to the original weight dtype before returning.

This preserves the existing default behavior when `accum_dtype=None` while allowing callers to opt into higher-precision weight-gradient accumulation for improved numerical stability.

## Testing Done

Validated the modified files with Python bytecode compilation and editor diagnostics:

```bash
python3 -m py_compile \
  fused_linear_jsd.py \
  fused_linear_jsd.py \
  functional.py \
  test_fused_linear_jsd.py
```

Added new unit tests via parameterization for the new functionality.

Will update the checkboxes below as the results come in, then bring out of draft.

- Hardware Type: 8x NVIDIA H100 80GB HBM3
- [x] run `make test` to ensure correctness (Unrelated failures)

### DIFF
--- a/src/liger_kernel/ops/fused_linear_jsd.py
+++ b/src/liger_kernel/ops/fused_linear_jsd.py
@@ -3,6 +3,8 @@ from typing import Optional
 import torch
 import triton
 
+from packaging.version import Version
+
 from liger_kernel.ops.jsd import _jsd_kernel
 from liger_kernel.ops.utils import amp_custom_bwd
 from liger_kernel.ops.utils import amp_custom_fwd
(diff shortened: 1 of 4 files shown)


vaibhavjindal (approved):
LGTM, Thanks for this change!

PR #1324, opened 2026-07-25
[Perf] Use direct FLCE weight-gradient accumulation for FP16, BF16, and FP32

### LINKED ISSUE #1323: Reduce FLCE weight-gradient allocation overhead across FP16, BF16, and FP32
# 🐛 Describe the bug field

## Background and motivation

Fused linear cross entropy reduces the cost of language-model training by avoiding a single full `[tokens, vocabulary]` logits tensor. Liger FLCE does this by processing the linear projection and cross entropy in token chunks.

PyTorch 2.13 now provides a native chunked `linear_cross_entropy` implementation for FP16, BF16, and FP32. That gives Liger a useful framework-level comparison point: Liger FLCE should remain at least competitive with native PyTorch LCE in both execution time and memory use across the same training dtypes.

While comparing the two implementations, I found an avoidable allocation in Liger's weight-gradient update. Liger chunks the logits, but its fallback weight-gradient expression first creates a complete `[vocabulary, hidden_size]` matrix for each chunk before adding that contribution to the persistent weight-gradient buffer:

```python
grad_weight += torch.mm(grad_logits_chunk.t(), input_chunk).float()
```

The size of this intermediate is determined by the model dimensions `V` and `H`, not by the number of tokens in the chunk. Therefore, logits chunking does not reduce this part of the peak memory or the associated allocation and memory-traffic cost.

For FP16 and BF16, `torch.mm` creates a low-precision `[V, H]` result and `.float()` creates an additional FP32 copy. For FP32, the cast is a no-op, but the separate `[V, H]` matrix-multiplication result is still created.

At `V=32000` and `H=4096`, the avoidable intermediates are:

| Dtype | Matrix-multiplication result | Additional FP32 conversion | Total avoidable allocation |
|---|---:|---:|---:|
| BF16 | 250 MiB | 500 MiB | 750 MiB |
| FP16 | 250 MiB | 500 MiB | 750 MiB |
| FP32 | 500 MiB | none | 500 MiB |

This is a general FLCE implementation cost. Larger GPUs can accommodate larger allocations, but the allocation and memory traffic are still unnecessary and scale with `V * H`.

## Proposed direction

When `grad_weight`, `grad_logits`, and the input chunk have the same native training dtype, Liger can accumulate the matrix product directly into the existing destination:

```python
grad_weight.addmm_(grad_logits_chunk.t(), input_chunk)
```

The proposed support matrix is:

| Configuration | Accumulation path |
|---|---|
| Same-dtype FP16 on NVIDIA CUDA | Direct `addmm_` |
| Same-dtype BF16 on NVIDIA CUDA, compute capability 8.0+ | Direct `addmm_` |
| Same-dtype FP32 on NVIDIA CUDA | Direct `addmm_` |
| FP16/BF16 operands with an FP32 destination | Existing `addmm(..., out_dtype=torch.float32, out=...)` path |
| Other devices, dtypes, or unsupported combinations | Existing fallback |

This changes only how each chunk's weight-gradient contribution is accumulated. It does not change the FLCE API, token chunking, loss calculation, or returned gradient dtype.

## Comparison with PyTorch 2.13 native LCE

The following end-to-end benchmark measures the complete forward and backward training operation at `N=16384`, `H=4096`, and `V=32000`. Providers use identical tensors, run in an interleaved seeded order, and are timed with CUDA events after three warm-up rounds. Each latency result contains 10 samples; memory contains two isolated samples.

`torch-lce-auto` is PyTorch 2.13 native LCE with automatic chunking. `torch-lce-512` is the same native implementation with a fixed 512-token chunk. `liger` is Liger FLCE with the proposed direct accumulation.

| Dtype | Provider | Tokens per chunk | p20 / median / p80 | Throughput | Peak allocated | Peak increase during operation |
|---|---|---:|---:|---:|---:|---:|
| BF16 | PyTorch LCE auto | 1,024 | 1,020.31 / 1,026.39 / 1,029.58 ms | 15,962.8 tokens/s | 1,150.38 MiB | 756.00 MiB |
| BF16 | PyTorch LCE 512 | 512 | 1,050.54 / 1,057.01 / 1,058.44 ms | 15,500.3 tokens/s | 1,150.38 MiB | 756.00 MiB |
| BF16 | Liger FLCE | 2,048 | 892.08 / 897.07 / 900.65 ms | 18,263.8 tokens/s | 1,024.45 MiB | 630.08 MiB |
| FP16 | PyTorch LCE auto | 1,024 | 1,011.01 / 1,012.98 / 1,014.89 ms | 16,174.1 tokens/s | 1,150.38 MiB | 756.00 MiB |
| FP16 | PyTorch LCE 512 | 512 | 1,065.28 / 1,066.65 / 1,068.67 ms | 15,360.3 tokens/s | 1,150.38 MiB | 756.00 MiB |
| FP16 | Liger FLCE | 2,048 | 897.90 / 899.94 / 901.24 ms | 18,205.7 tokens/s | 1,024.45 MiB | 630.08 MiB |
| FP32 | PyTorch LCE auto | 1,024 | 3,313.70 / 3,317.26 / 3,324.13 ms | 4,939.0 tokens/s | 2,284.38 MiB | 1,512.00 MiB |
| FP32 | PyTorch LCE 512 | 512 | 3,329.99 / 3,332.67 / 3,338.46 ms | 4,916.2 tokens/s | 2,284.38 MiB | 1,512.00 MiB |
| FP32 | Liger FLCE | 2,048 | 3,451.62 / 3,454.83 / 3,462.58 ms | 4,742.3 tokens/s | 2,028.45 MiB | 1,256.08 MiB |

Compared with PyTorch LCE auto at this shape:

- BF16 Liger has 12.6% lower median latency and 10.9% lower peak allocated memory.
- FP16 Liger has 11.2% lower median latency and 10.9% lower peak allocated memory.
- FP32 Liger has 11.2% lower peak allocated memory, while its median latency is 4.1% higher. The remaining FP32 latency difference is separate follow-up optimization work; this change improves the accumulation path without claiming that every dtype is already faster.

The correctness gate compared loss, input gradients, and weight gradients for ordinary PyTorch, both PyTorch LCE chunking modes, and Liger in BF16, FP16, and FP32. All comparisons passed their dtype-specific tolerances. The complete Liger FLCE test file also passes (`140 passed`), including direct-path regression tests for all three dtypes.

## Relationship to earlier work

This follows issue #1232 and PR #1239, which optimized FP16/BF16 operands accumulating into an FP32 destination. The proposed change covers the separate same-dtype FP16, BF16, and FP32 paths.

Issue #533 discusses the lifetime of the required persistent weight-gradient buffer. This proposal instead removes intermediate `[V, H]` results that do not need to exist.

# Reproduce field

The allocation pattern can be inspected independently of total GPU capacity. This example measures the extra peak allocation of the existing same-dtype update at `V=32000`, `H=4096`, and a 512-token chunk:

```python
import torch


N, H, V = 512, 4096, 32000
dtype = torch.float16  # also test torch.bfloat16 and torch.float32

grad_weight = torch.zeros(V, H, device="cuda", dtype=dtype)
grad_logits = torch.randn(N, V, device="cuda", dtype=dtype)
input_chunk = torch.randn(N, H, device="cuda", dtype=dtype)

torch.cuda.synchronize()
torch.cuda.reset_peak_memory_stats()
baseline = torch.cuda.memory_allocated()

grad_weight += torch.mm(grad_logits.t(), input_chunk).float()

torch.cuda.synchronize()
extra_mib = (torch.cuda.max_memory_allocated() - baseline) / 1024**2
print(f"extra peak allocation: {extra_mib:.2f} MiB")
```

The direct form to compare is:

```python
grad_weight.addmm_(grad_logits.t(), input_chunk)
```

# Versions field

- Liger Kernel baseline: `main` at `ed08f6ea75e388a63b7493b5bc97042b1f142712`
- Python: `3.14.4`
- PyTorch: `2.13.0+cu130`
- Triton: `3.7.1`
- Transformers: `5.14.1`
- CUDA: `13.0`
- Benchmark hardware: NVIDIA GeForce RTX 3050 Laptop GPU, compute capability 8.6

Anubhav-2003:
## Summary

This PR reduces allocation and memory-traffic overhead in Liger fused linear cross entropy by accumulating same-dtype FP16, BF16, and FP32 weight-gradient contributions directly into the existing `grad_weight` buffer on supported NVIDIA CUDA devices.

The broader goal is to keep Liger FLCE competitive with PyTorch 2.13 native `linear_cross_entropy` across the dtypes used for training. Both implementations avoid retaining a full `[tokens, vocabulary]` logits tensor, but Liger's fallback weight-gradient update still materialized a complete `[vocabulary, hidden_size]` matrix before adding each chunk's contribution to the persistent gradient. Direct accumulation removes that intermediate without changing the API, chunking policy, loss calculation, or returned gradient dtype.

The existing mixed-precision path for FP16/BF16 operands with an FP32 destination is preserved. Unsupported devices and dtype combinations continue to use the existing fallback.

Fixes #1323 .

## Details

The previous same-dtype update used:

```python
grad_weight += torch.mm(grad_logits.t(), input_chunk).float()
```

This creates a separate `[V, H]` matrix-multiplication result. FP16 and BF16 then create an additional FP32 conversion of that result. At `V=32000` and `H=4096`, those intermediates account for 750 MiB in FP16/BF16 and 500 MiB in FP32, independently of the token chunk size.

This PR introduces a shared accumulation helper and selects direct accumulation when all three operands use the same supported dtype:

```python
grad_weight.addmm_(grad_logits.t(), input_chunk)
```

### Dispatch behavior

| Configuration | Behavior after this PR |
|---|---|
| Same-dtype FP16 on NVIDIA CUDA | Direct accumulation into `grad_weight` |
| Same-dtype BF16 on NVIDIA CUDA, compute capability 8.0+ | Direct accumulation into `grad_weight` |
| Same-dtype FP32 on NVIDIA CUDA | Direct accumulation into `grad_weight` |
| FP16/BF16 operands with an FP32 destination | Preserve existing `out_dtype=torch.float32` path |
| ROCm, older BF16 hardware, other dtypes, or mixed unsupported operands | Preserve existing fallback |

The BF16 compute-capability check remains BF16-specific. FP16 and FP32 use the same CUDA operand types as the previously supported matrix-multiplication fallback and do not inherit the BF16 hardware restriction.

The new regression test parametrizes FP16, BF16, and FP32 and deliberately fails if a supported same-dtype case reaches the allocating `torch.mm` fallback.

### Methodology

- Complete forward and backward training operation; forward-only timings are not compared.
- Shape: `N=16384`, `H=4096`, `V=32000`.
- Identical inputs, weights, and targets for every provider in a dtype.
- Providers interleaved in a seeded random order.
- Three warm-up rounds and 10 CUDA-event timing samples per provider.
- Two isolated peak-memory samples per provider.
- Correctness gate for loss, input gradient, and weight gradient before timing.
- PyTorch `2.13.0+cu130`; Liger candidate applied to baseline `ed08f6e`.

`torch-lce-auto` is PyTorch 2.13 native LCE with automatic chunking. `torch-lce-512` uses the same native implementation with a fixed 512-token chunk.

### Full all-dtype comparison

| Dtype | Provider | Tokens per chunk | p20 / median / p80 | Throughput | Peak allocated | Peak increase during operation |
|---|---|---:|---:|---:|---:|---:|
| BF16 | PyTorch LCE auto | 1,024 | 1,020.31 / 1,026.39 / 1,029.58 ms | 15,962.8 tokens/s | 1,150.38 MiB | 756.00 MiB |
| BF16 | PyTorch LCE 512 | 512 | 1,050.54 / 1,057.01 / 1,058.44 ms | 15,500.3 tokens/s | 1,150.38 MiB | 756.00 MiB |
| BF16 | Liger FLCE | 2,048 | 892.08 / 897.07 / 900.65 ms | 18,263.8 tokens/s | 1,024.45 MiB | 630.08 MiB |
| FP16 | PyTorch LCE auto | 1,024 | 1,011.01 / 1,012.98 / 1,014.89 ms | 16,174.1 tokens/s | 1,150.38 MiB | 756.00 MiB |
| FP16 | PyTorch LCE 512 | 512 | 1,065.28 / 1,066.65 / 1,068.67 ms | 15,360.3 tokens/s | 1,150.38 MiB | 756.00 MiB |
| FP16 | Liger FLCE | 2,048 | 897.90 / 899.94 / 901.24 ms | 18,205.7 tokens/s | 1,024.45 MiB | 630.08 MiB |
| FP32 | PyTorch LCE auto | 1,024 | 3,313.70 / 3,317.26 / 3,324.13 ms | 4,939.0 tokens/s | 2,284.38 MiB | 1,512.00 MiB |
| FP32 | PyTorch LCE 512 | 512 | 3,329.99 / 3,332.67 / 3,338.46 ms | 4,916.2 tokens/s | 2,284.38 MiB | 1,512.00 MiB |
| FP32 | Liger FLCE | 2,048 | 3,451.62 / 3,454.83 / 3,462.58 ms | 4,742.3 tokens/s | 2,028.45 MiB | 1,256.08 MiB |

### Comparison with PyTorch LCE auto

| Dtype | Liger median-latency difference | Liger peak-memory difference | Interpretation |
|---|---:|---:|---|
| BF16 | 12.6% lower | 10.9% lower | Liger is faster and uses less peak memory |
| FP16 | 11.2% lower | 10.9% lower | Liger is faster and uses less peak memory |
| FP32 | 4.1% higher | 11.2% lower | Liger uses less memory; FP32 latency remains follow-up work |

These results do not treat one GPU's capacity as the motivation for the change. The optimization removes an allocation proportional to `V * H`; the benchmark hardware is reported only so that the measurements can be reproduced and compared fairly.

### Focused accumulation result

At `V=32000`, `H=4096`, and a 512-token chunk, extending direct accumulation beyond the BF16-only branch produced:

| Dtype | Before | After | Before temporary | After temporary |
|---|---:|---:|---:|---:|
| BF16 control | 12.726 ms | 12.738 ms | 0 MiB | 0 MiB |
| FP16 | 20.774 ms | 10.262 ms | 750 MiB | 0 MiB |
| FP32 | 35.277 ms | 28.643 ms | 500 MiB | 0 MiB |

BF16 remains within measurement noise, while FP16 and FP32 remove the full-size temporary and reduce accumulation latency.

### Correctness

The standalone gate compared ordinary PyTorch, PyTorch LCE auto, PyTorch LCE with 512-token chunks, and Liger. All loss, input-gradient, and weight-gradient comparisons passed in BF16, FP16, and FP32.

| Dtype | Absolute tolerance | Relative tolerance | Maximum loss error | Maximum input-gradient error | Maximum weight-gradient error |
|---|---:|---:|---:|---:|---:|
| BF16 | 0.005 | 0.05 | 0.25 | 3.74e-4 | 3.05e-4 |
| FP16 | 0.005 | 0.05 | 1.80e-3 | 7.63e-6 | 1.53e-5 |
| FP32 | 1e-5 | 5e-4 | 3.81e-6 | 6.05e-8 | 3.73e-8 |

The BF16 loss comparison passes through the relative-tolerance term; its gradient errors remain below `4e-4`.

## Testing Done

- Hardware Type: NVIDIA GeForce RTX 3050 Laptop GPU, compute capability 8.6

Additional validation completed:

- [x] Full FLCE test file: `python -m pytest test/transformers/test_fused_linear_cross_entropy.py -x -q` (`140 passed`)
- [x] Direct-accumulation branch tests for FP16, BF16, and FP32 (`3 passed`)
- [x] Repository-wide `ruff check .`
- [x] Repository-wide `ruff format --check .`
- [x] Standalone all-dtype correctness gate across four providers
- [x] Interleaved PyTorch 2.13 LCE versus Liger latency and memory benchmark

### DIFF
--- a/src/liger_kernel/ops/fused_linear_cross_entropy.py
+++ b/src/liger_kernel/ops/fused_linear_cross_entropy.py
@@ -213,12 +213,19 @@ def fused_linear_cross_entropy_forward(
 
         if grad_weight is not None and input_requires_grad:
             grad_logits_t = grad_logits_chunk.t()
+            same_dtype = grad_weight.dtype == grad_logits_t.dtype
+            reduced_to_fp32 = grad_weight.dtype == torch.float32 and grad_logits_t.dtype in (
+                torch.float16,
+                torch.bfloat16,
+            )
+            requires_ampere = grad_logits_t.dtype == torch.bfloat16 or reduced_to_fp32
             if (
                 _ADDMM_SUPPORTS_OUT_DTYPE
                 and grad_weight.device.type == "cuda"
-                and torch.cuda.get_device_capability(grad_weight.device)[0] >= 8
-                and grad_weight.dtype == torch.float32
-                and grad_logits_t.dtype in (torch.float16, torch.bfloat16)
+                and not is_hip()
+                and grad_weight.dtype in (torch.float16, torch.bfloat16, torch.float32)
+                and (same_dtype or reduced_to_fp32)
+                and (not requires_ampere or torch.cuda.get_device_capability(grad_weight.device)[0] >= 8)
             ):
                 # Unlike torch.mm, torch.addmm's out_dtype path does not participate in
                 # autocast operand casting, so under AMP (fp32 params, no bias) _input_chunk
@@ -231,11 +238,11 @@ def fused_linear_cross_entropy_forward(
                     grad_weight,
                     grad_logits_t,
                     input_chunk,
-                    out_dtype=torch.float32,
+                    out_dtype=grad_weight.dtype,
                     out=grad_weight,
                 )
             else:
-                grad_weight += torch.mm(grad_logits_chunk.t(), _input_chunk).float()
+                grad_weight.add_(torch.mm(grad_logits_t, _input_chunk))
 
         if bias is not None and input_requires_grad:
             torch.add(
--- a/test/transformers/test_fused_linear_cross_entropy.py
+++ b/test/transformers/test_fused_linear_cross_entropy.py
@@ -7,6 +7,8 @@ from test.transformers.test_cross_entropy import CrossEntropyWithZLoss
 from test.utils import assert_verbose_allclose
 from test.utils import set_seed
 
+import liger_kernel.ops.fused_linear_cross_entropy as fused_linear_cross_entropy
+
 from liger_kernel.ops import LigerFusedLinearCrossEntropyFunction
 from liger_kernel.transformers.functional import CrossEntropyOutput
 from liger_kernel.transformers.functional import liger_fused_linear_cross_entropy
@@ -98,6 +100,84 @@ class LigerLMHeadCE(torch.nn.Module):
 
     def forward(self, x, y):
         return self.ce_loss(self.lin.weight, x, y, self.lin.bias)
+
+
+def _run_grad_weight_accumulation(operand_dtype, accum_dtype):
+    N, H, V = 11, 41, 37
+    _input = torch.randn(N, H, device=device, dtype=operand_dtype, requires_grad=True)
+    weight = torch.randn(V, H, device=device, dtype=operand_dtype, requires_grad=True)
+    target = torch.randint(0, V, (N,), device=device)
+
+    ref_input = _input.detach().clone().requires_grad_(True)
+    ref_weight = weight.detach().clone().requires_grad_(True)
+    ref_loss = torch.nn.functional.cross_entropy(ref_input @ ref_weight.t(), target)
+    ref_loss.backward()
+
+    _, _, _, _, _, grad_weight, _ = fused_linear_cross_entropy.fused_linear_cross_entropy_forward(
+        _input,
+        weight,
+        target,
+        accum_dtype=accum_dtype,
+    )
+    return grad_weight, ref_weight.grad
+
+
+@pytest.mark.skipif(
+    not fused_linear_cross_entropy._ADDMM_SUPPORTS_OUT_DTYPE or device != "cuda" or torch.version.hip is not None,
+    reason="grad_weight addmm out_dtype requires PyTorch 2.8 or newer on NVIDIA CUDA",
+)
+@pytest.mark.parametrize(
+    "grad_weight_dtype, operand_dtype",
+    [
+        (torch.float16, torch.float16),
+        (torch.bfloat16, torch.bfloat16),
+        (torch.float32, torch.float32),
+        (torch.float32, torch.float16),
+        (torch.float32, torch.bfloat16),
+    ],
+)
+def test_accumulate_grad_weight_out_dtype(grad_weight_dtype, operand_dtype, monkeypatch):
+    requires_ampere = operand_dtype == torch.bfloat16 or grad_weight_dtype != operand_dtype
+    if requires_ampere and torch.cuda.get_device_capability()[0] < 8:
+        pytest.skip("this addmm dtype combination requires compute capability 8.0 or newer")
+
+    original_addmm = torch.addmm
+    observed_out_dtypes = []
+
+    def fail_mm(*args, **kwargs):
+        raise AssertionError("eligible accumulation unexpectedly used the allocating mm fallback")
+
+    def record_addmm(*args, **kwargs):
+        observed_out_dtypes.append(kwargs.get("out_dtype"))
+        return original_addmm(*args, **kwargs)
+
+    monkeypatch.setattr(torch, "mm", fail_mm)
+    monkeypatch.setattr(torch, "addmm", record_addmm)
+    grad_weight, expected = _run_grad_weight_accumulation(operand_dtype, grad_weight_dtype)
+
+    assert observed_out_dtypes == [grad_weight_dtype]
+    torch.testing.assert_close(grad_weight, expected, atol=5e-3, rtol=5e-2)
+
+
+@pytest.mark.parametrize("operand_dtype", [torch.float16, torch.bfloat16])
+def test_accumulate_grad_weight_fallback_mixed_dtype(operand_dtype, monkeypatch):
+    original_mm = torch.mm
+    observed_operand_dtypes = []
+
+    def fail_addmm(*args, **kwargs):
+        raise AssertionError("fallback accumulation unexpectedly used addmm")
+
+    def record_mm(mat1, mat2, *args, **kwargs):
+        observed_operand_dtypes.append((mat1.dtype, mat2.dtype))
+        return original_mm(mat1, mat2, *args, **kwargs)
+
+    monkeypatch.setattr(fused_linear_cross_entropy, "_ADDMM_SUPPORTS_OUT_DTYPE", False)
+    monkeypatch.setattr(torch, "addmm", fail_addmm)
+    monkeypatch.setattr(torch, "mm", record_mm)
+    grad_weight, expected = _run_grad_weight_accumulation(operand_dtype, torch.float32)
+
+    assert observed_operand_dtypes == [(operand_dtype, operand_dtype)]
+    torch.testing.assert_close(grad_weight, expected, atol=5e-3, rtol=5e-2)
 
 
 #############################################################################


src/liger_kernel/ops/fused_linear_cross_entropy.py
```diff
@@ -212,30 +261,7 @@ def fused_linear_cross_entropy_forward(
...
-                # can stay fp32 while grad_logits is the autocast dtype. addmm requires mat1
-                # and mat2 to share a dtype, so align _input_chunk before accumulating.
-                input_chunk = _input_chunk
-                if input_chunk.dtype != grad_logits_t.dtype:
-                    input_chunk = input_chunk.to(grad_logits_t.dtype)
-                torch.addmm(
-                    grad_weight,
-                    grad_logits_t,
-                    input_chunk,
-                    out_dtype=torch.float32,
```
Tcc0403, line 234: wouldn't it be more straightforward to set `out_dtype=grad_weight.dtype`? with a little bit change on branching condition.

Anubhav-2003: Thank you very much for the review. I see how that would simplify the path for PyTorch ≥2.8. Since Liger supports PyTorch ≥2.1.2, using out_dtype for all these cases would either require retaining a separate path for older versions or leave them on the allocating fallback. That’s why I kept addmm_ for same-dtype accumulation. Would you prefer limiting this optimization to PyTorch ≥2.8 instead? But then older versions will keep using the unoptimized path.

Tcc0403: yes, modify output_dtype in 2.8+ path. 

`add_` supports mismatching dtype but I don't recall which version starts supporting it. We can change the other path to `grad_weight.add_(torch.mm(grad_logits_chunk.t(), _input_chunk))` if 2.1.2 has it covered

Anubhav-2003: Thank you for the clarification. I checked PyTorch 2.1.2, and add_ supports the floating-point dtype conversions needed here. I’ll use out_dtype=grad_weight.dtype for the eligible 2.8+ path and grad_weight.add_(torch.mm(...)) as the fallback, with coverage for both paths.

Anubhav-2003:
@Tcc0403
Respectd Sir,

I have made the changes as per your advice. You may take a look whenever you are free.

Thank You.

PR #1326, opened 2026-07-27
[Megatron] Add SwiGLU integration

buffett0323:
## Summary
Adds Liger's Triton SwiGLU to Megatron-Core. Megatron's dense MLP routes its activation through `bias_swiglu_impl` in `megatron.core.fusions.fused_bias_swiglu`, a TorchScript `@jit_fuser` implementation. This PR replaces it with a **new fused gate-up Triton kernel** written for Megatron's memory layout.

At the Megatron training shape (`S=2048, B=4, bf16`) on **H100**, against `megatron-core==0.16.1`:

|  | Forward | Full (fwd+bwd) | Memory | Memory (`in_place=True`) |
|---|---|---|---|---|
| `ffn_local=4096` | **+57%** | **+146%** | +20% | **−20%** |
| `ffn_local=32768` | **+63%** | **+166%** | +20% | **−20%** |

megatron_swiglu_panel_H100

### Background

Liger's existing `LigerSiLUMulFunction` takes gate and up as **two separate tensors** (`a`, `b`), because that is how HuggingFace models arrive: `gate_proj(x)` and `up_proj(x)` are distinct GEMMs.

Megatron is the opposite. `linear_fc1` is a single fused `ColumnParallelLinear` producing one `[s, b, 2 * ffn_local]` tensor with gate in columns `[0, n)` and up in `[n, 2n)`. Bridging to the two-tensor kernel therefore requires a `torch.chunk`, and `chunk` on the last dimension returns **non-contiguous views**, so `@ensure_contiguous` materializes 2 full copies on every forward. We measured that bridge before writing anything new. 

| H100, forward, `ffn_local=32768` | Time | vs Megatron |
|---|---|---|
| Megatron `bias_swiglu_impl` | 0.853 ms | — |
| Liger via `chunk` bridge | 1.828 ms | **−53%** (2.1× slower) |
| **Liger fused gate-up (this PR)** | **0.536 ms** | **+59%** |

The chunk bridge is **3.4× slower than the fused kernel** and also inflated memory by +40%. The new `LigerFusedGateUpSiLUMulFunction` reads both halves via a column offset into the single buffer, with no copies, no `cat`. Input row stride is `2n`, output row stride is `n`.

```python

apply_liger_kernel_to_megatron(rms_norm=True, cross_entropy=True, swiglu=True) # swiglu defaults to False
```
Patches `bias_swiglu_impl` across Megatron's core modules (`fused_bias_swiglu`, `mlp`, and `moe.shared_experts`).

### Mode 2: Use the class directly

```python
from liger_kernel.megatron import LigerMegatronSwiGLU

swiglu = LigerMegatronSwiGLU()
out = swiglu(intermediate, bias, fp8_input_store, cpu_offload_input)
```

The signature mirrors `bias_swiglu_impl` positionally. To use Mode 2, subclass MLP (`see examples/megatron/run_mode2_hand_spec.py`).

### Optional: `in_place=True` for a 20% memory reduction

```python
LigerMegatronSwiGLU(in_place=True) # Opt-in, off by default
```
Overwrites the linear_fc1 output buffer during backward passes to reduce memory by 20% vs Megatron with zero speed penalty.

## Scope
- Supported: Dense MLP (`bias_swiglu_impl`) and shared experts at any TP size with `bias=None`.
- Transparent Fallbacks to Megatron: Auto-delegates to Megatron (and logs once) when:
  - `bias is not None` (Liger lacks bias support)
  - `fp8_input_store=True` (requires FP8 input retention)
  - `cpu_offload_input=True` (requires offloadable inputs)
- Pass `fallback_impl=None` to raise an error instead of falling back.
- Out of Scope: MoE routed experts (`weighted_bias_swiglu_impl`).

## What's added

| File | Purpose |
|---|---|
| `src/liger_kernel/ops/swiglu.py` | New `LigerFusedGateUpSiLUMulFunction` + fwd/bwd Triton kernels for the `[tokens, 2n]` layout. Existing two-tensor path untouched. |
| `src/liger_kernel/megatron/swiglu.py` | `LigerMegatronSwiGLU` nn.Module, fallback policy, shape validation. |
| `src/liger_kernel/megatron/monkey_patch.py` | `swiglu=` flag, `_patch_bias_swiglu_impl()`, `_rebind_consumer_symbol()`. |
| `src/liger_kernel/megatron/__init__.py` | Re-exports `LigerMegatronSwiGLU`. |
| `test/megatron/test_swiglu.py` | 36 tests — 28 dependency-free + 8 against real `megatron-core`. |
| `test/megatron/test_monkey_patch.py` | +14 SwiGLU patch-mechanism tests; stub-harness snapshot/restore rewrite. |
| `benchmark/scripts/benchmark_megatron_swiglu.py` | 4 providers: `liger`, `liger_in_place`, `megatron`, `torch`. |
| `examples/megatron/*` | Mode 1 + Mode 2 extended to cover SwiGLU. |

## Megatron feature compatibility

**Verified:** Tensor Parallelism (TP), fp32 + bf16, 2D and 3D inputs, distributed checkpointing (save/load round-trip in both example scripts), real `MLP` forward+backward parity against Megatron's TorchScript, and both consumer module paths.

**Should work, not yet E2E-verified:** multi-rank patch application (every rank must call `apply_...` before model build — single-process patching is verified, the multi-rank ordering is not), CP, `use_distributed_optimizer=True`.

**Out of scope:** activation recompute and CUDA-graph capture *with `in_place=True`* — precisely why that flag is opt-in. The default path is recompute-safe.

## Known issues / follow-ups
1. `in_place=True` and `retain_graph=True` incompatibility: In-place Triton kernel writes don't bump PyTorch's `_version` counter, so running backward a second time silently yields garbage gradients without triggering autograd guards. We added a `ctx.already_backward` flag to explicitly raise an error when reusing the graph with `in_place=True`.
2. MoE routed experts: `weighted_bias_swiglu_impl` requires a `per_token_scale`-aware kernel (left as a follow-up).
3. RMSNorm patch fix: Fixed a TypeError in `_patch_local_spec_provider_layer_norm` caused by forwarding an unsupported has_residual kwarg on `megatron-core 0.16.1`. Signature matching is now version-agnostic via `inspect.signature`.
4. Example script fixes: Fixed three pre-existing `megatron-core 0.16.1` compatibility issues in the example scripts (AttributeError on `as_mlp_submodule`, Makefile build errors in pip wheels, and invalid tokenizer metadata).

## Testing Done
- Hardware Type: NVIDIA A10G (correctness, 1× and 2×), H100 80GB HBM3 (benchmarks), B200 (benchmarks)
- [x] `pytest test/megatron/` **with** `megatron-core` installed — **175 passed, 49 skipped**
- [x] `pytest test/megatron/` **without** `megatron-core` (default CI shape) — **168 passed, 56 skipped**
- [x] `pytest test/transformers/test_swiglu.py` — **49 passed, 4 skipped, 4 xfailed** (existing two-tensor SwiGLU not regressed by the shared `ops/swiglu.py` change)
- [x] run `make checkstyle` to ensure code style — clean; `prek run -a` also clean
- [x] H100 + B200 benchmark sweeps; H100 rows committed to `all_benchmark_data.csv`

## Benchmark reproduction
```bash
python benchmark/scripts/benchmark_megatron_swiglu.py --overwrite

    --kernel-name megatron_swiglu --metric-name speed --overwrite
```


src/liger_kernel/ops/swiglu.py
```diff
@@ -281,3 +281,148 @@ def backward(ctx, dc):
...
+
+    # Gate occupies columns [0, n), up occupies [n, 2n) of the same row.
+    gate = tl.load(y_ptr + col_offsets, mask=mask, other=0).to(tl.float32)
+    up = tl.load(y_ptr + n_cols + col_offsets, mask=mask, other=0)
+    tl.store(c_ptr + col_offsets, silu(gate).cast(up.dtype) * up, mask=mask)
+
+
+@triton.jit
+def _swiglu_fused_gate_up_backward_kernel(
+    dc_ptr, y_ptr, dy_ptr, in_stride, out_stride, n_cols: tl.constexpr, BLOCK_SIZE: tl.constexpr
```
Tcc0403:
```suggestion
    dc_ptr, y_ptr, dy_ptr, in_stride, out_stride, ffn_size: tl.constexpr, BLOCK_SIZE: tl.constexpr
```
nit: name it `ffn_size` for readability

src/liger_kernel/ops/swiglu.py
```diff
@@ -281,3 +281,148 @@ def backward(ctx, dc):
...
+    tl.store(dy_ptr + n_cols + col_offsets, d_up, mask=mask)
+
+
+def swiglu_fused_gate_up_forward(y):
+    """SwiGLU over a fused ``[..., 2 * n]`` gate-up tensor. Returns ``(y, c)``."""
+    ori_shape = y.shape
+    two_n = ori_shape[-1]
+    if two_n % 2 != 0:
+        raise ValueError(f"fused gate-up input must have an even trailing dim; got {two_n}.")
+    n_cols = two_n // 2
```
Tcc0403: ditto

examples/megatron/README.md
```diff
@@ -2,8 +2,9 @@
 
 Two self-contained scripts demonstrating the integration modes shipped by
 `liger_kernel.megatron`. Both train a tiny GPT model with mock data on
-2 × GPU (TP=2, PP=1) for 5 iterations and print the resolved norm classes
-so you can see which slots picked up Liger.
+2 × GPU (TP=1, PP=1, DP=2) for 5 iterations and print the resolved norm
+classes, CE bindings and SwiGLU bindings so you can see which slots picked
```
Tcc0403: it would be better to create a op support matrix and append new op to it instead of inline new changes

examples/megatron/run_mode1_monkey_patch.py
```diff
@@ -93,9 +109,23 @@ def model_provider() -> GPTModel:
     )
 
 
+def _ensure_dataset_helpers() -> None:
```
Tcc0403: out of this PR scope? is it a required change to make this script work?

examples/megatron/run_mode2_hand_spec.py
```diff
@@ -174,9 +219,23 @@ def model_provider() -> GPTModel:
     )
 
 
+def _ensure_dataset_helpers() -> None:
```
Tcc0403: ditto

src/liger_kernel/megatron/monkey_patch.py
```diff
@@ -72,16 +85,20 @@ def _patch_local_spec_provider_layer_norm() -> None:
...
+        # after ``for_qk``), and a hardcoded forwarding call raises TypeError on any
+        # version whose signature does not match exactly. Binding to the real signature
+        # also means ``rms_norm`` is read correctly whether the caller passed it
+        # positionally or by keyword.
+        bound = _original_sig.bind(self, *args, **kwargs)
+        bound.apply_defaults()
+        if bound.arguments.get("rms_norm", False):
             return LigerMegatronRMSNorm
-        return original_layer_norm(self, rms_norm=rms_norm, for_qk=for_qk, has_residual=has_residual)
+        return original_layer_norm(*bound.args, **bound.kwargs)
```
Tcc0403: I prefer all arguments listed to keep it clean

src/liger_kernel/megatron/monkey_patch.py
```diff
@@ -260,3 +277,107 @@ def liger_vocab_parallel_cross_entropy(
...
+        liger_bias_swiglu_impl,
+    )
+
+    logger.info(
+        "Patched megatron.core.fusions.fused_bias_swiglu.bias_swiglu_impl with Liger SwiGLU (also rebound in: %s).",
+        ", ".join(patched_consumers) if patched_consumers else "no already-imported consumers",
+    )
+
+
+def _rebind_consumer_symbol(
```
Tcc0403: I don't fully understand the purpose of this function. In what scenario do we need this rebind function, could you elaborate further? 

Should we also apply it to ce and norm?

src/liger_kernel/megatron/swiglu.py
```diff
@@ -0,0 +1,162 @@
...
+
+import torch
+import torch.nn as nn
+
+# Force-import the submodule so liger_kernel.ops.swiglu can resolve
+# torch.distributed.tensor.DTensor on torch 2.11+, where the subpackage is no longer
+# auto-loaded as an attribute of torch.distributed.
+import torch.distributed.tensor  # noqa: F401  # isort: skip
+
+from liger_kernel.ops import LigerFusedGateUpSiLUMulFunction
```
Tcc0403: would `from liger_kernel.ops import LigerFusedGateUpSiLUMulFunction` fail without importing `torch.distributed.tensor` beforehand?

test/megatron/test_monkey_patch.py
```diff
@@ -188,26 +191,73 @@ class _SomeOtherNorm:
...
 
+_SAVED_REAL_MEGATRON: dict = {}
+_STUBS_ACTIVE = False
+
+
+def _is_megatron_module(name: str) -> bool:
+    return name == "megatron" or name.startswith("megatron.")
+
+
+def _snapshot_real_megatron():
```
Tcc0403: is there any conflict with stub megatron when megatron-core is installed? would the test fail without these functions?

test/megatron/test_monkey_patch.py
```diff
@@ -1023,3 +1164,288 @@ def test_rms_norm_only_patch_does_not_touch_ce_symbols(fake_megatron_ce):
...
     assert unfused_ce.vocab_parallel_cross_entropy is unfused_before
+
+
+# ===========================================================================
+# 6. SwiGLU patch tests
+# ===========================================================================
+# Liger replaces ``megatron.core.fusions.fused_bias_swiglu.bias_swiglu_impl`` — the symbol
+# ``MLP.forward`` and ``SharedExpertMLP.forward`` call when bias_activation_fusion is on.
+# Unlike CE and RMSNorm, its consumers use by-name imports, so the patch has to rewrite
+# their module attributes too; several tests below exist specifically to enforce that.
```
Tcc0403: Follow CE and RMSNorm tests convention for self-consistent and maintainability. Issues in megatron-installed environment should be landed in another PR

test/megatron/test_swiglu.py
```diff
@@ -0,0 +1,642 @@
...
+#
+# Everything above verifies the wrapper against our own reproduction of Megatron's
+# formula. That cannot catch a wrong belief about Megatron's API. The tests below import
+# the real package and pin: the true signature and positional order of
+# ``bias_swiglu_impl``, numerical parity with Megatron's own TorchScript implementation,
+# that the consumer module paths the monkey patch rebinds actually exist and are rebound,
+# and that a real ``MLP`` produces identical output and gradients when patched.
+#
+# The import is conditional rather than a module-level ``pytest.importorskip`` so that the
+# dependency-free tests above still run when megatron-core is absent.
```
Tcc0403: Although this section is skipped when megatron-core is not installed, I still suggest removing it to keep this PR concise. Feel free to raise an issue if you plan to handle actual megatron-core environment in the test suite.

Tcc0403 (commented):
Thanks for the contribution, the kernel impl looks good. 

However, there are plenty of new added functions to resolve environment issue with megatron-core installed in our test suite. I prefer moving them to another PR, ideally a proposal issue first, to address such issues.

buffett0323:
Hi @Tcc0403 , thanks for reviewing. The comments are pretty thorough, and I updated the code.

A big change is instead of patching `bias_swiglu_impl` and then rebinding it in a hardcoded list of consumer modules, I now patch `fused_bias_swiglu.SwiGLUFunction` directly.

`bias_swiglu_impl` is what consumers import, and they import it by name (`transformer/mlp.py`,  `moe/shared_experts.py`), so reassigning it in the defining module doesn't reach them. so the old rebind list, which would possibly miss any new consumer in a future release.

`SwiGLUFunction` doesn't have that problem. Instead, `bias_swiglu_impl` resolves it from its own module globals on every call, and it's referenced nowhere else in `megatron-core`. One binding, no stale copies, so the patch reaches every consumer regardless of import order.

buffett0323:
Hi @Tcc0403 , gentle reminder for this PR whenever you get a chance! Let me know if you have any questions or feedback. Thanks again for your help!!

Tcc0403 (approved):
LGTM. cc @vaibhavjindal for a second look

vaibhavjindal (approved):
LGTM! Thanks a lot for contributing this kernel and the integration!

PR #1328, opened 2026-07-31
Fix FLCE torch.compile failure from aten.addmm.dtype_out overload

### LINKED ISSUE #1327: `torch.compile` fails with FLCE due to `aten.addmm.dtype_out` overload
## Description

`LigerFusedLinearCrossEntropy` (FLCE) raises an error when traced via `torch.compile` under the following conditions:

- PyTorch 2.8.0 or later.
- NVIDIA CUDA, compute capability 8.0 or higher.
- The gradient computation path for `grad_weight` passes through `torch.addmm(..., out_dtype=..., out=...)`.
- PyTorch uses FakeTensor to trace the graph.

The error occurs at trace time, before any CUDA kernel is executed. Eager mode is unaffected.

## Root Cause

When both `out_dtype=` and `out=` are passed simultaneously, the dispatcher selects the overload:

```
aten.addmm.dtype_out
```

This is a two-tier issue with different fix statuses at each tier:

### Tier 1 — Core decomposition (Fixed on `main` & PyTorch 2.13.0)

The legacy decomposition registered for the entire `aten.addmm` packet used the `@pw_cast_for_opmath` decorator, which remaps positional arguments after upcasting. Consequently, `out_dtype` (of type `torch.dtype`) was passed into the position of `beta`, making the internal expression `beta * self` evaluate as `torch.dtype * FakeTensor` and raising:

```
TypeError: unsupported operand type(s) for *: 'torch.dtype' and 'FakeTensor'
```

[PR #179634](https://github.com/pytorch/pytorch/pull/179634) (commit [`c4c886d`](https://github.com/pytorch/pytorch/commit/c4c886d9f62b69150800e0fc96199e3b36d61181)) created a dedicated `addmm_dtype` handler for overloads containing `out_dtype` without using `@pw_cast_for_opmath`. This commit is an ancestor of `main` and is included in PyTorch 2.13.0.

### Tier 2 — TorchInductor (Unfixed)

TorchInductor maintains its own decomposition table and currently registers decompositions across the full `aten.addmm` packet, preventing the fixed Tier 1 core decomposition from handling `.dtype` and `.dtype_out` overloads. While the operation still fails under `torch.compile`, the failure signature following the core fix may change—for instance, shifting to a lowering error such as:

```
TypeError: tuned_addmm() takes 3 positional arguments but 4 were given
```

The complete fix for Tier 2 is currently tracked in [PR #190936](https://github.com/pytorch/pytorch/pull/190936) (open, unmerged, CI failing). The original bug was reported in [pytorch/pytorch#163880](https://github.com/pytorch/pytorch/issues/163880) (currently marked *Closed*, but outdated—the issue closed automatically following a partial fix that was subsequently reverted, while the bug remains reproducible on `main`).

## Scope of Impact on `main`

The affected fast path is triggered when all of the following conditions are met:

- PyTorch ≥ 2.8 (including 2.13.0).
- NVIDIA CUDA, compute capability ≥ 8.0.
- `grad_weight` has FP32 dtype.
- `grad_logits` has FP16 or BF16 dtype.
- Weight gradients are required (does not apply to inference-only, frozen weights, or passes where weight gradients are skipped).
- Code is executed via `torch.compile` / TorchInductor.

Configuring `accum_dtype=torch.float32` with FP16/BF16 inputs is the minimal setup to reproduce this issue.

## How to Reproduce

Run the following script in an environment with Liger-Kernel installed:

```python
import contextlib
import io
import logging
import platform
from importlib.metadata import PackageNotFoundError, version

import torch

KNOWN_FAILURES = (
    # Core decomposition error — PyTorch 2.8
    "unsupported operand type(s) for *: 'torch.dtype' and 'FakeTensor'",
    # Inductor lowering error — PyTorch 2.13+ / main after core fix
    "tuned_addmm() takes 3 positional arguments but 4 were given",
)


def package_version(name, fallback):
    try:
        return version(name)
    except PackageNotFoundError:
        return fallback


def main():
    cuda_available = torch.cuda.is_available()

    print(f"OS:      {platform.platform()}")
    print(f"Python:  {platform.python_version()}")
    print(f"PyTorch: {torch.__version__}")
    print(f"Liger:   {package_version('liger-kernel', 'source checkout')}")
    print(f"Triton:  {package_version('triton', 'not installed')}")
    print(f"CUDA:    {torch.version.cuda or 'not built'}")

    if cuda_available:
        capability = torch.cuda.get_device_capability()
        print(
            f"GPU:     {torch.cuda.get_device_name()} "
            f"(compute capability {capability[0]}.{capability[1]})"
        )

    print("Case:    FLCE BF16 operands -> FP32 accumulator, torch.compile/Inductor")

    if "dtype_out" not in torch.ops.aten.addmm.overloads():
        print("Result:  SKIP — requires PyTorch 2.8+")
        return 2

    if not cuda_available or torch.version.hip is not None:
        print("Result:  SKIP — requires NVIDIA CUDA")
        return 2

    if capability[0] < 8:
        print("Result:  SKIP — requires compute capability 8.0+")
        return 2

    from liger_kernel.transformers.functional import liger_fused_linear_cross_entropy

    x = torch.randn(8, 16, device="cuda", dtype=torch.bfloat16, requires_grad=True)
    weight = torch.randn(32, 16, device="cuda", dtype=torch.bfloat16, requires_grad=True)
    target = torch.randint(32, (8,), device="cuda")

    def step(x, weight, target):
        return liger_fused_linear_cross_entropy(
            x,
            weight,
            target,
            accum_dtype=torch.float32,
        )

    logging.disable(logging.CRITICAL)
    torch._dynamo.config.capture_scalar_outputs = True

    try:
        with contextlib.redirect_stderr(io.StringIO()):
            torch.compile(step)(x, weight, target).backward()
    except Exception as error:
        message = str(error)

        for expected in KNOWN_FAILURES:
            if expected in message:
                print(f"Result:  REPRODUCED — {message.splitlines()[0]}")
                return 0

        print(
            f"Result:  UNEXPECTED — "
            f"{type(error).__name__}: {message.splitlines()[0]}"
        )
        return 1

    print("Result:  NOT REPRODUCED")
    return 1


raise SystemExit(main())
```

## Actual Results on PyTorch 2.8.0

Reproduced on two separate GPUs. The observed failure mode is the Tier 1 core decomposition error characteristic of PyTorch 2.8.0 prior to the core fix merge.

### NVIDIA L4

```
OS:      Linux-4.19.0-gvisor-x86_64-with-glibc2.36
Python:  3.12.6
PyTorch: 2.8.0+cu129
Liger:   0.8.1
Triton:  3.4.0
CUDA:    12.9
GPU:     NVIDIA L4 (compute capability 8.9)
Case:    FLCE BF16 operands -> FP32 accumulator, torch.compile/Inductor
Result:  REPRODUCED — unsupported operand type(s) for *: 'torch.dtype' and 'FakeTensor'
```

### NVIDIA A100

```
OS:      Linux-4.19.0-gvisor-x86_64-with-glibc2.36
Python:  3.12.6
PyTorch: 2.8.0+cu129
Liger:   0.8.1
Triton:  3.4.0
CUDA:    12.9
GPU:     NVIDIA A100-SXM4-40GB (compute capability 8.0)
Case:    FLCE BF16 operands -> FP32 accumulator, torch.compile/Inductor
Result:  REPRODUCED — unsupported operand type(s) for *: 'torch.dtype' and 'FakeTensor'
```

## Expanded Scope from PR #1324

PR [linkedin/Liger-Kernel#1324](https://github.com/linkedin/Liger-Kernel/pull/1324) (*[Perf] Use direct FLCE weight-gradient accumulation for FP16, BF16, and FP32*, open, unmerged) extends the fast path to same-dtype cases (FP16/FP16, BF16/BF16, FP32/FP32):

```python
torch.addmm(
    grad_weight,
    grad_logits_t,
    input_chunk,
    out_dtype=grad_weight.dtype,
    out=grad_weight,
)
```

This invocation still dispatches to `aten.addmm.dtype_out` and triggers the same bug. If PR #1324 is merged, the default configuration (`accum_dtype=None`) will enter the `aten.addmm.dtype_out` overload for eligible same-dtype FP16, BF16, or FP32 training workloads. As a result, many callers using `torch.compile` will encounter errors without needing to opt in via `accum_dtype=torch.float32`, whereas the issue is currently limited to that opt-in setup.

yashb98:
## Summary

Fixes #1327.

FLCE's grad_weight fast path calls `torch.addmm(..., out_dtype=torch.float32, out=...)`, which dispatches to `aten.addmm.dtype_out`. TorchInductor cannot lower that overload yet (the Tier-2 fix, pytorch/pytorch#190936, is still open), so any `torch.compile`'d FLCE backward with bf16/fp16 inputs and `accum_dtype=torch.float32` dies at lowering time with:

```
LoweringException: TypeError: tuned_addmm() takes 3 positional arguments but 4 were given
```

This PR gates the fast path on `not torch.compiler.is_compiling()`. Under `torch.compile` the backward takes the existing generic matmul path, which Inductor lowers fine; eager mode is unchanged and keeps the fused `addmm` accumulation. When the PyTorch-side fix lands this guard can be revisited.

## Details

- One-line guard in `fused_linear_cross_entropy_forward`; no behavioural change outside `torch.compile`.
- Adds `test_torch_compile_fp32_accum_backward`: compiles FLCE with bf16 operands + `accum_dtype=torch.float32`, runs backward, and checks loss / grad_x / grad_w against eager. It fails on current main with the exact `tuned_addmm()` signature from the issue and passes with this change.
- Verified the compiled fallback is numerically correct, not just non-crashing: grad_x is bitwise identical to eager, grad_w differs only at bf16 rounding (max abs diff 0.00195 on the repro shape).

## Testing Done

Reproduced the issue on current main (9ac26bb) with the script from #1327 before changing anything:

```
PyTorch: 2.13.0+cu130, Triton 3.7.1, GB10 (compute capability 12.1)
Result:  REPRODUCED - LoweringException: TypeError: tuned_addmm() takes 3 positional arguments but 4 were given
```

After the change: `Result: NOT REPRODUCED` (compiled backward completes).

- `test/transformers/test_fused_linear_cross_entropy.py`: 142 passed (was 141 + the new regression test)
- New test, unpatched main: fails with the `tuned_addmm()` error above
- New test, with this change: passes
- Full unit suite (`pytest --ignore=test/convergence test/`), patched: 3916 passed, 2 failed — the 2 failures are `test_grpo_loss_vs_trl[...-luspo-sequence-1.5]`, and the identical 2 fail on unpatched main in the same full-suite run (3915 passed, 2 failed). They pass in isolation in both arms and live in `grpo_loss`, which this change does not touch: a pre-existing, order-dependent flake.
- `ruff check` and `ruff format --check` on both touched files: clean

- Hardware Type: NVIDIA GB10 (DGX Spark, sm_121, aarch64)

One honest caveat: I did not run `make test-convergence` (model-scale convergence runs); the change only reroutes an existing eager path under compilation, so convergence behaviour is unaffected in eager and the compiled path is checked against eager in the new test.

## Benchmarks

GB10 (DGX Spark, sm_121), torch 2.13.0+cu130, bf16 operands with fp32 accumulation (the affected configuration), forward + backward, median of 8 runs after 3 warmup. Branch is rebased on current main (91ae44a).

| shape | eager, main | eager, this PR | torch.compile, main | torch.compile, this PR |
|---|---|---|---|---|
| 2048 tok, H=1024, V=151936 (Qwen3-0.6B class) | 1232.9 ms | 1238.5 ms (+0.5%) | LoweringException (the bug) | 780.3 ms |
| 4096 tok, H=4096, V=128256 (Llama3-8B class) | 1047.1 ms | 1053.3 ms (+0.6%) | LoweringException (the bug) | 709.8 ms |

Two takeaways: eager is unchanged (parity within noise, as expected since the guard only fires under compilation), and the compiled path this PR unblocks is not merely working but about 1.5x faster than eager on both shapes. The unpatched compiled run still fails with the exact `tuned_addmm()` LoweringException from the issue, re-confirmed on main at 91ae44a today.

flce_bench

### DIFF
--- a/src/liger_kernel/ops/fused_linear_cross_entropy.py
+++ b/src/liger_kernel/ops/fused_linear_cross_entropy.py
@@ -234,6 +234,7 @@ def fused_linear_cross_entropy_forward(
             grad_logits_t = grad_logits_chunk.t()
             if (
                 _ADDMM_SUPPORTS_OUT_DTYPE
+                and not torch.compiler.is_compiling()
                 and grad_weight.device.type == "cuda"
                 and torch.cuda.get_device_capability(grad_weight.device)[0] >= 8
                 and grad_weight.dtype == torch.float32
--- a/test/transformers/test_fused_linear_cross_entropy.py
+++ b/test/transformers/test_fused_linear_cross_entropy.py
@@ -1092,3 +1092,48 @@ def test_correctness_reduction_none_per_token_grad_output(B, T, H, V, bias, dtyp
     assert_verbose_allclose(lig_gw, ref_gw, atol=atol, rtol=rtol)
     if bias:
         assert_verbose_allclose(lig_gb, ref_gb, atol=atol, rtol=rtol)
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="requires CUDA")
+@pytest.mark.skipif(
+    "dtype_out" not in torch.ops.aten.addmm.overloads(),
+    reason="addmm out_dtype path requires PyTorch >= 2.8",
+)
+def test_torch_compile_fp32_accum_backward():
+    """FLCE backward must work under torch.compile with fp32 accumulation.
+
+    Regression test for https://github.com/linkedin/Liger-Kernel/issues/1327:
+    the grad_weight fast path calls torch.addmm(..., out_dtype=..., out=...),
+    which TorchInductor cannot lower (aten.addmm.dtype_out), so compiling a
+    backward with bf16 inputs and accum_dtype=torch.float32 raised
+    "tuned_addmm() takes 3 positional arguments but 4 were given". The compiled
+    path must fall back to a lowerable formulation and still match eager.
+    """
+    set_seed(42)
+    B, H, V = 8, 16, 32
+
+    _input = torch.randn(B, H, device=device, dtype=torch.bfloat16)
+    weight = torch.randn(V, H, device=device, dtype=torch.bfloat16)
+    target = torch.randint(0, V, (B,), device=device)
+
+    def eager():
+        x = _input.clone().requires_grad_(True)
+        w = weight.clone().requires_grad_(True)
+        loss = liger_fused_linear_cross_entropy(x, w, target, accum_dtype=torch.float32)
+        loss.backward()
+        return loss, x.grad, w.grad
+
+    def compiled():
+        x = _input.clone().requires_grad_(True)
+        w = weight.clone().requires_grad_(True)
+        fn = torch.compile(liger_fused_linear_cross_entropy)
+        loss = fn(x, w, target, accum_dtype=torch.float32)
+        loss.backward()
+        return loss, x.grad, w.grad
+
+    ref_loss, ref_gx, ref_gw = eager()
+    cmp_loss, cmp_gx, cmp_gw = compiled()
+
+    assert_verbose_allclose(cmp_loss, ref_loss, atol=1e-3, rtol=1e-3)
+    assert_verbose_allclose(cmp_gx, ref_gx, atol=1e-3, rtol=1e-2)
+    assert_verbose_allclose(cmp_gw, ref_gw, atol=1e-3, rtol=1e-2)


yashb98:
CI has never run on this PR, all four workflows have been sitting at `action_required` since 2026-07-31. Could a maintainer approve the run?

In the meantime, here is a before/after against current `main` @ `08e8ccdc`. GB10 (DGX Spark, sm_121, aarch64), torch 2.13.0+cu130, Triton 3.7.1, same box as the PR description.

Unpatched `main` with only the new test applied:

```
E       torch._inductor.exc.InductorError: LoweringException: TypeError: tuned_addmm() takes 3 positional arguments but 4 were given
E         target: aten.addmm.dtype
E          File ".../liger_kernel/ops/fused_linear_cross_entropy.py", line 249, in torch_dynamo_resume_in_fused_linear_cross_entropy_forward_at_93

1 failed, 141 deselected, 15 warnings in 2.86s
```

Same tree plus the one line from this PR:

```
1 passed, 141 deselected, 15 warnings in 2.23s
```

Whole file:

```
142 passed, 15 warnings in 5.18s
```

One thing that may be worth recording in the issue: on torch 2.11.0+cu130 the same test fails earlier and differently, in Dynamo rather than in Inductor lowering, with `TypeError: unsupported operand type(s) for *: 'torch.dtype' and 'FakeTensor'` at the same line 249. That is the Tier 1 signature #1327 describes for torch versions without the core decomposition fix. Both signatures come from the same `aten.addmm.dtype_out` dispatch, and this one-line guard fixes both; I ran the patched test on 2.11.0 and 2.13.0 and it passes on each.

The branch shows as behind by 3 commits, but neither file it touches is among the files those commits changed, and the runs above are on `08e8ccdc` itself, so the rebase is already covered. Happy to rebase whenever you want it.

PR #1268, opened 2026-06-25
perf(ce): single-shot dx_y gradient correction

justinhh4:
## Summary

This PR isolates one optimization on top of the already-merged CE improvements in #1266 (`exp2`) and
#1267 (dtype-aware `num_warps`): replace the per-element true-class gradient selection with a single
post-loop `dx_y` correction.

At the reference shape (`BT=8192`, `V=128256`, full forward + backward), the isolated impact is:

| GPU | dtype | baseline (ms) | this PR (ms) | speedup |
|:----|:------|--------------:|-------------:|--------:|
| B200 | bf16 | 1.4229 | 1.1787 | **1.207x** |
| B200 | fp32 | 1.8820 | 1.8022 | **1.044x** |
| H100 | bf16 | 2.3162 | 2.2734 | **1.019x** |
| H100 | fp32 | 4.5362 | 4.5674 | **0.993x** |

## What changed

The CE gradient subtracts 1 only at the true-class index `y`. Previously, both pass-2 paths evaluated
a per-element `tl.where(X_offsets != y, ...)` across the entire vocabulary. This PR stores the plain
softmax gradient in the loop, then applies one scalar correction at `y` after the loop.

This removes an `O(V)` compare/select used to update one element and replaces it with an `O(1)` store.
A `tl.debug_barrier()` orders the loop's in-place writes before the correction reads and updates
`dx_y`. The correction retains the same weighting, reduction, label-smoothing, and softcap factors.

## Speed benchmarks

`baseline` is `main` with #1266 and #1267 but without this PR's single-shot `dx_y` change.
`speedup = baseline / this PR`, so values above `1.00x` mean this PR is faster.

Each latency is the average of two independent medians. B200 runs were collected on swapped NVIDIA
B200 GPUs; H100 runs were collected on one NVIDIA H100 80GB HBM3. Each median uses 40 timed iterations
after 20 warmup iterations with CUDA events. `full` is forward plus `loss.backward()` with
`grad_output=1.0`, matching the standard last-layer CE benchmark used in #1290.

### bf16

**Speed - full (forward + backward) - vocab sweep (`BT=8192`)**

| vocab | B200 baseline (ms) | B200 this PR (ms) | B200 speedup | H100 baseline (ms) | H100 this PR (ms) | H100 speedup |
|------:|-------------------:|------------------:|-------------:|-------------------:|------------------:|-------------:|
| 32000 | 0.5179 | 0.4455 | **1.163x** | 0.8484 | 0.7890 | **1.075x** |
[...]

PR #1329, opened 2026-08-02
fix(ce): restore true-class gradient precision in low-precision buffers

vaibhavjindal:
#1268 replaced the per-element `tl.where(X_offsets != y, ...)` selection with a post-loop read-modify-write of the true-class gradient:

    tl.store(X_ptr + y, tl.load(X_ptr + y) + dxy)

X_ptr is the logits buffer, which is bf16/fp16 in mixed-precision training (FLCE computes the gradient in place). The round-trip therefore rounds softmax(x_y) / N to the buffer dtype *before* the -(1 - label_smoothing) / N term is subtracted.

Since dx_y = (softmax(x_y) - 1) / N cancels catastrophically as softmax(x_y) -> 1, the pre-rounding error is amplified by roughly 1 / (1 - softmax(x_y)). Measured max relative error of dx_y against an fp64 reference (bf16, V=32000, H100):

    softmax(x_y)   before #1268   at #1268   torch
    0.90                  0.24%      1.95%   0.43%
    0.99                  0.30%     19.99%   0.46%
    0.999                 0.38%    100.00%   0.42%

Confident predictions are the common case once a model has trained for a while, so this silently degrades the dominant gradient signal. It surfaced as a convergence failure in
test_mini_model[mini_gemma3_text-32-1e-05-dtype25-0.01-0.01-0.1-0.01-0.01-0.01], where 32 steps of accumulated drift push a top-k logprob past tolerance.

Recompute dx_y once in fp32 after the loop, from the already-live ori_X_y, m and d, and write it with a single store. This keeps #1268's O(1) correction (no O(V) compare/select is reintroduced) while folding in the -(1 - label_smoothing) term before the result is rounded to the buffer dtype.

Accuracy returns to the pre-#1268 values exactly (0.02% / 0.36% / 0.24% / 0.30% / 0.38%), and the speedup is retained: 3.585 ms vs 3.571 ms (this PR vs #1268) at BT=8192, V=128256 bf16 forward+backward on H100, against 3.619 ms before #1268.

The existing CE suite passed with the regression because its random logits leave softmax(x_y) near zero, so add a regression test that drives softmax(x_y) -> 1 and requires the true-class gradient to be no less accurate than torch's own low-precision backward. It fails on the #1268 kernel and passes here with ~6x headroom.

## Testing Done

- Hardware Type: <BLANK>

### DIFF
--- a/src/liger_kernel/ops/cross_entropy.py
+++ b/src/liger_kernel/ops/cross_entropy.py
@@ -245,21 +245,36 @@ def liger_cross_entropy_kernel(
 
             tl.store(X_ptr + X_offsets, X_block, mask=X_offsets < n_cols)
 
-        # dx_y correction: apply the -(1 - label_smoothing) term once at index y
-        # (replaces the per-element tl.where removed above). Barrier first so the loop's
-        # in-place store to X[y] is visible before we read it back.
+        # dx_y correction: recompute the true-class gradient once, in fp32, and overwrite X[y]
+        # (replaces the per-element tl.where removed above). The -(1 - label_smoothing) term must
+        # be folded in *before* the result is rounded to X's dtype: dx_y = (softmax(x_y) - 1) / N
+        # cancels catastrophically as softmax(x_y) -> 1, so a read-modify-write of the value the
+        # loop already stored would lose most of the significant bits in bf16/fp16.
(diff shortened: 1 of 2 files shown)


justinhh4:
looks good to me! I also benchmarked it and got roughly same numbers as the 1268 PR

Mecoli1219 (approved):

PR #1331, opened 2026-08-02
Fix/device context for remaining kernels

### LINKED ISSUE #1303: Triton kernels can launch on the wrong GPU with multi-GPU inputs

kashif:
## Summary

Fix the remaining kernel calls

Fixes #1303

## Testing Done

- Hardware Type: <BLANK>


kashif:
cc @vaibhavjindal and @vasqu

src/liger_kernel/ops/cutile/ops/cross_entropy.py
vasqu, line 1: Oh are the same issues across cute and cutile? I thought it was a triton only issue but not familiar enough here

vasqu (commented):
Thanks, pretty much what we did as well. Unfamiliar with cute/cutile to say it's an issue there as well tbh

kashif:
Good catch on the cutedsl side — turns out the fix we had was incomplete. The original commit only wrapped `_cute_stream(device)` in `device_context`, but the actual failure happens earlier in `to_cute_tensor()` -> `from_dlpack()`. PyTorch refuses to DLPack-export a tensor on `cuda:1` when the current device is still `cuda:0`, so we get `BufferError: Can't export tensors on a different CUDA device index` before the stream is even acquired.

I just pushed a commit (df76d01) that wraps the entire `_launch_*` function body in `device_context(X.device)` for both rms_norm and cross_entropy, so the DLPack export and stream acquisition both run on the right device. Tested on 2x H100 via `srun` — both ops pass on `cuda:1` with the fix and fail without it.

kashif:
Gentle reminder, @vaibhavjindal, just in case you missed it!

vaibhavjindal:
> Gentle reminder, @vaibhavjindal, just in case you missed it!

Thanks for this change @kashif! Will soon merge this once we are done with some refactoring that we've been trying to do.

PR #1307, opened 2026-07-21
Add requires-python = ">=3.10" to pyproject.toml

feiiiiii5:
## Summary

Add `requires-python = ">=3.10"` to `[project]` in `pyproject.toml`.

Without this field, pip installs `liger_kernel` on any Python interpreter,
including unsupported versions, and fails later at import or runtime with
cryptic errors. Pinning `>=3.10` aligns the package metadata with the de-facto
support floor already enforced elsewhere:

- `ruff target-version = "py310"` (`pyproject.toml` line 58)
- release publish workflow uses 3.10 (`.github/workflows/publish-release.yml`)
- CI matrix starts at 3.10
- minimum `transformers >= 4.49.0` itself requires Python >= 3.9

This also addresses the workflow inconsistency noted in #1302
(`publish-nightly.yml` uses 3.8 while `publish-release.yml` uses 3.10) by
making 3.10 the canonical minimum declared in package metadata; the
nightly workflow's 3.8 reference is now stale and can be cleaned up in a
follow-up if maintainers prefer.

Closes #1302.

## Root Cause

`pyproject.toml` declared `dynamic = ["dependencies", "optional-dependencies"]`
but never set `requires-python`. Per PEP 440 / pip's resolver behavior, a
missing `requires-python` is treated as "any Python version", so `pip install
liger_kernel` succeeds on 3.7 / 3.8 / 3.9 etc. and only breaks later when
3.10-only syntax or `transformers>=4.49` APIs are touched.

## Fix

One line added to `[project]`:

```toml
license = { file = "LICENSE" }
requires-python = ">=3.10"
dynamic = ["dependencies", "optional-dependencies"]
```

## Testing Done

- `python3 -m pip install -e . --dry-run --no-deps --no-build-isolation` →
  `Preparing editable metadata (pyproject.toml): finished with status 'done'`
  (validates that the TOML parses correctly with the new field)
- `ruff check pyproject.toml` → `All checks passed!`
- Manual TOML syntax validation (bracket / quote balance) → OK

No `make test` run because this is metadata-only; no Python source files
changed and no test files exist for `pyproject.toml`. No `make test-convergence`
run because the change does not affect kernel behavior.

- Hardware Type: Apple M2 Pro (MacBook Pro 14, 2023) — macOS 15, dev env only,
  no GPU
- [ ] run `make test` to ensure correctness — N/A (metadata-only change)
- [x] run `make checkstyle` to ensure code style — `ruff check pyproject.toml`
      passes
[...]

PR #1332, opened 2026-08-02
Add requires-python >=3.10 to pyproject.toml

### LINKED ISSUE #1302: pyproject.toml has no requires-python — pip installs on any interpreter; publish workflows disagree (3.8 vs 3.10)
Hi, and thank you for Liger Kernel.

While packaging-reviewing the repo (at HEAD `72a4ed4`) I noticed `pyproject.toml` has **no `requires-python`** in `[project]`. Consequences:

- `pip` will happily resolve and install liger-kernel on any interpreter, however old — users on unsupported Pythons get runtime failures instead of a clean "requires Python >= X" resolver message.
- The two publish workflows don't agree on a floor either: `publish-nightly.yml` builds with Python **3.8**, `publish-release.yml` with **3.10** — so the effective supported floor is ambiguous even inside CI.

I'd have sent a one-line PR adding `requires-python = ">=3.10"`, but the floor is a support commitment only you can make (and it hard-blocks installs below it), so asking first: **what floor do you want?** Happy to send the PR (plus aligning the nightly workflow's Python) once you name it.

For transparency (per AGENTS.md): I used AI assistance to spot and draft this; I verified pyproject and both workflows myself.

yashb98:
## Summary

Fixes #1302.

`pyproject.toml` has no `requires-python`, so pip resolves and installs liger-kernel on any interpreter. Users on unsupported Pythons get runtime failures instead of a clean "requires Python >= X" resolver message.

This adds `requires-python = ">=3.10"`, matching the project's own floor: every CI test workflow (nvi-ci, amd-ci, checkstyle, benchmark) and the release-publish workflow run Python 3.10.

Picking up the hand-off from #1307 (closed by its author due to inactivity, fix noted as still valid).

## Testing Done

Metadata-only change. Verified the line parses and the floor matches CI:

```
$ python -c "import tomllib; print(tomllib.load(open('pyproject.toml','rb'))['project']['requires-python'])"
>=3.10
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,6 +9,7 @@ description = "Efficient Triton kernels for LLM Training"
 urls = { "Homepage" = "https://github.com/linkedin/Liger-Kernel" }
 readme = { file = "README.md", content-type = "text/markdown" }
 license = { file = "LICENSE" }
+requires-python = ">=3.10"
 dynamic = ["dependencies", "optional-dependencies"]
 
 [tool.setuptools]


yashb98:
Same as #1328, the workflow runs here are held at `action_required` and have never run. Could a maintainer approve them?

PR #1336, opened 2026-08-03
perf(cutedsl): fused dX+dW backward + host-enqueue caching for RMSNorm

Charlesrizzly:
## Summary

Follow-up optimization to the CuTe-DSL RMSNorm op added in #1299. That kernel's
own docstring lists **"fused dx/dw pass, vectorized/pipelined loads"** as future
work, and profiling shows its device kernels are already fast enough that
per-call **host (Python) launch overhead** dominates at benchmark shapes. This PR
delivers the fused backward and removes that launch overhead.

**No public-API or backend-registry change** — the cutedsl backend stays opt-in
via `LIGER_KERNEL_IMPL=cutedsl`.

## What changed (3 files)

- **`ops/cutedsl/ops/rms_norm.py`** — the optimization:
  - **Fused single-pass dX+dW backward**, made the default, with a **split
    fallback** for shapes the fast vector path can't serve and an env escape
    hatch (`LIGER_RMS_FORCE_SPLIT_BWD`).
  - **Host-enqueue caching:** `LIGER_RMS_*` env knobs hoisted to module
    constants (read once at import rather than ~12 `os.environ` lookups per
    fwd+bwd pass); a **bounded FIFO cache** of marshaled CuTe tensor handles
    keyed on `data_ptr` (`LIGER_RMS_TENSOR_CACHE_CAP`, default 8); a per-device
    `multi_processor_count` cache.
- **`ops/cutedsl/ops/rms_norm_fastpath.py`** *(new)* — two pure-Python fast-path
  eligibility helpers (vector width + Triton-parity warp count), split out so
  they are unit-testable without CUDA.
- **`test/transformers/test_cutedsl_rms_norm_fastpath.py`** *(new)* — CPU-only
  unit tests for those helpers.

## Numerics — equivalent to the merged kernel

Verified head-to-head against the current `main` kernel on identical inputs, the
fused backward matches the existing split path to within floating-point
reduction-order noise:

| dtype | Y | dX | dW |
|---|---|---|---|
| fp32  | ~1e-6            | ~1e-6   | ~6e-6            |
| bf16  | **bit-identical** | ≤ 3.9e-3 | **bit-identical** |

Every delta is far inside the parity suite's bf16 tolerance (atol 5e-2). The
existing correctness oracle `test/transformers/test_cutedsl_rms_norm.py` is
**left untouched**.

## Performance (repo harness, median µs, eager, bf16, hidden = 4096)

ratio = cutedsl / Triton, so **< 1.0 is faster**:

| GPU | seq | forward | backward |
|---|---:|---:|---:|
| **B200** (cc 10.0) | 8192 | **0.93×** (win) | **0.83×** (win) |
| B200 | 4096 | 1.00× | 1.05× |
| H100 (cc 9.0) | 8192 | 1.02× | 1.10× |
| H100 | 4096 | 1.41× | 1.60× |

A **Blackwell-favorable** play: on B200 cutedsl ties Triton at 4096 and **wins at
8192**, and wins universally under CUDA-graph replay (the device kernels were
never the bottleneck — host enqueue was). On H100 the device-kernel margin is
smaller, so eager stays parity-or-slightly-worse. Peak memory is ~1.5× Triton at
cache cap 8, architecture-independent.

## Testing

- `test/transformers/test_cutedsl_rms_norm.py` (existing parity oracle) +
  `test_cutedsl_rms_norm_fastpath.py` (new) — run with `LIGER_RMS_*` /
  `LIGER_KERNEL_IMPL` scrubbed from the env.
- `ruff check` / `ruff format --check` clean on all three files.

> **Note on B200/Blackwell:** on this specific arch the suite's bf16 `dW` parity
> cases exceed the bf16 tolerance. This reproduces **identically with the current
> `main` kernel** (verified head-to-head, same inputs) — i.e. a pre-existing
> Blackwell reduction-order gap that is orthogonal to this change, not a
> regression introduced here. fp32 and all forward/`dX` cases pass.

## Tuning knobs (all opt-in, safe defaults)

`LIGER_RMS_FORCE_SPLIT_BWD`, `LIGER_RMS_BACKWARD_WARPS`,
`LIGER_RMS_FUSED_STRIP_MULT`, `LIGER_RMS_TENSOR_CACHE_CAP`,
`LIGER_RMS_COMPILE_BUCKET`, `LIGER_RMS_FORCE_NO_FAST`,
`LIGER_RMS_RELOAD_POLICY`, `LIGER_RMS_AUTOTUNE_FILE`, `LIGER_RMS_DEBUG` — each
defaults to the fast behavior and can be trimmed if a narrower surface is
preferred.

---
*Authored with GitHub Copilot.*

### DIFF
--- a/src/liger_kernel/ops/cutedsl/ops/rms_norm.py
+++ b/src/liger_kernel/ops/cutedsl/ops/rms_norm.py
@@ -17,14 +17,22 @@ toggle, in-place backward, and DTensor gathering.
     dx   = rstd * [ m - (1/N) * rstd^2 * (m . x) * x ],   m = dy * (w + offset)
     dw   = sum_rows dy * (x * rstd)
 
-The kernels are written for correctness and clarity first (fp32 accumulation, one
-CTA per row for the reductions). The weight gradient is reduced in parallel across
-row strips (one strip per SM, partials summed on the host — the same strategy as the
-Triton backward) rather than a single thread walking every row. They keep the hidden
-dimension fully dynamic, so a single compiled kernel serves every shape — regular or
-irregular. Further performance tuning (fused dx/dw pass, vectorized/pipelined loads)
-is left to the ``liger-kernel-perf`` workflow on real Hopper/Blackwell hardware.
+The scalar forward and split backward kernels remain correctness fallbacks for
+irregular, unaligned, and unsupported shapes. Aligned widths through 8192 use
+vector paths: forward keeps X in registers across its reduction and output phase;
+affine backward mirrors Triton's execution shape with one CTA per SM, contiguous
+persistent row ranges, width-dependent warp counts, and register-resident X/dY
+fragments reused for both dX and dW. The fast path is selected host-side only when
+its alignment assumptions are true. Its compilation key therefore includes the
+exact width, vector width, and launch geometry; the scalar path remains
+shape-generic.
 """
+
+import json
+import os
+import time
+
+from collections import OrderedDict
 
 import cuda.bindings.driver as cuda
 import cutlass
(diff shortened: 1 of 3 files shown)


src/liger_kernel/ops/cutedsl/ops/rms_norm.py
```diff
@@ -173,6 +332,86 @@ def _rms_norm_fwd_kernel(
...
+    ELEMENTWISE_AFFINE: cutlass.Constexpr,
+    N_COLS: cutlass.Constexpr,
+    VEC: cutlass.Constexpr,
+    NUM_VEC_TILES: cutlass.Constexpr,
+):
+    """Aligned one-row forward kernel with register-resident X fragments."""
+    tid, _, _ = cute.arch.thread_idx()
+    lane = tid % 32
+    warp = tid // 32
+    row, _, _ = cute.arch.block_idx()
```
kolehma8, line 353: persistent kernels (e.g. have each CTA to process multiple rows) will likely produce a faster kernel than launching one CTA per row

src/liger_kernel/ops/cutedsl/ops/rms_norm.py
```diff
@@ -173,6 +332,86 @@ def _rms_norm_fwd_kernel(
...
+    gXv = cute.tiled_divide(gX, (VEC,))
+    gYv = cute.tiled_divide(gY, (VEC,))
+
+    # Keep every X fragment live through normalization.  N_COLS is constexpr on the
+    # fast path, so this is a compact fixed register tile (at most 16 vectors/thread).
+    x_frags = cute.make_rmem_tensor((VEC, NUM_VEC_TILES), mX.element_type)
+    out_frag = cute.make_rmem_tensor((VEC,), mY.element_type)
+    partial = Float32(0.0)
+    n_vec = N_COLS // VEC
+    for ct in cutlass.range_constexpr(NUM_VEC_TILES):
```
kolehma8, line 381: If this is sum over the hidden states, I don't think NUM_VEC_TILES can be constant expression as that would depend on the input. Not 100% sure of this as maybe in cuteDSL that is acceptable

src/liger_kernel/ops/cutedsl/ops/rms_norm.py
```diff
@@ -173,6 +332,86 @@ def _rms_norm_fwd_kernel(
...
+        vec_idx = ct * _FAST_THREADS + tid
+        if vec_idx < n_vec:
+            cute.autovec_copy(gXv[None, vec_idx], x_frags[None, ct])
```
kolehma8, line 384: doing the copy using int4 might be better if you are using raw CUDA threads. That will fully utilize the 4Kib page that one warp can read at the time.

src/liger_kernel/ops/cutedsl/ops/rms_norm.py
```diff
@@ -173,6 +332,86 @@ def _rms_norm_fwd_kernel(
...
+    gYv = cute.tiled_divide(gY, (VEC,))
+
+    # Keep every X fragment live through normalization.  N_COLS is constexpr on the
+    # fast path, so this is a compact fixed register tile (at most 16 vectors/thread).
+    x_frags = cute.make_rmem_tensor((VEC, NUM_VEC_TILES), mX.element_type)
+    out_frag = cute.make_rmem_tensor((VEC,), mY.element_type)
+    partial = Float32(0.0)
+    n_vec = N_COLS // VEC
+    for ct in cutlass.range_constexpr(NUM_VEC_TILES):
+        vec_idx = ct * _FAST_THREADS + tid
```
kolehma8, line 382: might be worthwhile to experiment with a design where each warp handles separate tokens. We usually have at least ~2000 tokens for training (typically more than that). With 128 CTAs and 8-16 warps for each, you can process that in single instance. That should also saturate HBM bandwidth which is likely the bottle neck here.

test/transformers/test_cutedsl_rms_norm_fastpath.py
```diff
@@ -0,0 +1,42 @@
+"""CPU-only checks for CuTe DSL RMSNorm fast-path launch policy."""
```
kolehma8, line 1: is there any numerical tests here?

PR #1338, opened 2026-08-04
fix: torch.compile loss corruption at ignore_index rows

KyleMylonakisProtopia:
## Summary

`LigerFusedLinearCrossEntropy` returns a **silently wrong loss under `torch.compile`** whenever the
targets contain at least one `ignore_index` position. There is no exception, warning, or graph-break
error — the compiled graph runs to completion and returns a numerically incorrect value. An all-valid
mask (zero ignored positions) does **not** reproduce it.

```python
import torch
from liger_kernel.transformers.fused_linear_cross_entropy import LigerFusedLinearCrossEntropyLoss

torch.manual_seed(0)
BT, H, V = 256, 32, 4096
x = torch.randn(BT, H, device="cuda")
w = torch.randn(V, H, device="cuda")
target = torch.randint(0, V, (BT,), device="cuda")
target[torch.randperm(BT)[: BT // 4]] = -100        # some ignored targets

ce = LigerFusedLinearCrossEntropyLoss(ignore_index=-100)
call = lambda x, t: ce(w, x, t)

with torch.no_grad():
    print(call(x, target).item())                        # 21.404566
    print(torch.compile(call)(x, target).item())         # 61.734364  <-- wrong
```

Reproduces with both plain `torch.compile(fn)` and `torch.compile(fn, fullgraph=True)`.

`LigerFusedLinearJSD` carried the identical pattern when this PR was opened. It is **no longer
reachable on current `main`** — see “What changed since this PR was opened” below — but the
kernel-side half of the fix is retained there as defense-in-depth.

## What changed since this PR was opened

`main` has moved under this PR, and three claims in the original description no longer hold. The net
diff is now **6 files, +160 / -6**.

**1. The `BT x V` loss-buffer perf commit (`4b77c6b`) is superseded.** `main` landed the identical
scalar-accumulator change independently, as “Optimization #1” in `fused_linear_jsd_forward`. Merging
brought it in, so that commit now contributes nothing and the memory-reduction table that used to be
in this description has been removed. No functionality is lost — `main`'s version has the same
effect, and the reduction-order caveat that used to be flagged here now belongs to `main`'s change
rather than this PR.

**2. The JSD `torch.compile` bug is no longer reachable, so that half is now defense-in-depth.**
`fused_linear_jsd_forward` reaches `_jsd_kernel` through `liger_kernel.backends.dispatch`, which is
decorated [`@torch.compiler.disable`](https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/backends/dispatch.py).
That graph break keeps the Triton kernel out of Inductor entirely, so the functionalization clone
that caused the over-read never happens on the default path. Verified by reverting the `jsd.py`
change on the merged branch: all 9 JSD ignore/compile tests still pass. The change is kept anyway —
it costs two `tl.store`s on a path that already returns early, and it is the only thing that keeps
the contract true if the kernel is ever lowered by Inductor again (a different `jsd_impl`, or the
`torch.compiler.disable` being lifted).

**3. The JSD regression test can no longer use `fullgraph=True`.** Same cause: requesting a full
graph now raises `Unsupported: Skip calling torch.compiler.disable()d function`. That test asserts
the eager-vs-compiled numeric contract under plain `torch.compile` instead, and its docstring says
so. CE is unaffected — it bypasses `dispatch` when `ce_impl is None`, and its `fullgraph=True` test
still actively reproduces the bug.

**The CE half is unchanged in substance and still fixes a live bug.** `main`'s CE refactor introduced
an `if ce_impl is None:` fast path that reintroduced the offset-view pattern this PR removes, so the
fix was re-applied inside the new structure. `main`'s own dispatched branch already allocates the way
this PR does, so the two branches are now consistent rather than one being a special case.

### Root cause

`fused_linear_cross_entropy_forward` passes `loss_1d[start_idx:end_idx]` — a view with a **non-zero
storage offset** — as a *mutated* pointer to a raw `triton.jit` kernel.

`torch.compile` functionalizes that mutation with
[`clone_preserve_strides`](https://github.com/pytorch/pytorch/blob/main/torch/_prims_common/__init__.py#L2168),
which must clone `storage_offset + numel` elements so the kernel's pointer arithmetic still lands
correctly:

```python
needed_size = compute_required_storage_length(x.size(), x.stride(), x.storage_offset())
buffer = torch.as_strided(x, (needed_size,), (1,), 0).clone()
return torch.as_strided(buffer, x.size(), x.stride(), x.storage_offset())
```

That is correct in eager, where the real base storage is large enough. But Inductor may realize the
view as a buffer sized to the **view alone**, and then lowers the `as_strided` as an unmasked copy —
reading past the end. From `TORCH_LOGS=output_code` on the repro above (`BT=256, H=32, V=4096`, which
chunks into 8 x 32 rows), the clone sizes are exactly `needed_size` for each chunk while every source
slice holds only 32 elements:

```
%as_strided_default_14 : "f32[32][1]"  = aten.as_strided.default(%slice_3,  [32],  [1], 0)
%as_strided_default_12 : "f32[64][1]"  = aten.as_strided.default(%slice_11, [64],  [1], 0)
%as_strided_default_10 : "f32[96][1]"  = aten.as_strided.default(%slice_19, [96],  [1], 0)
%as_strided_default_8  : "f32[128][1]" = aten.as_strided.default(%slice_27, [128], [1], 0)
%as_strided_default_6  : "f32[160][1]" = aten.as_strided.default(%slice_35, [160], [1], 0)
%as_strided_default_4  : "f32[192][1]" = aten.as_strided.default(%slice_43, [192], [1], 0)
%as_strided_default_2  : "f32[224][1]" = aten.as_strided.default(%slice_51, [224], [1], 0)
%as_strided_default    : "f32[256][1]" = aten.as_strided.default(%slice_59, [256], [1], 0)
```

The last chunk reads 224 elements past the end of a 32-element buffer, and the generated pointwise
kernels are unmasked (`xmask = tl.full([XBLOCK], True, tl.int1)`).

That over-read is normally harmless, because the kernel overwrites the whole slice. **But on an
`ignore_index` row `liger_cross_entropy_kernel` returns early without writing `loss_ptr` (or
`z_loss_ptr`)**, so the garbage survives into `torch.sum(loss_1d)`.

Both preconditions are required, which is exactly why an all-valid mask is unaffected — every
`loss_ptr` gets fully written, so the garbage is always overwritten.

Confirming it is foreign memory rather than a logic error: whether a given shape reproduces depends
on **allocator state**, not on the math. At `BT=256, H=32, V=4096` a `randperm` ignore mask
reproduces (`21.40 -> 61.73`) while a regular `target[::4] = -100` mask does not, because the memory
immediately past the realized buffer happens to be zeros in the latter case. `compute-sanitizer`
reports nothing, because the over-read stays inside the caching allocator's segment.

### Measured

Re-measured on the merged branch, so these reflect current `main`'s CE chunk sizing
(`_CHUNK_MEM_CONST = 16`, which produces far fewer/larger chunks than when this PR was opened — the
old table's `nan` values are no longer the observed failure mode at these shapes):

| shape (`BT`, `H`, `V`) | chunks | eager | `compile` before | `fullgraph` before | after (both) |
| --- | --- | --- | --- | --- | --- |
| 256, 32, 4096 | 8 | 21.404566 | **61.734364** | **45.291565** | 21.404568 |
| 1024, 64, 32000 | 32 | 33.239548 | 33.239548 | **37.219460** | 33.239548 |
| 1024, 64, 32000, `z_loss` | 32 | 33.239548 | **33.380241** | **33.649807** | 33.239548 |
| 4096, 128, 32000 | 16 | 46.831673 | **51.175064** | **55.405586** | 46.831673 |
| 2048, 256, 128000 | 32 | 71.590370 | **72.349937** | **85.460114** | 71.590370 |

9 of these 10 compiled configurations were wrong before this change; 0 of 10 after.

### The fix

Two layers, **each independently sufficient** — verified by reverting one at a time on the merged
branch, where either alone gives `4 passed` on the CE regression test and reverting both gives
`3 failed, 1 passed`:

1. **Allocate standalone contiguous per-chunk output buffers** instead of views, so
   `needed_size == numel` and the functionalization clone is exact. This also makes the pre-existing
   `loss_1d[start_idx:end_idx] = loss_1d_slice` write-back meaningful rather than a self-assignment.
2. **Write zeros to the loss pointers on the ignore path**, so the kernel does not rely on the caller
   having zeroed a buffer it does not own. `_tv_distance_kernel` already does exactly this, so this
   brings CE and JSD in line with existing convention.

Eager numerics are unchanged.

### Not changed

- **The backward pass is unaffected.** It only ever hands `element_mul_kernel` whole tensors at
  storage offset 0, so the clone is exact, and that kernel has no early-return path. Verified in
  isolation under Inductor lowering: fed whole tensors it matches eager exactly (maxdiff `0.0`); fed
  offset views the same kernel diverges by `9.286`, confirming the calling convention is the
  discriminator. End-to-end gradients also match eager with `ignore_index` present, for both
  `grad_output == 1.0` (where `element_mul_kernel` is skipped) and `grad_output == 2.5` (where it
  runs).
- **The Ascend backend** carries the same pattern in
  [`_ascend/ops/fused_linear_cross_entropy.py`](https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/ops/backends/_ascend/ops/fused_linear_cross_entropy.py)
  (offset views) and its generic CE kernel's ignore path, which writes `lse` / `token_accuracy` /
  `predicted_tokens` but not `loss_ptr` / `z_loss_ptr`. Its JSD kernel already zeroes `loss_row_ptr`.
  I have no NPU to test on, so I left it alone rather than ship an unverified change to a hardware
  backend — flagging it here for someone with access.
- **Arguably also an upstream Inductor bug**, since it lowers `as_strided` beyond a realized buffer's
  extent as an unmasked read. Worth reporting separately; this change removes Liger's dependence on
  that behavior either way.

### Adjacent issues found, worth separate tickets

- **`@torch.compiler.disable` on `dispatch` makes every dispatched op non-`fullgraph`-compilable.**
  This is what removed JSD's `fullgraph` coverage (point 2 above). It is presumably deliberate — the
  registry lookup is Python-level — but it is worth a conscious decision, since it silently converts
  `fullgraph=True` from “works” to “raises” for every op routed through the dispatcher, and it means
  the dispatched kernels never get Inductor-lowered.
- Both backward functions branch on a tensor value (`if torch.ne(grad_output, ...)` /
  `torch.equal`), so `fullgraph=True` cannot capture the backward pass. Relatedly,
  `torch._dynamo.compiled_autograd` fails outright on both with
  `aten::_local_scalar_dense() ... Unable to cast 1.0 to Tensor`.

## Testing Done

- Hardware Type: **NVIDIA GB10 (compute capability 12.1, 48 SMs)** — `torch 2.13.0+cu130`,
  `triton 3.7.1`, Python 3.12.3
- [ ] run `make test` to ensure correctness — **in progress** on the merged branch; will post the
      full numbers when it finishes. The targeted suites below are already re-run post-merge.
- [ ] run `make test-convergence` to ensure convergence — queued behind the `make test` re-run

Everything below was re-run on the merged branch (`6a770f5`).

### Targeted regression tests

```
12 passed   # both test_correctness_with_torch_compile sets
463 passed  # test_cross_entropy.py test_fused_linear_cross_entropy.py test_jsd.py test_fused_linear_jsd.py
352 passed, 49 skipped  # test/backends test/ops -- main's dispatcher machinery
```

Layer isolation on the merged branch, confirming the CE test still actively reproduces the bug:

| CE regression test | result |
| --- | --- |
| both layers present | `4 passed` |
| layer 1 reverted (kernel zeroing only) | `4 passed` |
| layer 2 reverted (standalone buffers only) | `4 passed` |
| both layers reverted | **`3 failed, 1 passed`** |

For JSD the same experiment yields `9 passed` with the kernel change reverted, which is the
measurement behind point 2 above — the JSD path is no longer exposed to the bug.

### `make checkstyle`

```
All checks passed!
446 files already formatted
```


KyleMylonakisProtopia:
Any possibility of landing this? It would help overcome the bug in pytorch (https://github.com/pytorch/pytorch/pull/192227)

KyleMylonakisProtopia:
Any possibility of landing this?

KyleMylonakisProtopia:
Some changes were introduced into main which resolved some, but not all of the bugs found in this PR. Can we get the rest of these in?

KyleMylonakisProtopia:
> Any possibility of landing this? It would help overcome the bug in pytorch ([pytorch/pytorch#192227](https://github.com/pytorch/pytorch/pull/192227))

The pull request in Pytorch has landed. If its included in the next release of Pytorch, this issue would be resolved for that version but present on all other versions. Is there any possibility of landing this?

PR #1339, opened 2026-08-04
Add SM100 CuTe DSL fused linear cross entropy

justinhh4:
## Summary

Adds a reusable CuTe DSL SM100 GEMM and uses it to implement fused linear cross
entropy (FLCE) for NVIDIA B100/B200/GB200.

The shared GEMM provides:

- aligned and row-padded output views for bounded FLCE workspaces;
- a single-fragment custom epilogue entry point for future fused operators;
- shared TMA/UMMA pipelines, persistent scheduling, TMEM management,
  stream/device handling, compile serialization, and TVM-FFI/DLPack launch paths.

FLCE computes each bounded logits chunk once, transforms that storage into
`dlogits` in place with the CuTe DSL cross-entropy kernel, and immediately forms
`grad_input`, `grad_weight`, and optional `grad_bias`. The first autograd
backward hands off those precomputed gradients without recomputing logits or
allocating a duplicate parameter-sized gradient.

At the Llama 3-sized reference shape (`BT=8192`, `H=4096`, `V=128256`, BF16),
full forward + backward is **6.96x faster** than upstream Liger Triton FLCE and
uses **2.17x less peak allocated memory** on B200.

## Implementation

- Native execution requires exact compute capability `(10, 0)` and FP16/BF16
  operands.
- The architecture gate uses the input tensor's device index and
  `liger_kernel.utils.infer_device_arch`.
- The native path uses one bounded 512/1024-row workspace; a full `(BT, V)`
  logits tensor is never materialized.
- Cross entropy overwrites each logits chunk with normalized `dlogits`, which
  is consumed immediately by the input-, weight-, and bias-gradient
  projections.
- Hidden dimensions not aligned to the native K tile are zero-padded internally
  and sliced back after backward.
- Odd vocabulary sizes use an aligned physical workspace stride while the
  logical vocabulary size excludes padding from softmax mass, label smoothing,
  and argmax.
- The single native path supports bias, class weights, label smoothing, z-loss,
  softcap, token scaling, token accuracy, predicted tokens, `mean`/`sum`
  reduction, partial/frozen-gradient inputs, and configurable gradient
  accumulation dtype.
- This implementation is native-only: non-SM100 hardware, non-FP16/BF16
  operands, and `reduction="none"` raise an explicit error. Frontend fallback
  and cross-backend routing are intentionally outside this PR.
- Retained-graph repeated backward recomputes only after the first precomputed
  gradient handoff has been consumed.

## Scope

[...]

PR #1340, opened 2026-08-04
Add SM100 fused linear SwiGLU

justinhh4:
## Summary

Adds a unified CuTe DSL SwiGLU implementation on top of the current `main`
SM100 GEMM:

- a differentiable elementwise forward/backward replacement for
  `LigerSiLUMulFunction`;
- an exact-SM100 fused-linear forward path that computes gate and up projections
  in one grouped GEMM and applies SwiGLU in the epilogue without materializing
  either projection, reading a plainly **stacked** (or **separate**) gate/up
  weight directly with no pre-packing and no extra weight copy;
- safe fallback to two linear projections plus the elementwise CuTe DSL kernel
  for unsupported architectures, dtypes, and K alignment.

This PR is rebased directly onto `main`; the former dependency on #1339 has
been removed.

## Implementation

- The fused path consumes a plainly **stacked** `[2N, K]` weight (gate rows
  `[0:N]` then up rows `[N:2N]`) directly: `run_grouped_epilogue_gemm_stacked`
  rebuilds the B tensor with a nested `((32, 2, T):(K, N·K, 32·K))` N-layout so
  gate/up for the same 32 output features still land as adjacent accumulator
  fragments local to the same SM100 CTA — **without any physical repack or a
  second weight copy**. The MMA mainloop and the FP32 TMEM epilogue are
  unchanged from the interleaved-pack version.
- The grouped custom-epilogue mode is integrated into the current reusable
  `_sm100_gemm.py` without restoring the retired FLCE row-reduction paths.
- The existing single-fragment FLCE configuration remains unchanged.
- Native fused execution requires exact compute capability `(10, 0)`,
  FP16/BF16 operands, and K divisible by 64.
- Unsupported inputs run two `torch.nn.functional.linear` calls followed by the
  CuTe DSL elementwise kernel (separate tensors are used directly here, no copy).
- Elementwise and fused kernels launch under the input tensor's CUDA device
  guard and on PyTorch's active CUDA stream for both TVM-FFI and DLPack calling
  paths.

## Weight API

```python
# Stacked (zero-copy): gate rows [0:N] then up rows [N:2N]
# e.g. a model's fused gate_up_proj.weight, or torch.cat([gate, up], dim=0)
h = fused_swiglu(x, gate_up_weight)

# Separate tensors: concatenated once internally for the native path
h = fused_swiglu(x, gate_weight, up_weight)
```

- The **stacked** form is read as a re-strided view of a single buffer, so there
  is **no pre-pack step and no transient 2× weight copy** (previously incurred by
  the 32-row interleave). This resolves the review feedback on the weight-packing
  memory spike.
- **Separate** `gate`/`up` tensors are accepted for convenience; the native path
  concatenates them once (a strided view cannot span two independent
  allocations), while the fallback path consumes them with no copy. Prefer the
  stacked form on hot paths.
- Because the weight stays a standard `[2N, K]` (or two `[N, K]`) tensor, it is
  directly usable as an `nn.Parameter` and is checkpoint-/shard-friendly.
- `output_features` is **keyword-only** and inferred as `N` when omitted;
  trailing padded columns (when `N % 32 != 0`) are trimmed automatically.
- `pack_swiglu_weights` is retained as a thin `torch.cat` convenience helper.

## B200 optimization

NCU at `M=8192, H=4096, N=14336` classified the grouped kernel as
compute-bound:

- compute throughput: **96.07%**
- memory throughput: **62.11%**
- registers/thread: **96**
- dynamic shared memory/block: **213.25 KiB**
- achieved occupancy: **8.63%**

A manual pipeline-stage/swizzle sweep selected:

- five A/B stages at 1K tokens;
- six stages from 2K through 16K;
- four stages with the existing eight-way swizzle from 32K upward.

The large-M specialization raises normalized speedup from `1.015x`/`0.981x`
at 32K/64K to `1.053x`/`1.052x`.

## BF16 benchmarks

NVIDIA B200, `H=4096`, `N=14336`, seven warmups, 21 interleaved samples. The
stacked weight is read as a zero-copy view, so nothing is excluded from latency
on the fast path (separate tensors incur a one-time `torch.cat`). Peak memory is
measured in isolated provider processes and includes input, one weight
representation, output, and temporary PyTorch CUDA allocations.

| Tokens | Two GEMMs + SwiGLU (ms) | Fused SwiGLU (ms) | Speedup | Baseline peak (MiB) | Fused peak (MiB) | Memory ratio |
|---:|---:|---:|---:|---:|---:|---:|
| 1,024 | 0.214208 | 0.223136 | 0.960x | 348 | 260 | 1.34x |
| 2,048 | 0.355776 | 0.356896 | 0.997x | 440 | 296 | 1.49x |
| 4,096 | 0.691104 | 0.647360 | 1.068x | 624 | 368 | 1.70x |
| 8,192 | 1.285664 | 1.217984 | 1.056x | 992 | 512 | 1.94x |
| 16,384 | 2.565120 | 2.412224 | 1.063x | 1,728 | 800 | 2.16x |
| 32,768 | 5.130048 | 4.873984 | 1.053x | 3,200 | 1,376 | 2.33x |
| 65,536 | 10.174336 | 9.672512 | 1.052x | 6,144 | 2,528 | 2.43x |

## Correctness

- CuTe DSL SwiGLU, upstream elementwise SwiGLU, SM100 GEMM, and FLCE regression
  suites: **301 passed, 1 skipped**
- Stacked zero-copy path, separate-weight path (byte-identical to stacked), and
  weight-API validation are covered in `test/cutedsl/test_swiglu.py` (verified
  alongside `test/cutedsl/test_sm100_gemm.py` on B200 SM100)
- Current-stream coverage: elementwise forward, elementwise backward, and
  fused-linear forward
- FP16/BF16 fused correctness, padded output tails, exact-SM100 dispatch, FP32
  fallback, arbitrary elementwise shapes, multipliers, and PyTorch/Triton parity
- `make checkstyle`: **passed**

Validation environment:

- NVIDIA B200, SM100, driver `580.105.08`
- PyTorch `2.13.0.1+cu130`
- Triton `3.7.1`
- `nvidia-cutlass-dsl` `4.6.0`
- `apache-tvm-ffi` `0.1.12`

## Scope

- `src/liger_kernel/ops/cutedsl/ops/swiglu.py`
- `src/liger_kernel/ops/cutedsl/ops/_sm100_gemm.py`
- `test/cutedsl/test_swiglu.py`

### DIFF
--- a/src/liger_kernel/ops/cutedsl/ops/swiglu.py
+++ b/src/liger_kernel/ops/cutedsl/ops/swiglu.py
@@ -1,0 +1,943 @@
+"""CuteDSL (NVIDIA CUTLASS Python DSL) implementation of the SwiGLU activation.
+
+This is a drop-in alternative to the Triton kernels in
+``liger_kernel.ops.swiglu``. It computes the element-wise SwiGLU gate
+
+    forward:   c  = silu(a * gate_multiplier) * b
+    backward:  da = dc * (silu(a') * (1 - sig(a')) + sig(a')) * b * gate_multiplier
+               db = dc * silu(a')          with a' = a * gate_multiplier
+
+where ``silu(x) = x * sigmoid(x)``.
+
+Design notes (tuned on B200 / sm_100):
+  * The problem is memory-bound, so the kernel is a flat element-wise pass that
+    issues a single **128-bit (vectorized) load/store** per thread per tensor.
+    ``vec = 128 // dtype.width`` contiguous elements are processed per thread.
+    Vectorization is obtained with an *ordered* thread/value layout
+    (``make_ordered_layout((1, NT), order=(1, 0))`` + ``val=(1, vec)``) and a
+    copy atom carrying ``num_bits_per_copy = vec * dtype.width``; this is the
+    same recipe the tuned ``rms_norm`` CuteDSL kernel uses and is what lifts
+    achieved DRAM throughput past the Triton baseline on B200.
+  * The sigmoid math runs in fp32 using Blackwell **packed-f32x2** SFU ops
+    (``mul_packed_f32x2`` / ``add_packed_f32x2`` / ``exp_packed_f32x2``, which
+    process two fp32 lanes per instruction) plus ``rcp_approx`` for the
+    reciprocal. This roughly halves the issued math instructions versus the
+    scalar ``1/(1+exp2(-x))`` form, which profiling showed to be issue-bound.
+  * The *fast* path compiles **once per dtype** against an abstract
+    (fake) tensor with a dynamic batch dimension and is invoked with
+    ``--enable-tvm-ffi`` so PyTorch tensors are passed **directly** to the
+    compiled function. This removes the per-call ``from_dlpack`` /
+    memref-construction host overhead (~26us -> ~4us per launch on B200).
+    It is used whenever ``numel`` is a multiple of the CTA tile
+    (``_NUM_THREADS * vec``) -- i.e. every realistic workload.
+  * A *general* predicated fallback handles arbitrary ``numel`` for odd shapes:
+    it uses scalar per-element predicated loads/stores (128-bit vectorized copies
+    are incompatible with per-element bounds masking) but still applies
+    **packed-f32x2** SFU math when available — the predicated load zeroes OOB
+    elements, so pairwise computation on them is safe and the predicated store
+    prevents writing them back. When TVM-FFI is available the fallback is also
+    compiled with ``--enable-tvm-ffi`` against a 1-D ``sym_int`` fake tensor so
+    PyTorch tensors are passed directly (no ``from_dlpack`` overhead); the 1-D
+    shape needs no divisibility because the reshape to ``(rows, tile)`` is only
+    required by the vectorized copy atom, not by TVM-FFI itself.
+  * Backward writes the gradients in place into the saved ``a`` / ``b`` buffers,
+    exactly like the Triton kernel, so peak memory matches.
+
+The element-wise public API mirrors ``liger_kernel.ops.swiglu``:
+``swiglu_forward`` / ``swiglu_backward`` / ``LigerSiLUMulCuteDSLFunction``.
+``fused_swiglu`` additionally fuses the gate/up projections on exact SM100
+Blackwell GPUs and delegates to the element-wise implementation elsewhere.
+"""
+
+import functools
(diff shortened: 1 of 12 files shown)


src/liger_kernel/ops/cutedsl/ops/swiglu.py
```diff
@@ -628,14 +668,135 @@ def fake():
...
+    run_grouped_epilogue_gemm(
+        a.contiguous(),
+        packed_gate_up_weight.contiguous(),
+        out,
+        _fused_swiglu_epilogue,
+    )
+    return out[:, :output_features]
+
+
+def fused_swiglu(a, packed_gate_up_weight, output_features=None):
```
kolehma8, line 785: The up-down weights may not be packed in all implementations. Packing them will cause memory spikes when the packing is performed. The other way around it will work fine as you can easily "unpack" the two tensors as long as they are simply stacked.

src/liger_kernel/ops/cutedsl/ops/swiglu.py
```diff
@@ -628,14 +668,135 @@ def fake():
...
+
+def _fused_swiglu_sm100(a, packed_gate_up_weight, output_features):
+    padded_features = packed_gate_up_weight.shape[0] // 2
+    out = torch.empty(
+        a.shape[0],
+        padded_features,
+        device=a.device,
+        dtype=a.dtype,
+    )
+    run_grouped_epilogue_gemm(
```
kolehma8, line 776: Can you instead of doing the both up gates in separate GEMMs to do them side-by-side? The benefit of this is two fold: 1) you get logical "multi-cast" for the input operand X as it's used in both up projections (this will not increase the flops a lot); 2) the epilogue can happen in FP32 without any down-casting or HBM traffic as the both up gate projections are already in the TMEM as opposed to HBM.

src/liger_kernel/ops/cutedsl/ops/swiglu.py
vaibhavjindal: Fixed it.

src/liger_kernel/ops/cutedsl/ops/swiglu.py
vaibhavjindal: We are already doing this in this kernel.

kolehma8 (approved):

PR #1341, opened 2026-08-05
[NPU] Fix: recompute true-class CE gradient in fp32 on no-weight backward path

zheliuyu:
## Summary

PR #1329 added `test_correctness_true_class_grad_confident_predictions`, which compares the true-class logit gradient against an fp32 reference and requires Liger to be at least as accurate as PyTorch’s low-precision backward. On Ascend NPU, three cases fail when `label_smoothing=0.0` (`sum`/`mean`, bf16/fp16).

For plain cross-entropy (`no weight`, `no softcap`, `no z-loss`, `label_smoothing=0`), the Ascend backend uses `liger_cross_entropy_backward_kernel_no_weight`. That path wrote `softmax(x_y) * scale` into the low-precision `dX` buffer and then corrected the true-class entry with a read-modify-write (`load dX[y] - scale`). When `softmax(x_y) → 1`, rounding in bf16/fp16 before subtracting the `-1/N` term inflates error—the same failure mode fixed on GPU in #1329.

This change recomputes `dx_y = (softmax(x_y) - 1) * scale` once in fp32 from `X[y]` and `lse`, then overwrites `dX[y]`, matching the intent of the upstream Triton fix while keeping the existing O(V) loop unchanged.

## Performance

cross_entropy_speed_full_token_length

cross_entropy_memory_full_token_length

## Testing

```python
pytest test/transformers/test_cross_entropy.py 
```

image

- Hardware Type: All NPUs

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
@@ -348,6 +348,7 @@ def liger_cross_entropy_backward_kernel_no_weight(
                     # Per-row loss was stored without mean scaling (``reduction`` ``sum`` or ``none``).
                     lse = loss_row + x_y
 
+            ori_X_y = tl.load(X_ptr + X_ptr_offset + y).cast(tl.float32)
             for i in range(0, n_cols, BLOCK_SIZE):
                 X_offsets = i + tl.arange(0, BLOCK_SIZE)
                 X_block = tl.load(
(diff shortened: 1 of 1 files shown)


zheliuyu:
@Tcc0403 This PR is ready for review. Thanks.

Tcc0403 (approved):
LGTM

zheliuyu:
Hi @Tcc0403 — thanks again for reviewing #1341 and #1346. Both are approved and CI is green on our NPU runs. Could you help merge them when you have time? Happy to address any follow-ups if needed.

PR #1349, opened 2026-08-07
perf(cutedsl): optimize RMSNorm CuTe DSL kernel (fused backward + host launch + width-aware forward warps)

Charlesrizzly:
## Summary

Optimizes the CuTe DSL RMSNorm kernel introduced in #1299 on both passes:

1. **Fused backward + host-side launch** — fuses the RMSNorm backward into a
   single `@cute.kernel` and streamlines the host launch path (cached tensor
   marshaling, env knobs read once at import, a width/alignment-gated vector
   fast path), cutting backward-pass kernel time substantially.
2. **Width-aware forward warp count** — picks the vector-forward warp count from
   the row width (8 warps when a row has ≥ 512 vectors, e.g. bf16 hidden ≥ 4096,
   else 4) so wide rows saturate HBM bandwidth and lift SM occupancy
   (~55% → ~88%), while narrow rows stay at 4 warps to avoid idle threads.

Behavior is a drop-in match for the Triton implementation
(`liger_kernel.ops.rms_norm`): identical public `rms_norm_forward` /
`rms_norm_backward` / `LigerRMSNormFunction` signatures, the same math and
`llama` / `gemma` / `none` casting modes, offset, elementwise-affine toggle,
in-place backward, and DTensor gathering. A new `rms_norm_fastpath.py` holds the
(pure-Python, unit-tested) vector-width and warp-count heuristics.

## Details

**Fused affine backward:** one CTA per SM, contiguous persistent row ranges,
register-resident X/dY fragments reused for both dX and dW (mirrors the Triton
kernel's execution shape). Scalar / split paths remain as correctness fallbacks
for irregular, unaligned, and unsupported shapes.

**Host-side launch:** input-handle caching, env knobs read once at import
(removing per-launch `os.environ` lookups), and a compile cache keyed on the
exact width / vector width / launch geometry / warp count.

**Width-aware forward warps:** `fwd_warp_count(n_cols, vec)` returns
`8 if (n_cols // vec) >= 512 else 4`. The vector-forward kernel and its host
launch take `NUM_WARPS` / `NUM_THREADS` as constexprs; `num_vec_tiles` is
derived from the chosen thread count.

Measured vs the Triton kernel (pure op-level kernel time):

| Pass | B200 | H100 |
|------|------|------|
| Forward (wide rows) | 1.09–1.18x (8192×4096, 16384×4096, 8192×8192) | parity (already HBM-saturated) |
| Backward | favorable at large M | 1.12–1.52x |
| Full fwd+bwd | parity-to-favorable | 1.13–1.33x |

Forward is at parity for ≤ 4096-wide rows (already efficient at 4 warps). A B200
warp sweep confirms 8 warps is optimal for wide rows (12 / 16 within < 0.2% noise).

## Testing Done

- `pytest test/transformers/test_cutedsl_rms_norm.py test/transformers/test_cutedsl_rms_norm_fastpath.py`
  → **106 passed** (parity across llama/gemma/none casting, bf16 & fp32, offset,
  in-place, elementwise-affine variants), run against current `main`.
- `ruff check` + `ruff format --check` clean on all changed files.

- Hardware Type: NVIDIA B200 (also validated on H100 80GB HBM3)

### DIFF
--- a/src/liger_kernel/ops/cutedsl/ops/rms_norm_fastpath.py
+++ b/src/liger_kernel/ops/cutedsl/ops/rms_norm_fastpath.py
@@ -16,3 +16,19 @@ def triton_backward_warp_count(n_cols: int) -> int:
     if n_cols > 1024:
         return 8
     return 4
+
+
+def fwd_warp_count(n_cols: int, vec: int) -> int:
+    """Width-aware warp count for the vector forward kernel.
+
+    ``vec`` is the number of elements per 16-byte load, so ``n_cols // vec`` is the
+    number of vectors in a row. Wide rows (>= 512 vectors, e.g. bf16 hidden >= 4096)
+    profit from 8 warps: spreading the row over twice the threads halves the
+    register-resident vector tiles per thread, which lifts occupancy from ~55% to
+    ~88% and hides the DRAM-load latency that bounds this memory-bound kernel
+    (measured 4-7% faster at bf16 hidden 4096/8192, and faster still on tall inputs).
+    Narrow rows keep 4 warps: 8 warps would leave half the CTA idle (fewer vectors
+    than threads) and make the cross-warp reduction the bottleneck (up to ~30% slower
+    at bf16 hidden 1024/2048).
+    """
+    return 8 if (n_cols // vec) >= 512 else 4
--- a/src/liger_kernel/ops/cutedsl/ops/rms_norm_fastpath.py
+++ b/src/liger_kernel/ops/cutedsl/ops/rms_norm_fastpath.py
@@ -9,7 +9,7 @@ def fast_path_vector_width(*element_sizes: int) -> int:
     return 16 // largest
 
 
-def triton_backward_warp_count(n_cols: int) -> int:
+def backward_warp_count(n_cols: int) -> int:
     """Mirror calculate_settings() for hidden widths supported by the fast path."""
     if n_cols > 4096:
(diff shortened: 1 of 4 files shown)


src/liger_kernel/ops/cutedsl/ops/rms_norm_fastpath.py
```diff
@@ -0,0 +1,34 @@
...
+
+def fast_path_vector_width(*element_sizes: int) -> int:
+    """Return the common 16-byte vector width for all participating tensors."""
+    largest = max(element_sizes)
+    if largest <= 0 or 16 % largest:
+        raise ValueError(f"element sizes must divide 16, got {element_sizes}")
+    return 16 // largest
+
+
+def triton_backward_warp_count(n_cols: int) -> int:
```
kolehma8: lets not reference triton in the cutedsl sub-module

Charlesrizzly: Done in e2cbd93 — renamed to `backward_warp_count` (symmetric with `fwd_warp_count`). That was the only Triton-named symbol in the `cutedsl` submodule; `rms_norm_fastpath.py` now has zero `triton` references. Pure rename, no behavior change — 106/106 RMSNorm tests pass and ruff is clean.

kolehma8 (approved):

Issue #1330, opened 2026-08-02
[B300/SM103] Compiled Liger SwiGLU is ~3.1x slower than torch.compile for a Qwen3-shaped BF16 tensor


jessiewei7:
## 🐛 Describe the bug

On an NVIDIA B300 (SM103), the Blackwell Liger SwiGLU path is substantially
slower than the equivalent `torch.compile` expression for a Qwen3-shaped BF16
tensor.

With shape `(3072, 12288)`, including forward and backward:

| implementation | mean step time |
|---|---:|
| `torch.compile(lambda a, b: F.silu(a) * b)` | 2.169 ms |
| `torch.compile(lambda a, b: LigerSiLUMulFunction.apply(a, b))` | 6.831 ms |

The Liger path takes approximately 3.15x as long. The result reproduces on the
current Liger main (`91ae44ae659ebaf40d4a851777f4f02515e9ba65`), so it does
not appear to be specific to the 0.8.1 release.

This was initially observed in Qwen3-8B DFlash training. At sequence length
8192 and `max_anchors=3072`, enabling Liger increased the end-to-end training
step from 176.20 ms to 223.37 ms (+26.77%) and reduced throughput from 46.54k
to 36.70k tokens/s (-21.14%). Three alternating runs were used for each mode,
with 10 warmup and 50 measured steps per run.

The outputs and gradients remain numerically consistent at BF16 tolerances, so
this report is about performance rather than correctness.

## Reproduce

```python
import json
import statistics

import torch
import torch.nn.functional as F
from liger_kernel.ops.swiglu import LigerSiLUMulFunction


ROWS = 3072
COLS = 12288
WARMUP = 10
STEPS = 50


def benchmark(function):
    compiled = torch.compile(function)
    torch.manual_seed(42)
    a = torch.randn(
        (ROWS, COLS), device="cuda", dtype=torch.bfloat16, requires_grad=True
    )
    b = torch.randn(
        (ROWS, COLS), device="cuda", dtype=torch.bfloat16, requires_grad=True
    )
    torch.manual_seed(43)
    gradient = torch.randn_like(a)

    def step():
        a.grad = None
        b.grad = None
        compiled(a, b).backward(gradient)

    for _ in range(WARMUP):
        step()
    torch.cuda.synchronize()

    samples = []
    for _ in range(STEPS):
        start = torch.cuda.Event(enable_timing=True)
        end = torch.cuda.Event(enable_timing=True)
        start.record()
        step()
[...]

PR #1350, opened 2026-08-07
fix(triton): pin fp64 scalar kernel params to fp32 in swiglu and rms_norm

vaibhavjindal:
## Summary

Hand-written Triton kernels get their non-constexpr Python-float params specialized differently by eager Triton and by Inductor. Eager specializes them to `fp32`; when the same kernel is launched from inside `torch.compile`, Inductor specializes them to `fp64`:

```
# TORCH_LOGS=output_code, _swiglu_forward_kernel as generated by Inductor
triton_meta={'signature': {'a_ptr': '*bf16', ..., 'gate_multiplier': 'fp64', ...}}
```

Since `x.to(tl.float32) * gate_multiplier` mixes an fp32 tensor with an fp64 scalar, Triton promotes the **whole expression to float64** — including the transcendentals (`tl.sigmoid` in swiglu, `rsqrt` in rms_norm). These kernels are memory bound, so the promotion roughly halves their throughput.

Isolated repro at 8192x3072 bf16 on H100:

| scalar typing | time | bandwidth |
|---|---|---|
| fp32 (eager path) | 56.1 us | 2693 GB/s |
| fp64 (Inductor path) | 108.5 us | 1392 GB/s |
| fp64 + `.to(tl.float32)` | **56.0 us** | **2694 GB/s** |

The fix is an explicit `.to(tl.float32)` on the scalar inside the kernel — a no-op on the eager path (already fp32) and a full recovery on the compiled path. `rms_norm` already did exactly this for `_CASTING_MODE_NONE`; this extends it to the LLAMA/GEMMA casting modes and to the affine `offset` used by both backward kernels.

## Details

This was found while benchmarking Liger against `torch.compile` on Qwen3-0.6B. Worth noting what it was **not**: there are no graph breaks in either config (Liger actually lowers op count 1266 -> 844), and standalone the Liger SwiGLU kernel already matched Inductor's exactly (both 56us, ~80% of roofline). The regression was `torch.compile` silently degrading Liger's own kernels via scalar typing.

Per-kernel GPU time, Qwen3-0.6B trunk forward @ 8x1024, H100 bf16:

| kernel | before | after | compile-only equivalent |
|---|---|---|---|
| `_swiglu_forward_kernel` | 3.017 ms | **1.361 ms** | 1.316 ms |
| `_rms_norm_forward_kernel` | 0.848 ms | **0.655 ms** | — |
| total GPU | 24.396 ms | **22.958 ms** | 23.395 ms |

End-to-end speedup vs eager HF baseline (Liger patched with `rope + rms_norm + swiglu` only; CE / fused-linear-CE off so the LM head is identical across configs):

| shape | mode | `torch.compile` | `liger+compile` before | `liger+compile` after |
|---|---|---|---|---|
| 8x1024 | trunk fwd | 1.870x | 1.856x | **2.075x** |
| 8x2048 | trunk fwd | 1.557x | 1.509x | **1.674x** |
| 4x2048 | trunk fwd | 1.578x | 1.534x | **1.727x** |
| 8x1024 | fwd+bwd | 1.978x | 1.845x | **2.008x** |
| 8x2048 | fwd+bwd | 1.651x | 1.554x | **1.691x** |

So stacking Liger with `torch.compile` was previously a regression relative to `torch.compile` alone, and is now a win. Eager-only Liger is unchanged (27.765 ms vs 27.755 ms before), confirming the cast is free on that path.

Numerical drift versus the eager baseline also **improves**, since the compiled path now does the same fp32 arithmetic eager does:

```
liger        : rel=2.01e-02  argmax_agree=100.0%
compile      : rel=1.92e-02  argmax_agree=100.0%
liger+compile: rel=4.77e-02  ->  2.51e-02   argmax_agree=100.0%
```

Two things I deliberately left out of this PR:

1. **The same pattern likely affects other ops.** An AST scan turned up ~80 float-scalar kernel params across ~20 kernels (`layer_norm`, `group_norm`, `poly_norm`, `dyt`, `cross_entropy` softcap, `grpo_loss` temperature, `tvd` scale, ...). I only fixed the two ops I could measure end-to-end. Happy to do a sweep in a follow-up.
2. **Unrelated:** `torch.compile` with Liger RoPE requires `dynamic=False` — with dynamic shapes Dynamo hits an internal assertion in `higher_order_ops.py` (`assert subgraph_vt.is_tensor() or isinstance(subgraph_vt, SymNodeVariable)`) while tracing `LigerRopeFunction`, on the *second* distinct input shape. That's a separate issue, not addressed here.

## Testing Done

Correctness on the touched ops, plus convergence:

- `pytest test/transformers/test_swiglu.py test/transformers/test_rms_norm.py` -> **113 passed**, 4 skipped, 12 xfailed
- `pytest test/transformers/test_fused_add_rms_norm.py test/transformers/test_monkey_patch.py` -> **100 passed**, 2 xfailed
- `pytest test/convergence/bf16/test_mini_models.py -k "qwen3 or llama"` -> **10 passed**
- `make checkstyle` -> passed

- Hardware Type: **H100 80GB HBM3**

> Note: I ran the targeted suites listed above rather than the full `make test` /
> `make test-convergence` sweeps, so I've left those two boxes unchecked rather than
> claim more coverage than I actually have. Relying on CI for the full run.

### DIFF
--- a/src/liger_kernel/ops/rms_norm.py
+++ b/src/liger_kernel/ops/rms_norm.py
@@ -106,6 +106,14 @@ def _rms_norm_forward_kernel(
     if casting_mode == _CASTING_MODE_NONE:
         eps = eps.to(X_row_dtype)
         offset = offset.to(X_row_dtype)
+    else:
+        # Scalar kernel params are specialized to fp32 by eager Triton but to
+        # fp64 by Inductor when this kernel is launched from inside
+        # torch.compile. An fp64 scalar silently promotes mean_square, rsqrt
+        # and the weight multiply to float64, roughly halving throughput.
+        # Pinning to fp32 is a no-op in eager and keeps both paths identical.
+        eps = eps.to(tl.float32)
+        offset = offset.to(tl.float32)
 
     mean_square = tl.sum(X_row * X_row, axis=0) / n_cols
     rstd = rsqrt(mean_square + eps)
@@ -171,7 +179,8 @@ def _rms_norm_backward_kernel(
 
     if elementwise_affine:
         W_row = tl.load(W_ptr + col_offsets, mask=mask, other=0.0)
-        W_row = W_row + offset
+        # Pin the fp64 scalar Inductor passes to fp32 (see _rms_norm_forward_kernel).
+        W_row = W_row + offset.to(tl.float32)
 
     for row_idx in range(row_start, row_end):
         dy_base = dY_ptr + row_idx * dY_row_stride
@@ -280,6 +289,10 @@ def _block_rms_norm_forward_kernel(
     if casting_mode == _CASTING_MODE_NONE:
         eps = eps.to(X_row_dtype)
         offset = offset.to(X_row_dtype)
+    else:
+        # See _rms_norm_forward_kernel: pin fp64 scalars from Inductor to fp32.
+        eps = eps.to(tl.float32)
+        offset = offset.to(tl.float32)
 
     mean_square = tl.sum(X_row * X_row, axis=1) / n_cols
     rstd = rsqrt(mean_square + eps)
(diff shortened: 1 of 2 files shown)


kolehma8 (approved):

vaibhavjindal:
#1330

Issue #1347, opened 2026-08-06
[RFC] Full-pipeline SwiGLU MLP with forward/backward co-optimization


Pearblossom-M:
## Summary
I have implemented **Flash SwiGLU MLP**, a Triton-based SwiGLU MLP that co-optimizes the complete forward and backward dataflow rather than only the element-wise SiLU-and-gating stage.

The current `LigerSwiGLUMLP` keeps `gate_proj`, `up_proj`, and `down_proj` as `nn.Linear` modules and applies `LigerSiLUMulFunction` to the outputs of the first two projections. This implementation explores a broader fusion boundary:

- fuse the gate and up projections with SiLU and element-wise gating;
- store `A`, `G`, and `U` in one contiguous buffer;
- avoid materializing the full `dA` tensor during backward;
- overwrite expired `G/U` storage with `dG/dU`;
- aggregate the `dWg` and `dWu` computations into one GEMM;
- compute `dI` in one fused kernel with a single accumulator;
- use a dedicated no-grad path that does not save `G/U`.

Across the complete benchmark set, most measured end-to-end training speedups are in the **5%–15%** range compared with `LigerSwiGLUMLP`. The no-grad inference path also uses substantially less intermediate memory, although training-memory behavior is shape-dependent.

- Repository: https://github.com/Pearblossom-M/flash-swiglu-mlp
- Technical report: https://github.com/Pearblossom-M/flash-swiglu-mlp/blob/main/flash-swiglu-mlp-technical-report.en.pdf

## Question for maintainers
Would an implementation with this fusion scope be appropriate for Liger Kernel?
I would mainly appreciate feedback on whether this broader fusion scope is interesting and fits the project's goals.

## Validation environment
- GPU: RTX 5060 Ti 16 GB (SM120)
- Driver: 580.159.03
- CUDA: 13.1
- PyTorch: 2.13.0
- Triton: 3.7.1
- Liger Kernel: 0.8.1
- Dtypes: BF16 and FP16; the tables below show BF16
- Training shapes:
  - `(D, H) ∈ {(1024, 2816), (2048, 5632), (4096, 11008)}`
  - `B = 8`
  - `S ∈ {256, 512, 1024, 2048, 4096, 8192}`
- Inference shapes:
  - `B = 1`
  - effective request count `S ∈ {1, 2, 4, 8, 16, 32, 64, 128, 256}`

[...]

PR #1357, opened 2026-08-09
feat: add LigerMLP module

Pearblossom-M:
## Summary
Add LigerMLP, a Triton-based fused SwiGLU MLP that co-optimizes the complete forward and backward dataflow rather than only the element-wise SiLU-and-gating stage.

Linked issue: #1347

## Details
When running `make test`, an `xfail` occurs. This is not due to an error in the operator implementation, but because in `test_misaligned_intermediate_size_not_supported`, the `intermediate_size` is set to 431, which does not satisfy Triton's `TensorDescriptor` requirement of 16‑byte alignment. 
Additionally, since this implementation uses `tl.dot`, when the input data type is `fp32`, `tl.dot` performs computations using `tf32` internally, which differs from PyTorch's behavior of using pure `fp32` for computation. As a result, the numerical error between this implementation and the reference implementation is relatively larger. Therefore, when the data type is `fp32`, the tolerance range is set to a larger value.


## Testing Done

- Hardware Type: RTX 5060ti 16G（SM120）

```
python -m pytest test/transformers/test_mlp.py
================================================================================================== test session starts ===================================================================================================
platform linux -- Python 3.12.3, pytest-9.0.2, pluggy-1.6.0
rootdir: /home/ma-jh/Liger-Kernel
configfile: pyproject.toml
plugins: rerunfailures-16.4, asyncio-1.4.0, xdist-3.8.0, anyio-4.12.0, cov-7.1.0
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 24 items                                                                                                                                                                                                       

test/transformers/test_mlp.py::test_correctness_llamamlp[dtype0-200.0-0.02-2-256-256-512] PASSED                                                                                                                   [  4%]
test/transformers/test_mlp.py::test_correctness_llamamlp[dtype0-200.0-0.02-6-42-128-432] PASSED                                                                                                                    [  8%]
test/transformers/test_mlp.py::test_correctness_llamamlp[dtype0-200.0-0.02-3-37-96-264] PASSED                                                                                                                     [ 12%]
test/transformers/test_mlp.py::test_correctness_llamamlp[dtype1-10000.0-0.01-2-256-256-512] PASSED                                                                                                                 [ 16%]
test/transformers/test_mlp.py::test_correctness_llamamlp[dtype1-10000.0-0.01-6-42-128-432] PASSED                                                                                                                  [ 20%]
test/transformers/test_mlp.py::test_correctness_llamamlp[dtype1-10000.0-0.01-3-37-96-264] PASSED                                                                                                                   [ 25%]
test/transformers/test_mlp.py::test_correctness_llamamlp[dtype2-100.0-0.01-2-256-256-512] PASSED                                                                                                                   [ 29%]
test/transformers/test_mlp.py::test_correctness_llamamlp[dtype2-100.0-0.01-6-42-128-432] PASSED                                                                                                                    [ 33%]
test/transformers/test_mlp.py::test_correctness_llamamlp[dtype2-100.0-0.01-3-37-96-264] PASSED                                                                                                                     [ 37%]
test/transformers/test_mlp.py::test_correctness_inference_mode[dtype0-200.0-0.02-2-256-256-512] PASSED                                                                                                             [ 41%]
test/transformers/test_mlp.py::test_correctness_inference_mode[dtype0-200.0-0.02-6-42-128-432] PASSED                                                                                                              [ 45%]
test/transformers/test_mlp.py::test_correctness_inference_mode[dtype0-200.0-0.02-3-37-96-264] PASSED                                                                                                               [ 50%]
test/transformers/test_mlp.py::test_correctness_inference_mode[dtype1-10000.0-0.01-2-256-256-512] PASSED                                                                                                           [ 54%]
test/transformers/test_mlp.py::test_correctness_inference_mode[dtype1-10000.0-0.01-6-42-128-432] PASSED                                                                                                            [ 58%]
test/transformers/test_mlp.py::test_correctness_inference_mode[dtype1-10000.0-0.01-3-37-96-264] PASSED                                                                                                             [ 62%]
test/transformers/test_mlp.py::test_correctness_inference_mode[dtype2-100.0-0.01-2-256-256-512] PASSED                                                                                                             [ 66%]
test/transformers/test_mlp.py::test_correctness_inference_mode[dtype2-100.0-0.01-6-42-128-432] PASSED                                                                                                              [ 70%]
test/transformers/test_mlp.py::test_correctness_inference_mode[dtype2-100.0-0.01-3-37-96-264] PASSED                                                                                                               [ 75%]
test/transformers/test_mlp.py::test_invalid_hidden_act_raises[gelu] PASSED                                                                                                                                         [ 79%]
test/transformers/test_mlp.py::test_invalid_hidden_act_raises[relu] PASSED                                                                                                                                         [ 83%]
test/transformers/test_mlp.py::test_invalid_hidden_act_raises[tanh] PASSED                                                                                                                                         [ 87%]
test/transformers/test_mlp.py::test_supported_hidden_act[silu] PASSED                                                                                                                                              [ 91%]
test/transformers/test_mlp.py::test_supported_hidden_act[swish] PASSED                                                                                                                                             [ 95%]
test/transformers/test_mlp.py::test_misaligned_intermediate_size_not_supported XFAIL (TMA descriptors require 3 * intermediate_size * elem_bytes to be 16-byte aligned; intermediate_size=431 with bf16 (strid...) [100%]
```

```
make checkstyle
ruff check --output-format=concise .; ruff_check_status=$?; \
ruff format --check --diff .; ruff_format_status=$?; \
ruff check . --fix; \
ruff format .; \
if [ $ruff_check_status -ne 0 ] || [ $ruff_format_status -ne 0 ]; then \
        exit 1; \
fi
All checks passed!
354 files already formatted
All checks passed!
354 files left unchanged
```


test/transformers/test_mlp.py
```diff
@@ -0,0 +1,200 @@
...
+        # so hidden/intermediate sizes are kept multiples of 8)
+        (6, 42, 128, 432),
+        (3, 37, 96, 264),
+    ],
+)
+@pytest.mark.parametrize(
+    "dtype, atol, rtol",
+    [
+        # fp32: Triton (fp32 accum) vs cuBLAS (TF32) diverge through 3 chained
+        # GEMMs; activations reach ~1e4 magnitude, so atol must scale accordingly
```
Tcc0403: instead of assert close, try [cosine similarity approach](https://github.com/deepseek-ai/DeepGEMM/blob/559d79fb6994a58b8a15b4b93bf13ccc16edf247/tests/test_bf16.py#L40) that DeepGEMM uses and see if we can pass with 1e-5 tolerance

src/liger_kernel/ops/mlp.py
```diff
@@ -0,0 +1,632 @@
...
+    return input, gate_weight, up_weight, down_weight
+
+
+def swiglu_forward(
+    input: torch.Tensor,
+    gate_weight: torch.Tensor,
+    up_weight: torch.Tensor,
+    down_weight: torch.Tensor,
+) -> torch.Tensor:
+    """Training-mode forward: computes O, and returns AGU."""
```
Tcc0403: I prefer wrapping `gemm_swiglu` op only and call plain F.linear in autograd function directly for composibility, easiser to integrate into other framework.

src/liger_kernel/ops/mlp.py
```diff
@@ -0,0 +1,632 @@
...
+
+def swiglu_backward(
+    dO: torch.Tensor,
+    input: torch.Tensor,
+    gate_weight: torch.Tensor,
+    up_weight: torch.Tensor,
+    down_weight: torch.Tensor,
+    AGU: torch.Tensor,
+    fwd_best_config,
+) -> torch.Tensor:
```
Tcc0403: ditto, split this function into gemm_dswiglu, dgrad and wgrad, compose them in autograd function bwd

src/liger_kernel/ops/mlp.py
```diff
@@ -0,0 +1,632 @@
...
+
+    dtypes = {t.dtype for t in [input, gate_weight, up_weight, down_weight]}
+    if len(dtypes) != 1:
+        raise TypeError(f"All tensors must share the same dtype, got {dtypes}.")
+
+    input = _ensure_tma_compatible(input)
+    gate_weight = _ensure_tma_compatible(gate_weight)
+    up_weight = _ensure_tma_compatible(up_weight)
+    down_weight = _ensure_tma_compatible(down_weight)
+
```
Tcc0403, line 450: Do we support native hf/transformers weight layout? Ideally we want to avoid copying weight ops, such as `.reshape`, `.contiguous` (unless tensor is already contiguous), which can save redundant memory roundtrips. You might need to check with torch porfiler

It can be a follow-up PR, but good to know whether the current impl is copy-free or not

src/liger_kernel/ops/mlp.py
```diff
@@ -0,0 +1,632 @@
...
+        )
+        for BM in [64, 128]
+        for BN in [128]
+        for BK in [64, 128]
+        for GS in [1, 8]  # GROUP_SIZE_M = 1 equal to no L2 swizzle.
+        for w in [4, 8]
+        for s in [2, 3]
+    ],
+    key=["dim", "hidden_dim", "bucket_M"],
+)
```
Tcc0403, line 74: how long does it take for autotuning with current search space

test/transformers/test_mlp.py
Pearblossom-M: I tried the cosine similarity approach used by DeepGEMM, and all tests passed (tolerance = 1e‑5).

src/liger_kernel/ops/mlp.py
Pearblossom-M: I've addressed this comment and attached the code below – is this what you meant?
```
@staticmethod
    def forward(ctx, input, gate_weight, up_weight, down_weight):
        assert input.is_cuda and gate_weight.is_cuda and up_weight.is_cuda and down_weight.is_cuda
        input, gate_weight, up_weight, down_weight = _check_inputs(input, gate_weight, up_weight, down_weight)
        # Note:
        # PyTorch automatically disables global gradient computation during
        # the forward pass, so torch.is_grad_enabled() cannot be used.
        if any(ctx.needs_input_grad):
            AGU, fwd_best_config = gemm_swiglu(input, gate_weight, up_weight)
            ctx.save_for_backward(input, gate_weight, up_weight, down_weight, AGU)
            ctx.fwd_best_config = fwd_best_config
            hidden_dim = gate_weight.shape[0]
            return F.linear(AGU[..., :hidden_dim], down_weight)
        else:
            # Inference mode: no GU saving.
            A = gemm_swiglu_inference(input, gate_weight, up_weight)
            return F.linear(A, down_weight)
```

src/liger_kernel/ops/mlp.py
Pearblossom-M: Modified. Still 4 stages, but the kernels that compute dGU and dI are now separately wrapped. Here is the code:
```
@staticmethod
    def backward(ctx, dO):
        dO = _ensure_tma_compatible(dO)
        input, gate_weight, up_weight, down_weight, AGU = ctx.saved_tensors
        hidden_dim, dim = gate_weight.shape

        # Compute dWd.
        A = AGU[..., :hidden_dim]
        dWd = dO.reshape(-1, dim).T @ A.reshape(-1, hidden_dim)  # [dim, hidden_dim]

        # Compute dU and dG by triton kernel.
        # The original G and U will be overwritten by dG and dU.
        swiglu_backward_dGU(dO, down_weight, AGU, ctx.fwd_best_config)

        # Compute dWug by a single large GEMM.
        dGU = AGU[..., hidden_dim:]
        dWug = dGU.reshape(-1, 2 * hidden_dim).T @ input.reshape(-1, dim)  # [2*hidden_dim, dim]
        dWg, dWu = dWug[:hidden_dim], dWug[hidden_dim:]  # [hidden_dim, dim]
        
        # Compute dI by triton kernel.
        dI = swiglu_backward_dI(gate_weight, up_weight, AGU)

        return dI, dWg, dWu, dWd
```

src/liger_kernel/ops/mlp.py
Pearblossom-M: Yes — LigerMLP consumes the native HF nn.Linear weight layout directly, with zero adaptation required. As shown in LigerMLP.forward, we pass self.gate_proj.weight / self.up_proj.weight / self.down_proj.weight straight into LigerMLPFunction.apply — there's no .t(), .reshape(), or .contiguous() call anywhere at the module level.

Verified via data_ptr() identity checks — _ensure_tma_compatible does not trigger any additional .contiguous() copy across the layouts that matter in practice:
* gate_proj.weight: full weight, standard HF row-major layout.
* up_proj.weight: sharded along dim 0 (output-dim shard, e.g. column-parallel style).
* down_proj.weight: sharded along dim 1 (input-dim shard, e.g. row-parallel style).

All three preserve the same data_ptr() before/after processing, confirming zero-copy for both the native nn.Linear layout and the non-contiguous views produced by tensor-parallel sharding. Test:

```
def check_data_ptr_zero_copy(model):
    """Ground-truth check: does _ensure_tma_compatible return the same
    storage as the input, or a freshly allocated copy?"""
    w = model.gate_proj.weight
    assert w.data_ptr() == _ensure_tma_compatible(w).data_ptr(), "COPIED (gate_proj)"

    w = model.up_proj.weight[: model.up_proj.weight.size(0) // 2, :]
    assert w.data_ptr() == _ensure_tma_compatible(w).data_ptr(), "COPIED (up_proj)"

    w = model.down_proj.weight[:, : model.down_proj.weight.size(1) // 2]
    assert w.data_ptr() == _ensure_tma_compatible(w).data_ptr(), "COPIED (down_proj)" 
```

src/liger_kernel/ops/mlp.py
Pearblossom-M: Search space size: BM(2) × BN(1) × BK(2) × GS(2) × w(2) × s(2) = 32 configs total.

Measured on RTX 5060 Ti (SM120), for dim=1024, hidden_dim=4096, B=4, S=1024: autotuning took 1.51s for this shape.

This is a one-time cost per unique (dim, hidden_dim, bucket_M) key — since we key on bucket_M rather than the raw sequence length, the autotune cache is reused across all batches/steps that fall into the same bucket, so this isn't paid every iteration. In a typical training run with a fixed dim/hidden_dim and a handful of bucket_M buckets, this amounts to a few seconds of one-time warmup total, not a recurring cost.

Pearblossom-M:
@Tcc0403  I have completed all the modifications, specifically:

* In `test_mlp.py`, the cosine similarity judgment method is used, and all tests pass with tolerance = 1e‑5.

* The wrapping approach for the Triton kernel has been modified, pure PyTorch operations are now placed inside the `autograd.Function`.

* The internal logic of `_ensure_tma_compatible` has been revised. Previously, the implementation could not catch tensors that did not meet the requirements of `TensorDescriptor` in a timely manner—it would only raise an error when the `TensorDescriptor` was being created. Now, it raises an error promptly.

test/transformers/test_mlp.py
```diff
@@ -0,0 +1,199 @@
...
+        return 0.0
+    sim = 2 * (x * y).sum() / denominator
+    return 1 - sim
+
+
+def _assert_close(x: torch.Tensor, y: torch.Tensor, dtype: torch.dtype, atol, rtol):
+    if dtype == torch.float32:
+        assert calc_diff(x, y) < FP32_DIFF_THRESHOLD
+    else:
+        assert torch.allclose(x, y, atol=atol, rtol=rtol)
```
Tcc0403: let's use calc_diff for all dtypes. 1e4 and 1e2 atol for assert_close feel way too large to me.

src/liger_kernel/ops/mlp.py
```diff
@@ -0,0 +1,654 @@
...
+    A_block = acc_gate * tl.sigmoid(acc_gate) * acc_up  # [BLOCK_M, BLOCK_N]
+
+    # Write back
+    dtype = desc_input.dtype
+    A_block = tl.reshape(A_block, [1, BLOCK_M, BLOCK_N])
+    desc_A.store([b, M_start, N_start], A_block.to(dtype))
+    G_block = tl.reshape(acc_gate, [1, BLOCK_M, BLOCK_N])
+    desc_G.store([b, M_start, N_start], G_block.to(dtype))
+    U_block = tl.reshape(acc_up, [1, BLOCK_M, BLOCK_N])
+    desc_U.store([b, M_start, N_start], U_block.to(dtype))
```
Tcc0403, line 132: nit: since the only difference between training and inference is whether storing pre-activation tensors, add a flag to `_swiglu_kernel_forward` so that forward function callsite would be cleaner. `store_preact` flag name comes from https://github.com/Dao-AILab/quack/blob/60d88082272a256fa9b3b2ab631c82cfa78337c6/quack/gemm_interface.py#L2138 
```suggestion
	if store_preact:
	    G_block = tl.reshape(acc_gate, [1, BLOCK_M, BLOCK_N])
	    desc_G.store([b, M_start, N_start], G_block.to(dtype))
	    U_block = tl.reshape(acc_up, [1, BLOCK_M, BLOCK_N])
	    desc_U.store([b, M_start, N_start], U_block.to(dtype))
```

Tcc0403 (commented):
Overall lgtm,  you can also check out how liger integrates with hf/transformers via [monkey_patch.py](https://github.com/linkedin/Liger-Kernel/blob/a5d795efd2c1436549e70118ef519134e9c27833/src/liger_kernel/transformers/monkey_patch.py) as a follow-up PR.  cc @vaibhavjindal I just noticed there's a new liger cute working project on MoE, I wonder if there's any upcoming change on integration method along with new kernels.

test/transformers/test_mlp.py
Pearblossom-M: All tests have been updated to cosine similarity approach, and all passed.

src/liger_kernel/ops/mlp.py
Pearblossom-M: Thank you for your suggestion. Since training and inference have different workloads, the autotune search spaces for `_swiglu_kernel_forward` and `_swiglu_kernel_forward_inference` differ. I suggest keeping this distinction to enable finer‑grained tuning for each path in the future. However, to keep the code cleaner, I merged the wrappers of the two kernels by adding a store_preact flag. Now there is only a single function, `gemm_swiglu`.

Pearblossom-M:
@Tcc0403 Thanks for the review! I'll consider this PR (add `LigerMLP` module) feature-complete as is, and follow up on the `monkey_patch.py` integration in a separate PR.

As a quick sanity check for that follow-up, I made a small-scale attempt on the Llama path — swapped `LigerSwiGLUMLP` for `LigerMLP` inside `apply_liger_kernel_to_llama`, and it's passing `test/transformers/test_monkey_patch.py::test_apply_liger_kernel_to_instance_for_llama` plus the bf16 convergence tests for `mini_llama3` (both `test_mini_models.py` and `test_mini_models_with_logits.py`).

Before I start that follow-up PR, a couple of scoping questions:

1. Rollout scope: should the follow-up PR start as a Llama-only PoC (validate the approach, then extend to other SwiGLU-family models like Mistral, Qwen2, etc. in later PRs), or would you rather I cover all SwiGLU-family models in one PR from the start?
2. API: should `LigerMLP` fully replace `LigerSwiGLUMLP`, or keep `LigerSwiGLUMLP` as a backward-compatible alias?

Happy to go whichever direction makes sense.

Tcc0403:
> Rollout scope: should the follow-up PR start as a Llama-only PoC (validate the approach, then extend to other SwiGLU-family models like Mistral, Qwen2, etc. in later PRs), or would you rather I cover all SwiGLU-family models in one PR from the start?

Let's cover all at once

>API: should LigerMLP fully replace LigerSwiGLUMLP, or keep LigerSwiGLUMLP as a backward-compatible alias?

Keep LigerSwiGLUMLP for bwd-compatibility

src/liger_kernel/ops/mlp.py
Tcc0403: makes sense, thanks for the explanation

Pearblossom-M:
@Tcc0403 I've just pushed new commits addressing your feedback. Could you please take another look when you have a moment? Thanks!
## What this change includes

* Added **Falcon H1** support, with a new `LigerFalconH1MLP`.

* In `src/liger_kernel/transformers/monkey_patch.py`: a **global flag** `USE_FLASH_SWIGLU` controls whether every `LigerSwiGLUMLP` is swapped for `LigerMLP`, and `LigerFalconH1SwiGLUMLP` is swapped for `LigerFalconH1MLP`. The one exception is `apply_liger_kernel_to_llama4`: **Llama4**'s `Llama4MoE.forward` reshapes the hidden states into a 2D `[T, hidden]` tensor before feeding them to the MLP, and the current fused kernels do not support 2D input, so Llama4 keeps the original `LigerSwiGLUMLP`. 

  I first tried a **minimal workaround** — `unsqueeze(0)` to turn the 2D input into 3D, then `squeeze(0)` on the output — but this crashes at the model level: Llama4 performs an in-place addition `out.add_(...)`, and `squeeze(0)` returns a view. PyTorch does not allow in-place modification of a view produced by a custom autograd Function. `.clone()` would fix the view issue, but it adds an extra copy of the full hidden-state tensor, so that is not acceptable either. The plan is therefore to keep Llama4 on the original implementation for now, and later modify the kernel to natively support 2D input.

* Unit tests pass: `test/transformers/test_mlp.py` and `test/transformers/test_monkey_patch.py`.

## Two points about the `make convergence` results

* In bf16, everything passes except `mini_pixtral` (test_mini_models_multimodal), `mini_gpt_oss` (test_mini_models) and `mini_qwen3_moe` (test_mini_models_with_logits):

  * The `mini_gpt_oss` and `mini_qwen3_moe` failures are **not caused by this change** — they also fail on a completely unmodified branch.

  * The `mini_pixtral` failure comes from a **rounding-path difference**, not a formula error. The reference implementation rounds every intermediate result (gate/up GEMM outputs, SiLU output, etc.) to bf16, while the fused kernel keeps all intermediates in fp32 and rounds only once when writing the final result. Note that this test's "loss" is the sum of the vision encoder's `last_hidden_state` (not a cross-entropy loss), so values around 1e7 are expected by design — the large magnitude is not a sign of divergence. Measured over 32 steps, the largest relative deviation is 3.18e-2; raising `loss_rtol` from 1e-2 to 5e-2 makes the test pass with margin (4e-2 is also sufficient, the measured threshold is exactly 3.18e-2). 

    Maintainer input is needed on whether this **relaxation** is acceptable, or whether a **different testing approach** would be preferred.

* In fp32 there are many failures. The **root cause** is that `tl.dot` uses **TF32** for fp32 matrix math by default, while the reference implementation calls PyTorch's `nn.Linear`, which computes in true FP32. This mismatch causes the failures. Triton does provide `input_precision="ieee"` to force true FP32, but that is **not a simple fix**: switching to IEEE precision triggers a **shared memory overflow** (the error reports 114712 bytes needed vs. a hardware limit of 101376 bytes). The reason is that TF32 compiles to Tensor Core MMA instructions with a compact data layout, while IEEE cannot use Tensor Cores and falls back to a software FMA path that needs more shared memory. Even the smallest tile configuration in the current autotune search space overflows, so there is no fallback option. **Possible fixes** include:

  * Relax the fp32 tolerances
  * Force IEEE precision and re-tune the autotune search space (with a performance cost)
  * Use a different way of testing fp32 correctness

  Which approach to take is up to the maintainers.

The raw **failure logs** are as follows:

### pytest test/convergence/fp32/test_mini_models.py

```
FAILED test/convergence/fp32/test_mini_models.py::test_mini_model[mini_llama3-32-0.0001-dtype1-1e-08-2e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 9
FAILED test/convergence/fp32/test_mini_models.py::test_mini_model[mini_llava-32-0.0001-dtype2-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 33
FAILED test/convergence/fp32/test_mini_models.py::test_mini_model[mini_qwen2-32-0.0001-dtype5-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 33
FAILED test/convergence/fp32/test_mini_models.py::test_mini_model[mini_qwen2_vl-32-0.0001-dtype9-1e-05-0.1-0.005-1e-05-0.005-1e-05] - AssertionError: [Top k logprobs]Number of mismatched elements: 17046
FAILED test/convergence/fp32/test_mini_models.py::test_mini_model[mini_qwen2_5_vl-32-0.0001-dtype10-1e-05-0.1-0.005-1e-05-0.005-1e-05] - AssertionError: [Top k logprobs]Number of mismatched elements: 1894
FAILED test/convergence/fp32/test_mini_models.py::test_mini_model[mini_qwen3-32-0.0001-dtype6-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 33
FAILED test/convergence/fp32/test_mini_models.py::test_mini_model[mini_mistral-32-0.0001-dtype19-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 17
FAILED test/convergence/fp32/test_mini_models.py::test_mini_model[mini_ministral-32-0.0001-dtype20-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 17
FAILED test/convergence/fp32/test_mini_models.py::test_mini_model[mini_granite3-32-0.0001-dtype24-1e-08-0.0001-0.04-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 13
FAILED test/convergence/fp32/test_mini_models.py::test_mini_model[mini_exaone4-32-1e-05-dtype34-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 32

10 failed
```

### pytest test/convergence/fp32/test_mini_models_with_logits.py

```
FAILED test/convergence/fp32/test_mini_models_with_logits.py::test_mini_model[mini_llama3-32-0.0001-dtype1-1e-08-2e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 10
FAILED test/convergence/fp32/test_mini_models_with_logits.py::test_mini_model[mini_llava-32-0.0001-dtype2-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 32
FAILED test/convergence/fp32/test_mini_models_with_logits.py::test_mini_model[mini_mllama-32-0.0001-dtype3-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Top K Logprobs]Number of mismatched elements: 2
FAILED test/convergence/fp32/test_mini_models_with_logits.py::test_mini_model[mini_qwen2-32-0.0001-dtype5-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 32
FAILED test/convergence/fp32/test_mini_models_with_logits.py::test_mini_model[mini_qwen2_vl-32-0.0001-dtype8-1e-08-2e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 26
FAILED test/convergence/fp32/test_mini_models_with_logits.py::test_mini_model[mini_qwen2_5_vl-32-0.0001-dtype9-1e-08-2e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 24
FAILED test/convergence/fp32/test_mini_models_with_logits.py::test_mini_model[mini_qwen3-32-0.0001-dtype6-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 32
FAILED test/convergence/fp32/test_mini_models_with_logits.py::test_mini_model[mini_mistral-32-0.0001-dtype18-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 18
FAILED test/convergence/fp32/test_mini_models_with_logits.py::test_mini_model[mini_ministral-32-0.0001-dtype19-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 18
FAILED test/convergence/fp32/test_mini_models_with_logits.py::test_mini_model[mini_granite3-32-0.0001-dtype23-1e-08-0.0001-0.005-1e-05-0.005-1e-05] - AssertionError: [Top K Logprobs]Number of mismatched elements: 59
FAILED test/convergence/fp32/test_mini_models_with_logits.py::test_mini_model[mini_exaone4-32-1e-05-dtype32-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 30

11 failed
```

### pytest test/convergence/fp32/test_mini_models_multimodal.py

```
FAILED test/convergence/fp32/test_mini_models_multimodal.py::test_mini_model_multimodal[mini_qwen2_vl-32-0.0001-dtype0-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 30
FAILED test/convergence/fp32/test_mini_models_multimodal.py::test_mini_model_multimodal[mini_qwen2_5_vl-32-0.0001-dtype5-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 33
FAILED test/convergence/fp32/test_mini_models_multimodal.py::test_mini_model_multimodal[mini_mllama-32-0.0001-dtype8-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 28
FAILED test/convergence/fp32/test_mini_models_multimodal.py::test_mini_model_vision[mini_pixtral-32-0.0001-dtype0-1e-08-1e-05-0.005-1e-05-0.005-1e-05] - AssertionError: [Loss]Number of mismatched elements: 32

4 failed
```

### pytest test/convergence/bf16/test_mini_models.py

```
FAILED test/convergence/bf16/test_mini_models.py::test_mini_model[mini_gpt_oss-32-1e-05-dtype8-0.05-0.05-0.1-0.1-0.01-0.01] - AssertionError: [Loss]Number of mismatched elements: 1
```

### pytest test/convergence/bf16/test_mini_models_with_logits.py

```
FAILED test/convergence/bf16/test_mini_models_with_logits.py::test_mini_model[mini_qwen3_moe-32-1e-05-dtype7-0.01-0.2-0.1-0.01-0.01-0.01] - AssertionError: [Loss]Number of mismatched elements: 5
```

### pytest test/convergence/bf16/test_mini_models_multimodal.py

```
FAILED test/convergence/bf16/test_mini_models_multimodal.py::test_mini_model_vision[mini_pixtral-32-0.0001-dtype0-0.001-0.01-1.0-0.01-0.01-0.01] - AssertionError: [Loss]Number of mismatched elements: 1
```

Pearblossom-M:
@Tcc0403 Since the testing methodology for fp32 convergence tests on the fused kernel still needs further discussion (see #1383), I've reverted the monkey patch changes for now so they don't block this PR from being merged.

This PR is now scoped to just the core kernel implementations:
* LigerMLP
* LigerFalconH1MLP
* LigerMLPFunction

The monkey patch integration will move to a follow-up PR once we've settled on an fp32 testing methodology.

Also, it looks like a workflow run is waiting on approval in the Checks tab — would you mind approving it when you get a chance?

Tcc0403 (approved):
LGTM

src/liger_kernel/transformers/mlp.py
```diff
@@ -0,0 +1,55 @@
...
+        y = down_proj(silu(gate_proj(x) * gate_mult) * up_proj(x)) * down_mult
+    https://github.com/huggingface/transformers/blob/main/src/transformers/models/falcon_h1/modeling_falcon_h1.py
+    """
+
+    def __init__(self, config):
+        super().__init__()
+        self.config = config
+        self.hidden_size = config.hidden_size
+        self.intermediate_size = config.intermediate_size
+        bias = getattr(config, "mlp_bias", False)
```
Tcc0403: I don't think we support mlp_bias currently. We should raise an error or it silently generates incorrect output.

Tcc0403 (changes requested):
tiny error

src/liger_kernel/transformers/mlp.py
Pearblossom-M: @Tcc0403 I fixed the potential silent error with `mlp_bias`. It now raises an explicit error when `mlp_bias=True`. Tested correctness with **transformers 4.52.0** – all tests passed.

Tcc0403 (approved):
LGTM

PR #1359, opened 2026-08-10
[cuTile] Add fused scaled cross entropy

yueyiming2009:
## Summary

Adds a cuTile implementation of `LigerFusedLinearScaledCrossEntropyFunction`, matching the existing per-token scaled NLL and optional differentiable entropy contract.

**Compared with the VERL-derived PPO fallback on the combined NLL-plus-entropy workload, cuTile is 3.1x faster on H100 and 3.7x faster on B200 while reducing peak memory by 55% and 51%, respectively.**

| Hardware | Implementation | Forward | Backward | Full | Speedup vs VERL | Peak full memory | Memory saving vs VERL |
|---|---|---:|---:|---:|---:|---:|---:|
| H100 | **cuTile** | **6.24 ms** | **20.88 ms** | **26.15 ms** | **3.14x** | **2432 MiB** | **55%** |
| H100 | VERL-derived fallback | 22.36 ms | 59.40 ms | 82.03 ms | 1.00x | 5384 MiB | — |
| B200 | **cuTile** | **3.17 ms** | **9.18 ms** | **12.01 ms** | **3.68x** | **2640 MiB** | **51%** |
| B200 | VERL-derived fallback | 12.34 ms | 31.63 ms | 44.16 ms | 1.00x | 5336 MiB | — |

- registers the implementation behind `LIGER_KERNEL_IMPL=cutile`
- supports FP32, FP16, and BF16 matching `input`/`weight` dtypes, scalar positive temperature, ragged dimensions, and `ignore_index`
- bounds forward and backward logits workspaces to 256 MiB portably and automatically uses 512 MiB for large FP16/BF16 Blackwell workloads
- balances token chunks when power-of-two rounding would add a launch or leave a less-than-half-sized final chunk
- supports `LIGER_CUTILE_SCALED_CE_WORKSPACE_MB` for workload-specific workspace tuning
- recomputes logits chunk-by-chunk in backward, transforms them into `dZ` in place, writes `dX` directly, and accumulates `dW` directly with `mm`/`addmm` outputs to avoid full `V x H` temporaries
- tests the direct cuTile class against the independent VERL-derived fallback for combined, NLL-only, and entropy-only gradients
- adds the cuTile provider to the scaled-cross-entropy benchmark and fixes that benchmark to reset both input and weight gradients
- centralizes README documentation for the shared cuTile/CuTe/fallback public API

No open PR currently implements the same cuTile scaled-cross-entropy backend.

## Testing Done

- Hardware Type: NVIDIA H100 80GB HBM3 and NVIDIA B200
  - equivalent command: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest --disable-warnings --ignore=test/convergence --ignore=test/cutedsl test/`
  - result: `4096 passed, 929 skipped, 14 xfailed`
  - result: repository-wide `ruff check .` and `ruff format --check .` passed
  - not applicable: this change adds an opt-in operator implementation and does not patch a model
- cuTile-targeted command: `LIGER_KERNEL_IMPL=cutile PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest -q test/transformers/test_fused_linear_scaled_cross_entropy.py test/transformers/test_cutile_backend.py`
  - result: `29 passed` on both H100 and B200

H100 BF16 benchmark at `M=4096`, `H=4096`, `V=131072`:

| Implementation | Forward | Backward | Full | Peak full memory |
|---|---:|---:|---:|---:|
| cuTile | 6.45 ms | 20.70 ms | 26.94 ms | 2.43 GiB |
| CuTe SM90 | 7.26 ms | 19.81 ms | 27.74 ms | 2.43 GiB |

B200 BF16 benchmark for the same shape, using automatic workspace selection:

| Implementation | Forward | Backward | Full | Peak full memory |
|---|---:|---:|---:|---:|
| cuTile | 2.97 ms | 8.35 ms | 11.36 ms | 2.64 GiB |
| Torch | 5.24 ms | 7.15 ms | 12.85 ms | 7.22 GiB |

AI assistance was used. The submitter reviewed the changed lines and ran the validation above.

### DIFF
--- a/test/transformers/test_fused_linear_scaled_cross_entropy.py
+++ b/test/transformers/test_fused_linear_scaled_cross_entropy.py
@@ -198,7 +198,7 @@ def test_frontend_fallback_supports_entropy_only_backward():
     ("shape", "dtype", "atol", "rtol"),
     [
         ((17, 31, 257), torch.float32, 6e-5, 6e-5),
-        ((521, 64, 5003), torch.bfloat16, 8e-2, 5e-2),
+        ((1025, 64, 131072), torch.bfloat16, 8e-2, 5e-2),
     ],
     ids=["fp32-ragged", "bf16-multichunk"],
 )
--- a/test/transformers/test_fused_linear_scaled_cross_entropy.py
+++ b/test/transformers/test_fused_linear_scaled_cross_entropy.py
@@ -305,6 +305,59 @@ def test_cutile_token_chunk_size_balances_small_tails(token_count, vocab_size, e
 
 @pytest.mark.skipif(not torch.cuda.is_available(), reason="cuTile requires CUDA")
 @pytest.mark.skipif(not _CUTILE_ENABLED, reason="requires LIGER_KERNEL_IMPL=cutile")
+@pytest.mark.parametrize(
+    ("architecture", "token_count", "vocab_size", "element_size", "expected_mb"),
+    [
+        ("hopper", 4096, 131072, 2, 256),
+        ("blackwell", 4096, 131072, 2, 512),
+        ("blackwell_ultra", 4096, 131072, 2, 512),
+        ("blackwell", 2048, 131072, 2, 256),
+        ("blackwell", 4096, 65536, 2, 256),
+        ("blackwell", 4096, 131072, 4, 256),
+    ],
+)
+def test_cutile_selects_workspace_by_architecture(
+    monkeypatch,
+    architecture,
+    token_count,
+    vocab_size,
+    element_size,
+    expected_mb,
+):
+    import liger_kernel.ops.cutile.ops.fused_linear_scaled_cross_entropy as scaled_ce
+
+    monkeypatch.delenv(scaled_ce._WORKSPACE_MB_ENV, raising=False)
+    monkeypatch.setattr(scaled_ce, "infer_device_arch", lambda _: architecture)
+
+    actual = scaled_ce._select_logits_workspace_bytes(0, token_count, vocab_size, element_size)
+    assert actual == expected_mb * scaled_ce._MIB
+
+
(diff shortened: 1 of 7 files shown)


test/transformers/test_fused_linear_scaled_cross_entropy.py
```diff
@@ -170,6 +192,130 @@ def test_frontend_fallback_supports_entropy_only_backward():
...
+        ((17, 31, 257), torch.float32, 6e-5, 6e-5),
+        ((1025, 64, 131072), torch.bfloat16, 8e-2, 5e-2),
+    ],
+    ids=["fp32-ragged", "bf16-multichunk"],
+)
+def test_cutile_matches_torch_forward_and_combined_backward(shape, dtype, atol, rtol):
+    if dtype == torch.bfloat16 and not torch.cuda.is_bf16_supported():
+        pytest.skip("BF16 is not supported")
+
+    from liger_kernel.ops import LigerFusedLinearScaledCrossEntropyFunction
```
kolehma8: I think we should use the Verl implementation that is the fall back implementation of that op for testing purposes instead of the op itself (this might test the kernel against itself if we make the cutile the default option).

yueyiming2009: Addressed in 6774063. The correctness tests now import the cuTile implementation directly from `liger_kernel.ops.cutile.ops.fused_linear_scaled_cross_entropy` and compare it against the independent `_FusedLinearPPOFallbackFunction`, which is adapted from VERL. I also switched the BF16 test to model-like input/weight initialization, matching the existing CuTe VERL-parity tests. The targeted suites pass on both H100 and B200 (29 passed each).

kolehma8 (approved):

PR #1361, opened 2026-08-10
[Liger Cute] Fix fused MoE combine for partial token tiles

supercharleszhu:
## Summary

Fix fused expert-parallel MoE output combine when the token count is not divisible by 32.

The existing TMA combine schedules `num_tokens / 32` tiles, so it never writes the final 1-31 output rows. The updated implementation schedules the final partial tile in the existing TMA combine instead of launching a separate direct-store tail path.

## Details

- Schedule combine tiles with ceiling division.
- Preserve the original load path for every fully valid warp.
- In only the final partial warp, predicate token metadata and weight reads and use `cp.async` zero-fill for invalid token rows.
- Write the full 32-token shared-memory tile with TMA; rows outside the tensor map's `[num_tokens, hidden_dim]` extent are discarded by hardware.
- Add a real eight-GPU NCCL/NVSHMEM forward regression test with `T=45`.
- Add an unaligned forward/backward regression that compares `T=45` gradients against an equivalent zero-padded `T=64` execution.

## Testing Done

- Hardware Type: NVIDIA H100 80GB HBM3 (8 GPUs)

Targeted validation:

- Built the standalone `liger_cute_kernels` wheel from this branch.
- Passed aligned forward and forward/backward CUDA graph tests.
- Passed unaligned `T=45` forward parity against the PyTorch reference.
- Passed unaligned `T=45` forward/backward parity against zero-padded `T=64`.
- Verified `T={31,32,33,45,63,64,65}` against the PyTorch reference.
- Confirmed the original unaligned regression test fails on the unpatched native core because rows 32-44 are incorrect.
- Independent code review found no correctness or synchronization issues in the unified path.

Performance:

- Aligned `T=16384`, two interleaved five-round runs: 13.0227 ms separate-tail vs. 12.9695 ms unified (0.41% faster).
- Unaligned `T=16385`, two interleaved five-round runs: 13.0091 ms separate-tail vs. 12.9673 ms unified (0.32% faster).
- Tail-heavy `T=45`, two interleaved five-round runs: 1.9948 ms separate-tail vs. 1.9611 ms unified (1.69% faster).
- Both variants use 168 registers; representative fused-kernel stack usage drops from as much as 408 bytes to 336 bytes.

Backward does not have the same floor-32 scheduling bug: its combine and reverse-dispatch kernels use bounded warp-per-token grid-stride loops. The new unaligned forward/backward test provides regression coverage for that path.


liger_cute_kernels/csrc/core/src/moe/moe.cu
```diff
@@ -721,6 +721,24 @@ moe_fused_kernel(
...
 			p.num_tokens,
 			p.top_k);
+
+		// Keep the optimized TMA path for every complete 32-token tile. The
+		// direct-store path handles only the disjoint 1–31 token tail and uses
+		// no shared memory, so it may run while another WG's final TMA store is
+		// draining.
+		const int tail_begin =
+			(p.num_tokens / kCombineWGTokens) * kCombineWGTokens;
+		if (tail_begin < p.num_tokens) {
```
kolehma8: I believe this could be handled in the main function as TMA can do OOB loads gracefully and the CUDA threads can handle the tail with a simple if-clause. Would that work here?

kolehma8 (dismissed):

liger_cute_kernels/csrc/core/src/moe/moe.cu
supercharleszhu: Yes. I updated the implementation to keep the tail in `combine_tokens_tma`. The grid now uses ceiling division; full warps take the original fast path, while only the final partial warp predicates metadata/weight reads and uses `cp.async` zero-fill. The full TMA output store safely discards rows beyond the descriptor extent. On 8x H100 this passes `T={31,32,33,45,63,64,65}`, aligned CUDA-graph tests, and a new unaligned forward/backward padded-parity test. Interleaved benchmarks were slightly faster than the separate-tail version: 0.41% at T=16384, 0.32% at T=16385, and 1.69% at T=45.

kolehma8 (approved):

PR #1385, opened 2026-08-15
Fix broken star history chart

Dessalines39394:
The star history chart on the README and docs was broken due to GitHub stargazer API restrictions. Update the chart to use a working data source that requires no API token so the chart renders correctly again.


vaibhavjindal (approved):
Thanks for this change!

PR #1390, opened 2026-08-16
Add Liger Kernel support for Muse Glimmer

buffett0323:
## Summary
Adds Liger Kernel support for Muse Glimmer (`model_type=muse_glimmer`), covering both capabilities requested in #1378:

1. FusedLinearCrossEntropy / skip-logits loss path, avoiding the `[B, S, V]` logits tensor (Muse Glimmer's `vocab_size` is 202,048)
2. `return_token_accuracy=True` → `outputs.token_accuracy`, so TRL's `SFTTrainer` can log `mean_token_accuracy` without a trainer-side fallback

Registered in `MODEL_TYPE_TO_APPLY_LIGER_FN`, so `use_liger_kernel=True` picks it up. Patches RoPE, RMSNorm, SwiGLU, vision LayerNorm (instance path only), CrossEntropy and FusedLinearCrossEntropy.

## Details
**Softcap.** HF computes `T * tanh(logits * m / T)` where `m = output_multiplier` and `T = final_logit_softcapping`, but Liger's fused softcap only implements `T * tanh(logits / T)`. `m` is folded into the hidden states instead, since `(m * h) @ W.T == m * (h @ W.T)` — scaling `[batch, seq, hidden]` rather than `[vocab, hidden]`.

**Two RMSNorm variants.** `MuseGlimmerRMSNorm` scales by `weight` directly (gemma-style fp32 compute, offset 0); `MuseGlimmerTextCenteredRMSNorm` scales by `(1 + weight)` from zero-init (offset 1). The `with_scale=False` instances, `qk_norm`, `embed_norm`, `perception_emb_norm`,  have no weight parameter and fall back to a torch implementation matching HF exactly; an assert guards against that fallback silently dropping a scale.

**`in_place=False`** on the TextCentered norms is required, not conservative: each decoder layer hands `pre_feedforward_layernorm` the same tensor it retains as the residual.

Note that `token_accuracy` is populated on the fused path only; with `skip_logits=False` (eval) it stays `None`, matching existing patched models such as Qwen3.

## Testing Done
- `test/transformers/test_muse_glimmer.py` (new) — instance-path numerical parity against the unpatched model: loss plus embedding / SwiGLU / RMSNorm gradients in fp32 and bf16, the non-fused softcap branch, vision LayerNorm, and the `token_accuracy` / `predicted_tokens` contract. The convergence suite patches at module level, so this is what covers `_apply_liger_kernel_to_instance` — the path TRL actually takes.
[...]

PR #1396, opened 2026-08-20
[Bug] FLCE Triton Fallback

buffett0323:
## Summary
The CuTe DSL backend's `LigerFusedLinearCrossEntropyFunction` currently raises a hard `RuntimeError` on capability misses (e.g., non-SM100 hardware, FP32, or unsupported reductions) instead of falling back to Triton. This breaks fused-linear-cross-entropy training across all models when using the `cutedsl` backend on non-SM100 hardware (reproducible on H100 via `LIGER_KERNEL_IMPL=cutedsl python -m pytest test/convergence/bf16/test_mini_models.py -k "mini_llama3 or mini_qwen2"`).

This PR makes capability misses fall back to the Triton kernel with a one-time warning. Found while validating #1390 against both backends, not specific to any model. The README already sets this expectation for ops with no CuTe DSL kernel, this PR extends it to an op that has one which declines the current hardware.

### Dispatch Implementation
- `LigerFusedLinearCrossEntropyFunction` is now a thin dispatcher rather than a `torch.autograd.Function`. The original autograd logic is renamed to `_NativeFusedLinearCrossEntropyFunction`.
- Call choice must occur before autograd records the node. Running Triton's `.apply` inside an autograd `forward()` would execute outside the graph and silently drop gradients.
- Call sites and `__module__` assertions remain unchanged.
- Validation executes before dispatch to guarantee uniform error messages across different GPU architectures.
- Low-level entry point `fused_linear_cross_entropy_forward` still raises on misses; only the dispatcher degrades.

## Testing Done
Validated across **H100 (SM90)** and **B200 (SM100)** to ensure both fallback and native execution paths work properly.

| Test Suite | H100 (SM90) | B200 (SM100) |
| :--- | :---: | :---: |
| `test/cutedsl` FLCE + convergence (`mini_llama3`, `mini_qwen2`, `muse_glimmer`) | 26 PASSED | 26 PASSED |
| Full `test/cutedsl/test_fused_linear_cross_entropy.py` (incl. `@sm100_required`) | — | 240 PASSED |

### Test Changes
- 3 tests now asset Triton fallback instead of a hard raise (`test_flce_falls_back_to_triton_on_non_sm100`, `..._for_fp32`, `..._for_reduction_none`).
- `test_flce_uses_native_path_when_supported` (`@sm100_required`) to prevent accidental permanent fallback.
- `test_flce_native_forward_still_raises_on_capability_miss` to check the low-level entry point.
- `test_cutedsl_flce_integration_does_not_raise_on_capability_miss` to verify the end-to-end integration path via a subprocess.

### DIFF
--- a/src/liger_kernel/ops/cutedsl/ops/fused_linear_cross_entropy.py
+++ b/src/liger_kernel/ops/cutedsl/ops/fused_linear_cross_entropy.py
@@ -1,3 +1,5 @@
+import functools
+import logging
 import operator
 
 import cutlass
@@ -11,6 +13,8 @@ from liger_kernel.ops.utils import amp_custom_bwd
 from liger_kernel.ops.utils import amp_custom_fwd
 from liger_kernel.ops.utils import compare_version
 from liger_kernel.utils import infer_device_arch
+
+logger = logging.getLogger(__name__)
 
 _SUPPORTS_OUT_DTYPE = compare_version("torch", operator.ge, "2.8.0")
 _FORWARD_GRAD_MIN_CHUNK = 512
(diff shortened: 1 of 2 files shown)


buffett0323:
Hi @PKUWZP , this is the pr extension for #1390 , please take a look, thanks!
Also, `cuTile` has a similar constraint (requires SM90/BF16 and raises). Should we address it here or in a follow-up PR?

PR #1397, opened 2026-08-21
Fix JSD underflow for low-probability vocabulary entries

mdeff:
## Summary

Fix NaNs in generalized JSD when both student and teacher probabilities underflow to zero in FP32.

The previous implementation constructed the mixture distribution in probability space. For log probabilities around `-104` or lower, both probabilities could underflow to zero, resulting in `log(0)` and NaN loss terms.

This change:
- computes `log(M)` using an element-wise log-sum-exp formulation
- evaluates the loss as weighted KL terms to avoid `0 * -inf`
- applies the fix consistently to the Triton, Ascend, and cuTile backends
- updates the PyTorch reference to use the equivalent stable formulation `torch.logaddexp`
- adds standalone and fused-linear regression tests

Both new regression tests return NaN on `main` and pass with this change.

## Testing Done

Hardware Type: A100
- [x] run `make test` to ensure correctness (18 unrelated chunked GRPO LUSPO failures)
- [ ] run `make test-convergence` to ensure convergence (does not exercise JSD)

Ascend and cuTile received the equivalent backend change but were not runtime-tested as I don't have the hardware.

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ops/jsd.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/jsd.py
@@ -74,24 +74,21 @@ def _jsd_kernel(
                     loss = X_prob * (X - Y)
                     dX = loss + X_prob
                 else:
-                    max_val = tl.maximum(tl.max(X, axis=0), tl.max(Y, axis=0))
+                    max_val = tl.maximum(X, Y)
                     X_shifted = X - max_val
                     Y_shifted = Y - max_val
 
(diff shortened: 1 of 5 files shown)


mdeff:
For context, I’ve regularly encountered this issue—JSD becoming NaN—while running QAT on Qwen3-4B-2507. I tested this patch for over 1,000 H100 GPU-hours and confirmed that it resolves the issue.

kashif:
cc @vaibhavjindal can be merged

PR #1418, opened 2026-08-28
feat(cutedsl): CuTeDSL backends for SwiGLU, GeGLU, Softmax, RoPE, KL-div, JSD

arde171:
# feat(cutedsl): CuTeDSL backends for SwiGLU, GeGLU, Softmax, RoPE, KL-div, JSD

**Head:** `arde171:arde/liger-C-cutedsl-elementwise` → **Base:** `main` — stacks on PR 2. PR **3/5**.

## Summary

Adds the **CuTeDSL elementwise/reduction** op backends, registered via `declare_op_locations` (Blackwell auto-select, Triton fallback). Includes several **correctness fixes** surfaced by an independent GPT-5.6 review of these kernels.

## What's added

- `ops/backends/_cutedsl/{swiglu,geglu,softmax,rope,kl_div,jsd}.py` + `ops/cutedsl/ops/rope.py`.
- `functional.py` — appends the `_cutedsl` paths for these six ops.
- `test/cutedsl/test_rope.py`, `test/ops/test_jsd_chunk_ignore_regression.py`.

## Correctness fixes included

- **SwiGLU / DTensor:** the CuTeDSL SwiGLU now **preserves DTensor (tensor-parallel) semantics** by delegating DTensor inputs to Triton (the previous path materialized the full tensor and returned plain tensors). Dense fast path unchanged. fp32 fwd tolerance calibrated for the fast sigmoid.
- **Softmax fp32 stability:** compute `exp2((x - max) * log2e)` instead of `exp2(x*log2e - max*log2e)` — shift-invariant, avoids fp32 cancellation at large offsets.
- **JSD chunked ignore-labels (P0):** decide the label-stream per chunk (was comparing a global non-ignore count with the per-chunk size), which had let ignored rows leak loss/gradient. Regression test added (asserts ignored rows get exactly zero input-grad across chunked shapes).
- **RoPE TMA guard:** `_tma_supported` rejects shapes whose launch block would exceed 1024 threads (e.g. bf16 head_dim=2048) → falls back to the portable token kernel.
- **GeGLU:** fp32 tolerance calibrated for the fast tanh-gelu.

## Correctness & testing (B300 sm_103 + H200 sm_90)

- Registration: `swiglu:[triton,cutedsl]`, `jsd:[triton,cutedsl]`, `jsd_loss_and_grad:[triton,cutedsl]` (inner op declared here, co-located with its `_cutedsl/jsd.py` impl).
[...]

PR #1417, opened 2026-08-28
feat(cutedsl): CuTeDSL backends for RMSNorm, LayerNorm, fused_add_rms_norm

arde171:
# feat(cutedsl): CuTeDSL backends for RMSNorm, LayerNorm, fused_add_rms_norm

**Head:** `arde171:arde/liger-B-cutedsl-norms` → **Base:** `main` — stacks on PR 1 (dispatcher). PR **2/5**.

## Summary

Adds the first **CuTeDSL** (NVIDIA CUTLASS Python DSL) backends — the three normalization ops — plus the shared `_cute_lib/` helpers. Registered into the dispatcher via `declare_op_locations`; **auto-selected on Blackwell** (where `cutlass.cute` is usable), with **Triton as the fallback** on Hopper/Ampere.

## What's added

- `ops/backends/_cutedsl/{rms_norm,layer_norm,fused_add_rms_norm}.py` — the backend impls.
- `ops/backends/_cutedsl/_cute_lib/*` — shared compile/copy/reduce/layout utilities used by all CuTeDSL kernels.
- `ops/cutedsl/ops/{rms_norm,rms_norm_fastpath}.py` — the low-level CuTeDSL kernels.
- `functional.py` — appends the `_cutedsl.{rms_norm,layer_norm,fused_add_rms_norm}` paths to those ops' declarations.

## Design

`min_cc=(9,0)` (Hopper+). The two-pass RMSNorm keeps a per-SM fp32 `dW` partial reduced across SMs; forward/backward mirror Quack's SM-count heuristic. The dispatcher auto-selects CuTeDSL on Blackwell and defers to Triton where `cutlass.cute` isn't usable.

## Correctness & testing (B300 sm_103 + H200 sm_90, torch 2.13/cu130)

- Registration: `rms_norm: [nvidia-triton, nvidia-cutedsl]`.
- Full backend suite `test/backends` + `test/ops`, deterministic (`-p no:randomly`): **B300 474 passed / 0 failed**, **H200 474 passed / 0 failed** (49 skipped); also confirmed on **H100 474 / 0**.
- Bf16 weight-grad (`dw`) parity uses a reduction-appropriate tolerance (it's a batch sum-reduction; the strict elementwise tol is for fwd/dx).
- ruff clean.

## Scope

CuTeDSL norm backends only; additive registration. No change to Triton behavior; non-Blackwell falls back to Triton.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>
---
[...]

PR #1419, opened 2026-08-28
feat(cutedsl): CuTeDSL backends for CrossEntropy, FusedLinearCE, FusedLinearJSD

arde171:
# feat(cutedsl): register CuTeDSL CrossEntropy / FusedLinearCE / FusedLinearJSD backends

**Head:** `arde171:arde/liger-D-cutedsl-ce` → **Base:** `main`. Stacks on #1418. PR **4/5**.

## What this PR is (and is not)

This is a **dispatcher-wiring + one correctness fix** PR. The CuTeDSL cross-entropy-family **kernels already exist on `main`** (added by #1334/#1355/#1370/#1380 etc.). This PR only **registers** them with the multi-backend dispatcher from #1416 and adds the composed-op adapters. **It does not modify any existing kernel implementation** — `fused_scaled_cross_entropy_sm90.py` (fragment forward, #1370), `_sm100_gemm.py` (selector, #1380), and `cross_entropy.py` `num_warps` are all **byte-identical to `upstream/main`**.

Diff: **532 insertions / 9 deletions**, 6 files.

## What's added
- `ops/backends/_cutedsl/{cross_entropy,fused_linear_cross_entropy,fused_linear_jsd}.py` — `register_op` adapters, plus the shared inner primitives `cross_entropy_loss_and_grad` / `jsd_loss_and_grad`.
- `functional.py` — `declare_op_locations` for the above.
- `ops/cutedsl/ops/cross_entropy.py` — **+33/-9**, the fp32 correctness fix only (see below).
- `test/cutedsl/test_cross_entropy.py` — **+46**, a focused confident-prediction precision test.

## Default-selection safety (addresses the perf review)
Large-vocab CE is **HBM-bandwidth-bound**. Measured on B200 (T=8192, V=128256, fwd+bwd): standalone cutedsl CE is **tied** in bf16 (1.185 vs Triton 1.179 ms) and **~5% slower** in fp32 (1.894 vs 1.803). So:
- **standalone `cross_entropy` cutedsl → `preference_rank=60`** (below Triton's 50): **Triton stays the auto-default**; cutedsl is opt-in via `impl="nvidia-cutedsl"`.
- **`cross_entropy_loss_and_grad` inner primitive stays `preference_rank=10`** — it's the kernel the fused-linear CE path composes.

## Correctness fix (CuTeDSL analogue of Triton #1329)
The cutedsl CE backward now **recomputes the true-class gradient `dx_y` fully in fp32 and overwrites the target slot with a single store** (was a read-modify-write of the already-low-precision value). `dx_y = softmax(x_y) − (1−ls)` cancels catastrophically as `softmax(x_y) → 1`, so the `−(1−ls)` term must be folded in fp32 before the round. Covers weighted / label-smoothing / z-loss / softcap / ignore-index; new `test_ce_true_class_grad_confident_predictions` locks it.

## Testing (Blackwell sm_100 ×2 + Hopper sm_90), deterministic `-p no:randomly`
| suite | B200 sm_100 | H200 sm_90 |
|---|---|---|
| `test/cutedsl/test_cross_entropy.py` (incl. new precision test) | **166 / 0** | ✓ |
| `test/transformers/test_cross_entropy.py` | **185 / 0** | 185 / 0 |
| `test/backends` + `test/ops` (D-only) | **634 / 0** | ✓ |

ruff check + format clean.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>
---
**Multi-backend series — review/merge in order:** #1416 (A) → #1417 (B) → #1418 (C) → **#1419 (D)** → #1420 (E · cuTile). Stacked; diffs auto-narrow as earlier PRs land.

### DIFF
--- a/src/liger_kernel/functional.py
+++ b/src/liger_kernel/functional.py
@@ -99,7 +99,10 @@ declare_op_locations(
 # uses an online-softmax reduction adapted from Quack.
 declare_op_locations(
     "cross_entropy",
-    ("liger_kernel.ops.backends._triton.cross_entropy",),
+    (
+        "liger_kernel.ops.backends._triton.cross_entropy",
+        "liger_kernel.ops.backends._cutedsl.cross_entropy",
+    ),
 )
 
 # cross_entropy_loss_and_grad: per-chunk CE primitive used by
@@ -107,7 +110,10 @@ declare_op_locations(
 # and picks up CuTe DSL on Hopper+.
 declare_op_locations(
     "cross_entropy_loss_and_grad",
-    ("liger_kernel.ops.backends._triton.cross_entropy",),
+    (
+        "liger_kernel.ops.backends._triton.cross_entropy",
+        "liger_kernel.ops.backends._cutedsl.cross_entropy",
+    ),
 )
 
 # fused_add_rms_norm: Triton (universal) + opt-in CuTe DSL (Hopper+). The
@@ -146,7 +152,10 @@ declare_op_locations(
 # JSD inputs to fp32, which safely falls back to Triton.
 declare_op_locations(
     "fused_linear_jsd",
-    ("liger_kernel.ops.backends._triton.fused_linear_jsd",),
+    (
+        "liger_kernel.ops.backends._triton.fused_linear_jsd",
+        "liger_kernel.ops.backends._cutedsl.fused_linear_jsd",
+    ),
 )
 
 # fused_linear_cross_entropy: Triton (universal) + CuTe DSL (Hopper+). The
@@ -155,7 +164,10 @@ declare_op_locations(
 # the CuTe DSL kernel on Hopper+).
 declare_op_locations(
     "fused_linear_cross_entropy",
-    ("liger_kernel.ops.backends._triton.fused_linear_cross_entropy",),
+    (
+        "liger_kernel.ops.backends._triton.fused_linear_cross_entropy",
+        "liger_kernel.ops.backends._cutedsl.fused_linear_cross_entropy",
+    ),
 )
 
 
--- a/src/liger_kernel/ops/backends/_cutedsl/cross_entropy.py
+++ b/src/liger_kernel/ops/backends/_cutedsl/cross_entropy.py
@@ -1,0 +1,254 @@
+"""CuTe DSL backend registration for ``cross_entropy``.
+
+Thin adapter: the kernel and ``LigerCrossEntropyFunction`` live in
+:mod:`liger_kernel.ops.cutedsl.ops.cross_entropy` (the CuTe DSL operator
+package).  This file only wraps it in a :func:`register_op`-decorated callable
+that the multi-DSL dispatcher can find — mirroring the
+:mod:`liger_kernel.ops.backends._triton.cross_entropy` sibling.
+
+Capability: requires the ``cutlass.cute`` package and compute capability >=
+sm_90 (Hopper or newer), matching the rms_norm / softmax CuTe DSL siblings.
+
+The underlying CuTe DSL cross-entropy kernel uses an online-softmax reduction
+adapted from Quack (Apache-2.0) and is numerically identical to the Triton
+kernel (:mod:`liger_kernel.ops.cross_entropy`).
+"""
+
+from __future__ import annotations
+
+from typing import Optional
+from typing import Tuple
+
+import torch
+
+from liger_kernel.backends import Capability
+from liger_kernel.backends import register_op
+from liger_kernel.backends.dispatch import emit_fallback_warning
+from liger_kernel.ops.cutedsl.ops.cross_entropy import LigerCrossEntropyFunction
+
+_CUTEDSL_CE_TOLERANCES = {
+    torch.float16: {"atol_fwd": 5e-3, "atol_bwd": 5e-2, "rtol_fwd": 1e-3, "rtol_bwd": 1e-2},
+    torch.bfloat16: {"atol_fwd": 2e-2, "atol_bwd": 1e-1, "rtol_fwd": 1e-2, "rtol_bwd": 2e-2},
+    torch.float32: {"atol_fwd": 1e-5, "atol_bwd": 1e-4, "rtol_fwd": 1e-5, "rtol_bwd": 1e-4},
+}
+
+
+@register_op(
+    "cross_entropy",
+    impl_name="nvidia-cutedsl",
+    capability=Capability(
+        min_cc=(9, 0),
+        modules=["cutlass.cute", "torch"],
+    ),
+    modes=("default",),
+    default_mode="default",
+    # Ranked BELOW Triton (50) for the standalone cross_entropy op: large-vocab CE is
+    # HBM-bandwidth-bound, and on Blackwell (B200/B300) the CuTe DSL kernel measured
+    # tied-to-slower than Triton (bf16 ~=, fp32 ~-5% @ T8192/V128256). Triton therefore
+    # stays the auto-default here; the CuTe DSL path remains available via explicit
+    # impl="nvidia-cutedsl". (The rank-10 cross_entropy_loss_and_grad primitive below is
+    # unchanged — it is the inner kernel the fused-linear CE path composes.)
+    preference_rank=60,
+    tolerances=_CUTEDSL_CE_TOLERANCES,
+    notes="CuTe DSL cross-entropy for Hopper+ (sm_90+); online-softmax reduction. Opt-in on Blackwell.",
+)
+def cross_entropy_cutedsl(
+    input: torch.Tensor,
+    target: torch.Tensor,
+    weight: Optional[torch.Tensor] = None,
+    ignore_index: int = -100,
+    lse_square_scale: float = 0.0,
+    label_smoothing: float = 0.0,
+    reduction: str = "mean",
+    softcap: Optional[float] = None,
+    return_z_loss: bool = False,
+    return_token_accuracy: bool = False,
+    return_predicted_tokens: bool = False,
+    *,
+    mode: Optional[str] = None,
+):
+    """CuTe DSL cross-entropy dispatch entry point.
+
+    ``mode`` is accepted only for API parity with the other backends; the only
+    valid value is ``"default"`` (or ``None``).
+    """
+    if mode not in (None, "default"):
+        raise ValueError(f"CuTe DSL cross_entropy has only mode='default'; got mode={mode!r}.")
+    vector_width = 16 // input.element_size()
+    if input.shape[-1] % vector_width:
+        from liger_kernel.ops.backends._triton.cross_entropy import cross_entropy_triton
+
+        emit_fallback_warning(
+            "cross_entropy",
+            "nvidia-cutedsl",
+            "nvidia-triton",
+            f"vocab size {input.shape[-1]} is not divisible by vector width {vector_width}",
+        )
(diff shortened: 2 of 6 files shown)


kolehma8 (dismissed):

arde171:
Rebased onto the new `main` now that #1418 landed (drops #1416-#1418's commits so the diff is D-only: CuTeDSL CrossEntropy / FusedLinearCE / FusedLinearJSD, checkstyle green). Content unchanged from your approval — the rebase just auto-dismissed it per the stale-review policy. Please re-approve so the queue can land it. 🙏 (#1420 is the last one and will follow the same way.)

arde171:
Pushed a follow-up fix to the CuTeDSL cross-entropy backward (force-update `f7da0123` → `6d07f7e8`).

**Why:** this backend registers at `preference_rank=10` / `min_cc=(9,0)`, so on Hopper+ it becomes the default for `LigerCrossEntropyLoss`. The previous backward computed the true-class gradient via a read-modify-write of the already-low-precision `gX[y]` and then added `-(1-ls)/N`. For confident predictions (`softmax(x_y) → 1`) this difference cancels catastrophically, losing significant bits in bf16/fp16 — the exact failure mode #1430 fixed in the Triton kernel. With this backend as the Hopper+ default, `test_correctness_true_class_grad_confident_predictions` (added by #1430) regressed (e.g. `mean/0.0/bf16`: err `1.5e-5` vs bound `1.8e-6`).

**Fix:** recompute the entire `dx_y` once, fully in fp32 — `softmax_y` from the fp32 logit `ori_xy`, plus the same coef / label-smoothing / z_loss / softcap terms as the per-element loop, and the `-(1-ls)·weight_y/N` true-class term — then **overwrite** `gX[y]` with a single store (no read-modify-write). Mirrors #1430's Triton approach.

**Verified (0 fail):**
| Arch | #1430 precision | full CE | backends+ops |
|---|---|---|---|
| B200 sm_100 (×2 indep pods) | 8 | 185 | 847 |
| H200 sm_90 | 8 | 185 | 640 (cuTile skips <sm_100) |

Also passes `ruff check` + `ruff format`. Ready for re-review.

arde171:
@kolehma8 — the force-push (CE fp32 true-class-grad fix, detailed above) dismissed your prior approval. Could you re-approve when you have a moment? #1420 is stacked on this and blocked behind it. Both are green on B200 sm_100 (×2) + H200 sm_90 and `ruff`-clean. Thanks!

arde171:
Synced onto latest `main` (`0a5e9a55`, incl #1428 TP-CE, #1435 SwiGLU-Blackwell, #1436 cuTile-CI, #1437) — branch is now **up-to-date** (behind_by=0), clean rebase (no conflicts). Re-verified on B200 sm_100: registration `cross_entropy:[nvidia-triton, nvidia-cutedsl]`, #1430 precision test **8/0**, `test/backends`+`test/ops` **847/0**, `test_cross_entropy.py` **185/0**, `test_swiglu.py` **81/0** (coexists with #1435). ruff clean; auto-merge armed. @kolehma8 ready for your re-approval — #1420 is stacked behind this. Thanks!

src/liger_kernel/ops/cutedsl/ops/_fused_scaled_cross_entropy_forward_sm90.py
```diff
@@ -0,0 +1,1104 @@
+"""SM90 fused scaled cross entropy **forward** kernel: producer-less
```
kolehma8: Any performance numbers for this?

arde171:
Thanks for the careful review — you were right to push on this. I did a deep pass (and cross-checked against our internal source-of-truth branch); it surfaced two stale-snapshot reverts hiding in the diff, plus a genuinely questionable default. **Force-pushed `f51bdc83`** which reduces D from **2897/213** to **532/9 lines** — it's now purely dispatcher wiring + one correctness fix. Point by point:

**1 & 5 — "FSLCE perf numbers" / "is this just moving code?"**
The diff had accidentally re-added the old `_fused_scaled_cross_entropy_forward_{mfast,sm90}.py` and switched the SM90 entry point to them — which **reverted #1370's `fragment` forward**. That branch was cut from a stale snapshot. I've reverted the SM90 entry + test to `upstream/main` (byte-identical) and deleted the two obsolete files. **There is no new/modified FSLCE kernel in this PR** — it uses #1370's fragment forward unchanged, so #1370's existing H100 data stands. Separately I found the diff was **deleting #1380's `_sm100_gemm.py` `_select_epilogue_config` selector** (hard-coding stages/swizzle) — also reverted; `_sm100_gemm.py` is now byte-identical to main. So yes: **D is now dispatcher exposure of existing kernels, not new kernels or code movement.**

**2 — "don't add CuTeDSL kernels without complete benchmarking" (the real issue: default selection)**
The valid concern is which impl runs **by default**. I benchmarked on B200 (T=8192, V=128256, fwd+bwd): standalone cutedsl CE is **tied** bf16 (1.185 vs 1.179 ms) and **~5% slower** fp32 vs Triton — it's HBM-bandwidth-bound, so cutedsl's tuning doesn't help there. FLCE inner-CE choice is **negligible** (22.0 vs 22.0 ms; FLCE is GEMM-bound). So I **moved standalone `cross_entropy` cutedsl to `preference_rank=60` (below Triton)** — Triton stays the auto-default; cutedsl is opt-in via `impl=`. The `cross_entropy_loss_and_grad` inner primitive stays rank 10 (the fused-linear path composes it). I also **reverted the elaborate Blackwell `num_warps` policy to upstream's simple version** — that's a real optimization but belongs in a separate, benchmarked Blackwell PR, not here.

**3 — "one kernel vs a family?"**
They can't be one device kernel: CE consumes materialized logits → in-place `dlogits`; FLCE consumes hidden states + classifier weight, chunks the projection, and projects `dlogits → dX/dW/db`; FLJSD consumes two projections and a different divergence. What *is* shared is correctly factored: `cross_entropy_loss_and_grad` / `jsd_loss_and_grad` are the reusable primitives the fused-linear adapters compose. So this is a family of adapters over shared primitives, not three duplicate kernels.

**4 — "#1420 also does major CuTeDSL changes"**
It doesn't. #1420's *own* diff (its commit range) is only `ops/backends/_cutile/*` + 6 lines of `functional.py` + one test — **zero CuTeDSL files**. GitHub shows CuTeDSL in #1420 because it's **stacked on this unmerged PR**; review E by its parent→tip range and it's cuTile-only. It'll auto-narrow once D lands.

**Correctness fix retained:** the cutedsl CE backward now recomputes `dx_y` fully in fp32 and overwrites once (analogue of Triton **#1329** — thanks, I had mis-cited #1430), with a new focused confident-prediction test. Verified on B200 sm_100 ×2 + H200 sm_90: cutedsl CE 166, transformers CE 185, backends+ops 634, 0 failed; ruff clean.

Open item I'd value your steer on: the FLCE (rank 10) and FLJSD (rank 4) native paths also auto-select on Hopper — I can either publish a forced-backend matrix for every admitted arch, or drop them to opt-in until that data exists. Happy to do whichever you prefer.

arde171:
Ran the full forced-backend performance matrix on B200 (sm_100), bf16, T=8192, V=128256, H=4096, forward+backward, 30-iter timing, 3 rounds (variance <0.1%):

| op | Triton | CuTeDSL | delta |
|---|---|---|---|
| standalone CE | 1.183 ms | 1.189 ms | −0.5% (tied/slower) |
| FLCE (fused-linear CE) | 22.10 ms | 22.07 ms | +0.1% (tied) |
| **FLJSD (fused-linear JSD)** | 48.48 ms | **46.94 ms** | **+3.3% (CuTeDSL win, reproducible 3.2–3.4%)** |

This resolves the default-selection question with data:

- **standalone `cross_entropy`** — CuTeDSL is tied/slower (HBM-bandwidth-bound), so it's now rank 60 → **Triton stays the default**, CuTeDSL opt-in. ✅
- **FLCE** — tied (the path is GEMM-bound; the inner CE is <1% of runtime), so cutedsl rank 10 is a no-op perf-wise — no regression, kept as-is. ✅
- **FLJSD** — CuTeDSL is a **measured +3.3% win** and reproducible, so its rank-4 default is justified. ✅

So no further rank changes are needed; the earlier standalone-CE demotion was the only default that wasn't perf-justified, and that's fixed in `f51bdc83`. (These are Blackwell numbers; I don't currently have a live Hopper pod for the sm_90 leg — happy to add H100/H200 rows when one frees up, but the Blackwell data alone already justifies every current default and removes the only regression.)

arde171:
### Detail on the one correctness change (`ops/cutedsl/ops/cross_entropy.py`)

Several folks asked what the `cross_entropy.py` hunk does — it's the CuTeDSL analogue of the Triton fix in #1329 (the true-class gradient precision fix for confident predictions). Summary of **why** and **what**:

**Background.** The backward writes `dx_i = softmax(x_i)` for non-target classes and `dx_y = softmax(x_y) − (1 − ls)` for the true class `y`. The pass-2 loop writes the per-element part `softmax·coef (+smoothing)` for *every* column — including `y` — into the in-place gradient buffer `gX`, **rounded to the output dtype (bf16/fp16)**. A one-thread epilogue then patches `gX[y]` to add the true-class `−(1−ls)` term.

**The bug (old code):**
```python
dxy  = -(1 - ls) * w_y * inv_n_loss     # only the "−1" term
corr = gX[y].to(Float32) + dxy          # read the ALREADY-bf16-rounded gX[y], add
gX[y] = corr.to(gX.element_type)        # store  (read-modify-write)
```
For a confident prediction `softmax(x_y) → 1`, `dx_y = softmax(x_y) − 1` is a tiny difference of two ~1.0 numbers — **catastrophic cancellation**. Because `gX[y]` was already rounded to bf16 (step ≈ 2⁻⁸ ≈ 0.0039 near 1.0), subtracting 1 exposes that rounding error as the whole answer:

```
softmax(x_y)=0.9995 → stored bf16 0.99609  (err 0.0034)
old:  0.99609 − 1.0 = −0.00391      true: 0.9995 − 1.0 = −0.00050   →  ~680% wrong
```
This only bites once the model is confident (late training / easy tokens), which is why random-logit tests didn't catch it.

**The fix (new code):** recompute the **entire** `dx_y` in fp32 (softmax at `y` from the fp32 logit `ori_xy`, plus coef / label-smoothing / z-loss / softcap), do the `−(1−ls)` subtraction in fp32, and **overwrite `gX[y]` with a single store**:
```
0.9995 − 1.0 = −0.00050  (fp32) → store bf16 −0.00050   →  err ≈ 2e-7, correct
```
Net idea: **old = round then subtract 1 (error exposed); new = subtract 1 in fp32 then round once** (small result stored with good relative precision). The extra lines just rebuild the loop's exact per-element value in fp32 so the overwrite is correct for every variant (weighted / label-smoothing / z-loss / softcap / ignore-index / all reductions).

**Why it's in this PR:** the CuTeDSL CE kernel already lived on `main` but was never dispatcher-reachable; #1419 wires it in (as the fused-linear inner primitive + explicit `impl="nvidia-cutedsl"`), so it must match the Triton reference's precision. New `test_ce_true_class_grad_confident_predictions` drives `softmax(x_y)→1` and requires the cutedsl true-class grad to be no worse than 4× torch's low-precision backward (bf16/fp16 × mean/sum × smoothing on/off). It's the **only** implementation change in the PR — everything else is dispatcher registration.

kolehma8 (approved):

PR #1421, opened 2026-08-29
fix(mlp): keep liger_kernel.ops importable without TensorDescriptor

zheliuyu:
## Summary

- `LigerMLP` (#1357) does a hard `from triton.tools.tensor_descriptor import TensorDescriptor`. That module first shipped in official Triton **3.4.0**, so any earlier / vendor build fails while importing `liger_kernel.ops`.
- `ops/__init__.py` always loads default kernels before backend overlay, so one missing import takes down every op (RMSNorm, Softmax, RoPE, …), not just MLP.
- Make the import optional. `LigerMLPFunction.forward` raises a clear `RuntimeError` if the API is missing. GPU / Triton >= 3.4 is unchanged.
- Skip `test/transformers/test_mlp.py` when `triton.tools.tensor_descriptor` is not installed.

## Testing Done

- `pytest test/transformers/ --collect-only` on a Triton 3.2 build: collection succeeds (no longer 33 collection errors).
- `test_mlp.py` is skipped when the API is absent; other transformers tests collect and run as before.

### DIFF
--- a/test/transformers/test_mlp.py
+++ b/test/transformers/test_mlp.py
@@ -1,3 +1,5 @@
+import importlib.util
+
 import pytest
 import torch
 
@@ -18,6 +20,11 @@ except ImportError:
     FALCON_H1_AVAILABLE = False
 
 device = infer_device()
+
+pytestmark = pytest.mark.skipif(
+    importlib.util.find_spec("triton.tools.tensor_descriptor") is None,
+    reason="LigerMLP requires triton.tools.tensor_descriptor",
+)
 
 DIFF_THRESHOLD = 1e-5
 # Why the FalconH1 tests use a looser tolerance only for bf16.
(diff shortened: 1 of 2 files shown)


zheliuyu:
@Tcc0403 please take a look at this PR. cc @Pearblossom-M

test/transformers/test_mlp.py
```diff
@@ -19,6 +21,11 @@
 
 device = infer_device()
 
+pytestmark = pytest.mark.skipif(
+    importlib.util.find_spec("triton.tools.tensor_descriptor") is None,
+    reason="LigerMLP requires triton.tools.tensor_descriptor",
+)
+
```
Tcc0403: nit: replace with `pytest.importorskip`

Tcc0403 (commented):
Thanks for catching it, just a nitpicking change for readability

test/transformers/test_mlp.py
zheliuyu: Done, switched to `pytest.importorskip`. Thanks for the review!

Tcc0403 (approved):
Thanks!

PR #1426, opened 2026-08-31
[NPU] Optimize Ascend CE/FLCE, RMSNorm, and RoPE kernels

sunyi0505:
## Summary

The core idea of the fusion operator in torch_npu is to combine multiple scattered computation operations (operators) into one larger operation. This aims to reduce unnecessary overhead, make full use of the hardware features of Ascend NPU, and thus significantly improve the performance of model training and inference. Typically, we directly call the corresponding interfaces in torch_npu, for example: torch_npu.npu_cross_entropy_loss/torch_npu.npu_rms_norm/torch_npu.npu_rotary_mul/torch_npu.npu_swiglu.

Tune the Ascend Triton backends for CrossEntropy, FusedLinearCrossEntropy, RMSNorm, and RoPE. The common theme is fewer HBM round‑trips, constexpr‑DCE'd masks, software‑pipelined tl.range, and in‑place buffers so last‑layer training does not peak at two full BT×V tensors, so as to achieve performance on par with or even exceeding the fused operators in torch_npu. Refer to [torch_npu_baselines.py](https://github.com/user-attachments/files/31633905/torch_npu_baselines.py) for the complete implementation script

### CrossEntropy
- Plain forward processes contiguous row chunks per vector core (sequential MTE2) instead of one row per program.
- Full tiles are unmasked; `HAS_TAIL` is constexpr so the remainder mask is DCE'd when `V` divides `BLOCK_SIZE`.
- `tl.range` pipelines MTE2 / VEC / MTE3; OOB labels are clamped before DMA to avoid a fault ahead of the host assert.
- Dedicated `get_plain_ce_block_size` for the smaller live set on the no-weight path.
- Last-layer backward (`grad_output == 1`) overwrites the logits buffer in-place.
### FusedLinearCrossEntropy
- One Cube GEMM for the full `(BT, V)` logits instead of many token chunks.
- Reuses the optimized CE kernels; writes `dlogits` in-place into the logits buffer (no extra BT×V).
- Optionally retains forward logits (capped at `min(4 GiB, device_mem / 8)`) to skip the backward logits GEMM.
- `torch.mm(..., out=)` for `dX` / `dW` when dtypes match, avoiding an extra V×H workspace.
### RMSNorm
- Full-width single-pass fused kernels when a row fits UB (same pattern as Ascend LayerNorm), avoiding a second load of X after the rstd reduction.
- 2D fused path (`ROWS_PER_BLOCK × n_cols`) when UB allows; column tiling only when a full row does not fit.
- On 910B (`BT=8192, H=4096, bf16`) the previous two-pass tiled forward was ~350 µs (`aiv_mte2_ratio=0.69`) vs `npu_rms_norm` at ~107 µs.
### RoPE
- Native BNSD layout; the old BSND transpose round-trip was ~40% of Liger RoPE time.
- Work maps to `(batch, head, s_tile)` so 48 AIV cores stay occupied.
- Fast path: contiguous `(BLOCK_S, D)` DMA, `rotate_half` in UB via `extract_slice` / `insert_slice`.
- Sequential q-then-k launches (dual-stream fights over the same 48 AIV cores). Short aligned shapes use one two-phase launch to hide the second host launch.

## Testing Done

pytest test/transformers/test_cross_entropy.py test/transformers/test_rms_norm.py test/transformers/test_rope.py test/transformers/test_fused_linear_cross_entropy.py 
image

- Hardware Type:  Atls 800T A3(X86)

| Kernel name | Performance |
|---------------|----------------|
| rope forward    | image|
| rope backward | image |
| rope full           | image|
| cross_entropy forward | image|
| cross_entropy backward | image|
| cross_entropy full | image |
| cross_entropy no grad forward | image|
| rms_norm forward | image |
| rms_norm backward | image|
| rms_norm full | image|
| fused_linear_cross_entropy forward |image|
| fused_linear_cross_entropy backward | image|
| fused_linear_cross_entropy full |image|
| swiglu forward |image|
| swiglu backward |image|
| swiglu full |image|

### DIFF
--- a/src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
+++ b/src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
@@ -96,7 +96,8 @@ def liger_cross_entropy_forward_kernel(
     target_row_ptr = Y_ptr + row_start_int64
     logits_row_ptr = X_ptr + row_start_int64 * X_stride
     loss_row_ptr = loss_ptr + row_start_int64
-    lse_row_ptr = lse_ptr + row_start_int64
+    if RETURN_LSE:
+        lse_row_ptr = lse_ptr + row_start_int64
     if RETURN_Z_LOSS:
         z_loss_row_ptr = z_loss_ptr + row_start_int64
     if RETURN_TOKEN_ACCURACY:
@@ -208,55 +209,68 @@ def liger_cross_entropy_forward_kernel_plain(
     X_stride,
     Y_ptr,
     loss_ptr,
+    lse_ptr,
     n_cols,
     n_rows,
     ce_stats_ptr,
     ignore_index,
     reduction: tl.constexpr,
     BLOCK_SIZE: tl.constexpr,
+    HAS_TAIL: tl.constexpr,
+    RETURN_LSE: tl.constexpr,
 ):
     """
-    Plain CE forward fast path (no weight/softcap/smoothing/z-loss/metrics).
-    1 program processes 1 row to maximize parallelism (unlike the generic kernel
-    which chunks rows per program for flexibility).
-    Writes per-row loss to ``loss_ptr`` (float32 recommended).
+    Plain CE fast path (no weight/softcap/smoothing/z-loss/metrics).
+
+    Contiguous row chunks per vector core keep MTE2 sequential. Full tiles are
+    unmasked; ``HAS_TAIL`` is constexpr so the masked remainder is DCE'd when
+    ``n_cols`` divides ``BLOCK_SIZE``. ``tl.range`` pipelines MTE2/VEC/MTE3.
     """
-    row = tl.program_id(0)
-    if row >= n_rows:
-        return
-
-    row_i64 = row.to(tl.int64)
-    y = tl.load(Y_ptr + row_i64)
-
-    # Pre-load mean scaling. For sum/none we treat scale as 1.0.
+    pid = tl.program_id(0)
+    nprog = tl.num_programs(0)
     inv_n = tl.load(ce_stats_ptr + 0).cast(tl.float32)
+    cols = tl.arange(0, BLOCK_SIZE)
+    n_full = (n_cols // BLOCK_SIZE) * BLOCK_SIZE
+    row_chunk = (n_rows + nprog - 1) // nprog
+    row_start = pid * row_chunk
+    row_end = tl.minimum(row_start + row_chunk, n_rows)
 
-    if y == ignore_index:
-        tl.store(loss_ptr + row_i64, 0.0)
-        return
-    logits_row_ptr = X_ptr + row_i64 * X_stride
-    m = float("-inf")
-    d = 0.0
-    # Keep x[y] as a direct load outside the scan. Folding it into the block
-    # loop adds extra selection state and tends to increase UB/register pressure
-    # on Ascend.
-    x_y = tl.load(logits_row_ptr + y).cast(tl.float32)
-    for i in range(0, n_cols, BLOCK_SIZE):
-        offs = i + tl.arange(0, BLOCK_SIZE)
-        x = tl.load(
-            logits_row_ptr + offs,
-            mask=offs < n_cols,
-            other=float("-inf"),
-        ).cast(tl.float32)
-        block_max = tl.max(x)
-        m_new = tl.maximum(m, block_max)
-        d = d * tl.exp(m - m_new) + tl.sum(tl.exp(x - m_new))
-        m = m_new
-    lse = m + tl.log(d)
-    loss = lse - x_y
-    if reduction == "mean":
-        loss = loss * inv_n
-    tl.store(loss_ptr + row_i64, loss)
+    for row in tl.range(row_start, row_end):
+        row_i64 = tl.cast(row, tl.int64)
+        y = tl.load(Y_ptr + row_i64)
+        logits_row_ptr = X_ptr + row_i64 * X_stride
+
+        if y == ignore_index:
+            tl.store(loss_ptr + row_i64, 0.0)
+            if RETURN_LSE:
+                tl.store(lse_ptr + row_i64, 0.0)
+        else:
+            # Clamp so an out-of-range target cannot DMA-fault before the host
+            # OOB assert (queued after this kernel) raises AssertionError.
+            y_idx = tl.minimum(tl.maximum(y, 0), n_cols - 1)
+            x_y = tl.load(logits_row_ptr + y_idx).cast(tl.float32)
+            m = float("-inf")
+            d = 0.0
+            for i in tl.range(0, n_full, BLOCK_SIZE):
+                x = tl.load(logits_row_ptr + i + cols).cast(tl.float32)
+                block_max = tl.max(x)
+                m_new = tl.maximum(m, block_max)
+                d = d * tl.exp(m - m_new) + tl.sum(tl.exp(x - m_new))
+                m = m_new
+            if HAS_TAIL:
+                offs = n_full + cols
+                x = tl.load(logits_row_ptr + offs, mask=offs < n_cols, other=float("-inf")).cast(tl.float32)
+                block_max = tl.max(x)
+                m_new = tl.maximum(m, block_max)
+                d = d * tl.exp(m - m_new) + tl.sum(tl.exp(x - m_new))
+                m = m_new
+            lse = m + tl.log(d)
+            if RETURN_LSE:
+                tl.store(lse_ptr + row_i64, lse)
+            loss = lse - x_y
+            if reduction == "mean":
+                loss = loss * inv_n
+            tl.store(loss_ptr + row_i64, loss)
 
 
 @triton.jit
@@ -276,6 +290,7 @@ def liger_cross_entropy_backward_kernel_no_weight(
     reduction: tl.constexpr,
     BLOCK_SIZE: tl.constexpr,
     HAS_LSE: tl.constexpr,
+    HAS_TAIL: tl.constexpr,
 ):
     """
     Specialized backward kernel for the common path without class weights, softcap, z-loss, or label smoothing.
@@ -286,7 +301,7 @@ def liger_cross_entropy_backward_kernel_no_weight(
         X_stride (int64): Stride between consecutive logits rows.
         Y_ptr: Target class index per row; ignored rows receive zero ``dX``.
         lse_ptr: Per-row LSE (fp32) when ``HAS_LSE``; otherwise aliases per-row loss buffer for LSE reconstruction.
-        grad_output_ptr: Scalar loss gradient or per-row vector (see ``HAS_GRAD_OUTPUT_VECTOR``).
+        grad_output_ptr: Scalar loss gradient, or per-row vector when ``reduction == "none"``.
         grad_output_stride (int64): Stride for vector ``grad_output`` (``0`` when scalar).
         dX_ptr: Output logits gradient; same logical layout as ``X_ptr`` with stride ``dX_stride``.
         dX_stride (int64): Stride between ``dX`` rows.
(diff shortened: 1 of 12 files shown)


zheliuyu:
Nice work! Before formal review: 
- please include `SwiGLU` perf improvements using the same optimization techniques — it's a common kernel alongside the ones in this PR.
- please share minimal `torch_npu` run scripts for each kernel (versions + commands), since not all reviewers are familiar with the `torch_npu` setup. Comment attachments are fine.
- please report peak device memory usage.

sunyi0505:
Prerequisite for this experiment: The triton‑ascend operator involves compilation time, so the second‑run time shall be taken as the benchmark.
I conducted complete model training tests using llamafactory with the qwen3‑8b model. The script is as follows:
```shell
### model
model_name_or_path: /mnt/weight/Qwen3-8B
trust_remote_code: true
# use_v1_kernels: true
enable_liger_kernel: true
flash_attn: fa2

### method
stage: sft
do_train: true
finetuning_type: lora
lora_rank: 8
lora_target: all

### dataset
dataset: alpaca_en_demo
template: qwen3
cutoff_len: 2048
max_samples: 1000
overwrite_cache: true
preprocessing_num_workers: 16
dataloader_num_workers: 4

### output
output_dir: saves/Qwen3-8B/full/sft
logging_steps: 1
save_steps: 20000
max_steps: 500
plot_loss: true
overwrite_output_dir: true
save_only_model: false
report_to: none  # choices: [none, wandb, tensorboard, swanlab, mlflow]

### train
per_device_train_batch_size: 8
gradient_accumulation_steps: 1
learning_rate: 1.0e-5
lr_scheduler_type: cosine
warmup_ratio: 0.1
bf16: true
ddp_timeout: 1800
resume_from_checkpoint: null
```
Test use_v1_kernels (enabling NPU fusion operators npu_rms_norm, npu_rope, npu_swiglu) and enable_liger_kernel (enabling rope, rms_norm, swiglu). The loss is aligned with slightly better performance, as shown in the figure below. The left side shows torch_npu fusion operators, and the right side shows liger_kernel operators (llamafactory does not implement the npu_cross_entropy fusion operator, so no comparison was conducted).


| torch_npu | liger_kernel |
|-----|-----|
| image | image|

sunyi0505:
@Tcc0403 This pr is ready for review. Could you please review this PR? Thanks!

zheliuyu:
> @Tcc0403 This pr is ready for review. Could you please review this PR? Thanks!

Add one more point for easier understanding: `torch_npu` is a separate package independent of `torch`, typically integrated into training frameworks such as `LLaMA‑Factory`, `verl`, and `ms‑swift`, serving as the performance baseline for NPU. Therefore, for Liger‑Kernel’s `triton‑ascend` kernels to be considered truly production‑ready for training software, their performance needs to surpass that of `torch_npu`. Due to the above reasons, we need to do this job

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -208,55 +209,68 @@ def liger_cross_entropy_forward_kernel_plain(
...
+            x_y = tl.load(logits_row_ptr + y_idx).cast(tl.float32)
+            m = float("-inf")
+            d = 0.0
+            for i in tl.range(0, n_full, BLOCK_SIZE):
+                x = tl.load(logits_row_ptr + i + cols).cast(tl.float32)
+                block_max = tl.max(x)
+                m_new = tl.maximum(m, block_max)
+                d = d * tl.exp(m - m_new) + tl.sum(tl.exp(x - m_new))
+                m = m_new
+            if HAS_TAIL:
```
Tcc0403, line 260: I'm curious why seperating loop body and tail instead of letting oob mask handle each load. Does ascend triton benefit a lot from removing oob check in most iterations?

I'm good with tail seperation structure, no need to change.

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
```diff
@@ -208,55 +209,68 @@ def liger_cross_entropy_forward_kernel_plain(
...
+            # OOB assert (queued after this kernel) raises AssertionError.
+            y_idx = tl.minimum(tl.maximum(y, 0), n_cols - 1)
+            x_y = tl.load(logits_row_ptr + y_idx).cast(tl.float32)
+            m = float("-inf")
+            d = 0.0
+            for i in tl.range(0, n_full, BLOCK_SIZE):
+                x = tl.load(logits_row_ptr + i + cols).cast(tl.float32)
+                block_max = tl.max(x)
+                m_new = tl.maximum(m, block_max)
+                d = d * tl.exp(m - m_new) + tl.sum(tl.exp(x - m_new))
```
Tcc0403, line 258: Does npu have hardware exp2 instruction which is more efficient than `tl.exp`? See #1266

src/liger_kernel/ops/backends/_ascend/ops/rms_norm.py
```diff
@@ -273,38 +266,102 @@ def _rms_norm_backward_kernel_no_tiling(
...
+    ROWS_PER_BLOCK: tl.constexpr,
+):
+    """2D fused backward: several rows x full n_cols."""
+    pid = tl.program_id(0)
+    num_progs = tl.num_programs(0)
+    col_offsets = tl.arange(0, BLOCK_SIZE)
+    row_offsets = tl.arange(0, ROWS_PER_BLOCK)
     col_mask = col_offsets < n_cols
-    row_offsets = tl.arange(0, BLOCK_SIZE_M)
+    n_cols_inv = 1.0 / n_cols
```
Tcc0403, line 371: Does ascend-triton have similar issue? #1358

src/liger_kernel/ops/backends/_ascend/ops/rms_norm.py
```diff
@@ -581,58 +592,72 @@ def rms_norm_forward(X, W, eps, offset, casting_mode):
...
+                RSTD,
+                RSTD.stride(0),
+                n_rows,
+                n_cols,
+                eps,
+                offset,
+                casting_mode,
+                elementwise_affine,
+                X_DTYPE,
+                BLOCK_SIZE=col_block,
```
Tcc0403: Let's pack them into a common dict args as bwd does, easier to track the diff of each kernels

src/liger_kernel/ops/backends/_ascend/ops/rope.py
```diff
@@ -1,238 +1,1125 @@
...
 from liger_kernel.ops.utils import get_npu_core_count
 
+_NUM_CORES = None
+
+
+def _npu_cores() -> int:
+    global _NUM_CORES
+    if _NUM_CORES is None:
+        _NUM_CORES = get_npu_core_count()
+    return _NUM_CORES
```
Tcc0403: Move it into a shared utils.py

src/liger_kernel/ops/backends/_ascend/ops/rope.py
```diff
@@ -253,10 +1141,9 @@ def backward(ctx, dq, dk):
         """
         dq size: (bsz, n_q_head, seq_len, head_dim)
         dk size: (bsz, n_kv_head, seq_len, head_dim)
-        cos size: (1, seq_len, head_dim) or (bsz, seq_len, head_dim)
-        sin size: (1, seq_len, head_dim) or (bsz, seq_len, head_dim)
         """
-
+        if dq is None and dk is None:
+            return None, None, None, None, None, None
```
Tcc0403, line 1007: when would they be both None?

src/liger_kernel/ops/backends/_ascend/ops/rope.py
```diff
@@ -1,238 +1,1125 @@
...
+
+def _npu_cores() -> int:
+    global _NUM_CORES
+    if _NUM_CORES is None:
+        _NUM_CORES = get_npu_core_count()
+    return _NUM_CORES
+
+
+@triton.jit
+def _triton_rope_aligned(
```
Tcc0403: There are too many kernel varaints for rope. 

Specialized kernels for both q/k layouts are acceptable, but current status would be difficult to maintain in future. 

Perhaps we can abstract out load/access patterns into helper `triton.jit` functions and share the common compute body to consolidate scattered variants

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
sunyi0505: Yes — on Ascend the full-tile path is unmasked so MTE2 stays sequential and the remainder predicate is DCE’d when HAS_TAIL is constexpr-false (V % BLOCK_SIZE == 0, the common training vocab sizes). Putting mask= on every load would keep the tail predicate live in the main loop and hurt the bandwidth-bound CE path. Leaving the split as-is.

src/liger_kernel/ops/backends/_ascend/ops/cross_entropy.py
sunyi0505: Ascend AIV has a native vector Exp (e^x). tl.exp2(x * log2(e)) is not a win here the way ex2.approx is on NVIDIA in #1266 — it adds a mul and still lowers through the same exp pipeline. We kept tl.exp. Happy to revisit if a later CANN/triton-ascend drop exposes a cheaper exp2.

src/liger_kernel/ops/backends/_ascend/ops/rms_norm.py
sunyi0505: Ascend Triton does not go through Inductor’s CUDA fp64 specialization, but the same Python-float scalars (eps / offset) are still in the kernels. Pinned them the same way as CUDA: NONE → input dtype, otherwise fp32, so rsqrt / W+offset cannot silently promote if torch.compile is used later.

src/liger_kernel/ops/backends/_ascend/ops/rms_norm.py
sunyi0505: Done — forward now builds a shared common dict and dispatches row / fused_2d / tiled the same way backward does.

src/liger_kernel/ops/backends/_ascend/ops/rope.py
sunyi0505: Moved the cached helper to src/liger_kernel/ops/backends/_ascend/ops/utils.py as get_npu_cores().

src/liger_kernel/ops/backends/_ascend/ops/rope.py
sunyi0505: ctx.set_materialize_grads(False) passes None for unused outputs. Both-None is the unused-outputs guard (torch.autograd.grad(..., grad_outputs=[None, None]) or neither stream used). A single None (only q or only k in the loss) is handled by _launch_rope.

src/liger_kernel/ops/backends/_ascend/ops/rope.py
sunyi0505: Agreed. Layout-specialized kernels stay (BNSD aligned/fullrow/masked, BSND flat/head-loop/masked, plus the short-seq qk two-phase launches), but load/store and rotate_half now go through shared @triton.jit helpers: _rotate_half, _decode_bh_stile, _load_cos_sin / _load_cos_sin_masked, _apply_rope_halves / _apply_rope_halves_masked / _apply_rope_fullrow / _apply_rope_flat. RoPE + RMSNorm tests passed locally on NPU.

10 launch kernels in ops/backends/_ascend/ops/rope.py have been consolidated into 4. The Python dispatch, BNSD/BSND branching, and two‑stage q‑k loops remain unchanged.

| Kernel | constexpr | Overwrite the original |
| -------| ----------- | ------------------------ |
|_triton_rope_bnsd | MASKED, USE_FULLROW | aligned / fullrow / masked |
| _triton_rope_bnsd_qk | USE_FULLROW | aligned_qk / fullrow_qk |
| _triton_rope_bsnd | MASKED, USE_FLAT | bsnd / bsnd_flat / bsnd_masked |
| _triton_rope_bsnd_qk | USE_FLAT | bsnd_qk / bsnd_qk_flat |

MASKED is still mutually exclusive with USE_FULLROW/USE_FLAT via dispatch, so no invalid specializations are generated. The q‑k loop still iterates over q first then k, with no pointer predicates. llama4 / mrope are not merged.

Correctness has been verified against HF 32/8, Qwen 28/4, tail block S=17, contiguous fullrow, small shape S=2, non‑power‑of‑two hd=92, and the corresponding fp16 cases in the official test_rope (including 423×73×92). Both forward and backward passes are aligned.

src/liger_kernel/ops/backends/_ascend/ops/utils.py
```diff
@@ -0,0 +1,13 @@
+"""Shared Ascend host helpers."""
+
+from liger_kernel.ops.utils import get_npu_core_count
```
Tcc0403: forgot there's one already, use it instead and add built-in cache decorator

src/liger_kernel/ops/backends/_ascend/ops/rope.py
```diff
@@ -1,238 +1,946 @@
...
+    return new_left, new_right
+
+
+@triton.jit
+def _decode_bh_stile(tile_id, n_s_tiles, n_heads):
+    s_tile = tile_id % n_s_tiles
+    tmp = tile_id // n_s_tiles
+    h = tmp % n_heads
+    batch_idx = tmp // n_heads
+    return batch_idx, h, s_tile
```
Tcc0403: nit
```suggestion
@triton.jit
def _get_work_tile_coord(tile_id, n_s_tiles, n_heads):
    seq_idx = tile_id % n_s_tiles
    tmp = tile_id // n_s_tiles
    head_idx = tmp % n_heads
    batch_idx = tmp // n_heads
    return batch_idx, head_idx, seq_idx
```

src/liger_kernel/ops/backends/_ascend/ops/rope.py
```diff
@@ -1,238 +1,946 @@
...
+def _launch_rope(q, k, cos, sin, backward: bool, q_out=None, k_out=None):
+    if q is None and k is None:
+        return None, None, cos, sin
+    cos, sin = _as_bnsd_cos_sin(cos, sin)
+    cos = _ensure_innermost_contiguous(cos)
+    sin = _ensure_innermost_contiguous(sin)
+    use_bsnd = (q is None or _is_free_bsnd(q)) and (k is None or _is_free_bsnd(k))
+    if use_bsnd:
+        return _launch_bsnd(q, k, cos, sin, backward)
+    return _launch_bnsd(q, k, cos, sin, backward, q_out=q_out, k_out=k_out)
```
Tcc0403, line 972: why doesn't bsnd variant support pre-allocated output tensors?

If I understand it correctly, q_out and k_out are designed for storing inplace in backward path. Both bsnd and bnsd layout have inplace mechanism, however, `_launch_bnsd `explicitly accepts input tensors as output tensors args, but `_launch_bsnd` completely ignores `q_out`/`k_out` and implicitly stores inplace when `backward==True`.

I prefer both having output tensors args, so users can decide whether or not passing input tensors at those positions for inplace store.

src/liger_kernel/ops/backends/_ascend/ops/rope.py
```diff
@@ -1,238 +1,946 @@
...
 
-    n_row = batch_size * seq_len
 
-    # ensure dq and dk are contiguous
-    dq = dq.contiguous()
-    dk = dk.contiguous()
+def _hf_like(t):
+    """Allocate (B, H, S, D) with BSHD storage so transpose(1, 2) is contiguous."""
+    b, h, s, d = t.shape
+    return torch.empty((b, s, h, d), device=t.device, dtype=t.dtype).transpose(1, 2)
```
Tcc0403, line 875: doesn't `torch.empty_like(t)` produce transposed layout?

src/liger_kernel/ops/backends/_ascend/ops/rope.py
```diff
@@ -1,238 +1,946 @@
...
+    seq_len,
+    n_s_tiles,
+    total_tiles,
+    hd: tl.constexpr,
+    BLOCK_S: tl.constexpr,
+    BLOCK_H: tl.constexpr,
+    BLOCK_D: tl.constexpr,
+    BACKWARD_PASS: tl.constexpr,
+    MASKED: tl.constexpr,
+    USE_FLAT: tl.constexpr,
```
Tcc0403, line 416: I'm not entirely understand what USE_FLAT path does and why/when we need it. 
```py
def _can_flat_bsnd(n_heads: int, block_s: int) -> bool:
    rows = block_s * n_heads
    return n_heads == triton.next_power_of_2(n_heads) and rows == triton.next_power_of_2(rows)
```
The condition of flat path is n_heads and rows are both power of 2 (unpadded). Could you elaborate more on how it differs from MASKED path?

src/liger_kernel/ops/backends/_ascend/ops/utils.py
sunyi0505: Delete src/liger_kernel/ops/backends/_ascend/ops/utils.py
• Add @functools.cache to get_npu_core_count
• Modify rope.py to directly import get_npu_core_count (consistent with other Ascend ops)

src/liger_kernel/ops/backends/_ascend/ops/rope.py
sunyi0505: _decode_bh_stile → _get_work_tile_coord, rename the variable to seq_idx / head_idx

src/liger_kernel/ops/backends/_ascend/ops/rope.py
sunyi0505: Add q_out / k_out parameters to _launch_bsnd, aligning its behavior with _launch_bnsd
• _launch_rope passes through parameters; backward still performs in‑place operations via q_out=dq, k_out=dk

src/liger_kernel/ops/backends/_ascend/ops/rope.py
sunyi0505: It is stated in the _hf_like documentation that empty_like(t) allocates a standard contiguous (B,H,S,D), while HF models adopt the BSHD strided layout. Therefore, (B,S,H,D).transpose(1,2) must be explicitly allocated.
torch.empty_like(t)`` would allocate standard contiguous ``(B, H, S, D)`` storage because ``t`` is already logically BNSD. The kernel indexes with ``t``'s strided BSHD layout, so a contiguous ``(B, H, S, D)`` buffer would store results at the wrong offsets.

src/liger_kernel/ops/backends/_ascend/ops/rope.py
sunyi0505: Add documentation in _can_flat_bsnd and _triton_rope_bsnd to distinguish three paths:
◦ MASKED: The sequence tail requires masking
◦ USE_FLAT: Both H and S×H are powers of two, flattened into a single‑vector load of (BLOCK_S*H, D)
◦ Head‑loop: Aligned but H is not a power of two (e.g., GQA H=8), uses head loop with no tail mask

Tcc0403 (approved):
LGTM

zheliuyu:
> > @Tcc0403 This pr is ready for review. Could you please review this PR? Thanks!
> 
> Add one more point for easier understanding: `torch_npu` is a separate package independent of `torch`, typically integrated into training frameworks such as `LLaMA‑Factory`, `verl`, and `ms‑swift`, serving as the performance baseline for NPU. Therefore, for Liger‑Kernel’s `triton‑ascend` kernels to be considered truly production‑ready for training software, their performance needs to surpass that of `torch_npu`. Due to the above reasons, we need to do this job

The previously mentioned experimental data now looks quite complete — both the kernel-level numbers and the `LLaMA-Factory` e2e results.

Looking forward to this landing. An end-to-end experiment with `ms-swift` will follow, and the results will be shared here.

PR #1428, opened 2026-08-31
[CUDA] Add tensor-parallel fused scaled cross entropy frontend

kolehma8:
## Summary

- Add `LigerFusedLinearScaledCrossEntropyTPFunction` and `liger_fused_linear_scaled_cross_entropy_tp`, with `tp_group` passed as a call-time argument.
- Route Hopper BF16 inputs to the optional `liger_cute_kernels` implementation when its FSLCE runtime surface is available.
- Keep the fallback entirely in Liger using a chunked tensor-parallel implementation adapted from Verl's fused PPO formulas; there is no runtime dependency on the Verl package.
- Match both paths on global per-token NLL/entropy, ignored targets, local `dW`, and TP-summed `dX`.
- Leave NVSHMEM bootstrap and team preparation application-owned. FSLCE resolves only the call-time `tp_group`, so it coexists with LigerMoE using a different `ep_group` without frontend-global process-group state.
- Remove the native NCCL dependency and all Torch communicator internals from the `liger_cute_kernels` path. Forward and backward communicate exclusively through the prepared NVSHMEM team, and the TVM-FFI core remains Torch-ABI-independent.
- Extend the reusable local reduction API with FP32 `MAX`. The forward WGMMA kernel performs split last-arrival merging, local MAX, correction, and local SUM in its epilogue.
- For two-host execution, each local GPU sends one unique contiguous token shard over IBRC; the follow-up kernel merges the remote online-softmax state, gathers shards with NVLS, and writes NLL/LSE/entropy.
- Bind native FFI calls to the input CUDA device so autograd worker threads have a valid context for TMA descriptor construction.

## Validation

- `make checkstyle`
- `python -m pytest test/transformers/test_fused_linear_scaled_cross_entropy.py -q` - 15 passed, 1 skipped
- H200 distributed frontend matrix using `torch.multiprocessing.spawn` - 10 passed in 77.45s:
  - WORLD process groups at TP1, TP2, TP4, and TP8
  - 4-GPU contiguous subgroups `[0,1]` / `[2,3]`
  - 4-GPU strided subgroups `[0,2]` / `[1,3]`
  - Built-in Liger fallback at TP2 WORLD and strided TP4
  - LigerMoE on `ep_group=WORLD` followed by FSLCE on independent strided TP groups in the same four processes
  - DirectPeer fallback with NVLS disabled
- Native forward self-test: aligned, entropy, ragged M/H/V, temperature, and large split-N cases passed.
- Raw TVM-FFI world/subgroup tests, including forward and backward CUDA Graph replay: 4 passed.
- Two-host TP16 IBRC random-data correctness: passed.
- `libliger_cute_kernels.so` has no NCCL dynamic dependency.

### Forward performance

M=8192, H=4096, Vglobal=131072, entropy disabled:

| TP | NVSHMEM p50 | Previous p50 | Delta |
|---:|---:|---:|---:|
| 1 | 11.227 ms | 11.148 ms | +0.7% |
| 2 | 5.639 ms | 5.594 ms | +0.8% |
| 4 | 2.832 ms | 2.821 ms | +0.4% |
| 8 | 1.442 ms | 1.453 ms | -0.8% |
| 16 (two-host IBRC) | 0.829 ms | 0.784 ms | +5.7% |

TP1-8 are within the agreed 1% run-to-run noise threshold; the TP16 IBRC gap is accepted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,10 +264,9 @@ nll, entropy = LigerFusedLinearScaledCrossEntropyTPFunction.apply(
 The TP frontend derives `vocab_start` from the process-group rank. BF16 inputs on Hopper use the optional
 `liger_cute_kernels` (LCK) implementation when its native core is available; other configurations use Liger's chunked
 tensor-parallel fallback adapted from Verl's fused PPO formulas. Both paths return globally correct per-token outputs
-and sum the input gradient across the TP group. LCK borrows the caller-owned PyTorch NCCL communicator only when the
-PyTorch and LCK NCCL runtime versions match; the TVM-FFI core remains independent of the Torch C++ ABI and never owns or
-destroys that communicator. Calls remain collectives on `tp_group`, so they must have the same ordering on every member
-and must not be reordered around other collectives on that group.
+and sum the input gradient across the TP group. The LCK forward and backward paths communicate exclusively through
+NVSHMEM; no Torch communicator or Torch C++ ABI crosses the TVM-FFI boundary. Calls remain collectives on `tp_group`, so
+they must have the same ordering on every member.
 
 LCK shares the process-wide NVSHMEM bootstrap used by `LigerExpertParallelFusedMoe`, while TP and EP use separate cached
 teams and separately named workspaces. Application setup must initialize NVSHMEM from a common parent group (normally
--- a/README.md
+++ b/README.md
@@ -250,6 +250,25 @@ The implementations share this public contract but use different schedules:
 - **CuTe SM90** uses `LigerFusedScaledCrossEntropySM90Function` for BF16 inputs on Hopper. Its sole forward uses the fixed cluster-M2 N160 fragment kernel, with a measured split-N lookup for profiled long-sequence shapes, and never writes logits to HBM; `m_tiles_per_cluster` remains accepted for API compatibility but does not change that schedule. Backward runs `dZ`, `dX`, and `dW` in one persistent cluster kernel with a reusable 1024-token `dZ` workspace.
 - **Fallback** uses a 512-token chunked PyTorch implementation adapted from Verl's fused PPO formulas when the default frontend cannot use the SM90 kernel.
 
+For a classifier weight sharded into equally sized contiguous ranges across a tensor-parallel process group, use the
+TP frontend. Targets remain global vocabulary indices:
+
+```python
+from liger_kernel.ops import LigerFusedLinearScaledCrossEntropyTPFunction
+
+nll, entropy = LigerFusedLinearScaledCrossEntropyTPFunction.apply(
+    x, local_weight, target, tp_group, 1.0, -100, 1, True
+)
+```
+
+The TP frontend derives `vocab_start` from the process-group rank. BF16 inputs on Hopper use the optional
+`liger_cute_kernels` (LCK) implementation when its native core is available; other configurations call Verl's
+tensor-parallel linear cross entropy. Both paths return globally correct per-token outputs and sum the input gradient
+across the TP group. The first LCK call initializes its NVSHMEM/NCCL runtime and fixes workspace capacity, so warm it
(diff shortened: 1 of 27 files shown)


liger_cute_kernels/liger_cute_kernels/nvshmem.py
```diff
@@ -135,17 +142,31 @@ def init_pmi(pg: Optional["torch.distributed.ProcessGroup"] = None) -> None:
...
 
 
+def is_initialized() -> bool:
+    """Return whether NVSHMEM was initialized through this module."""
+    return _INITIALIZED
+
+
+def ensure_initialized(pg: Optional["torch.distributed.ProcessGroup"] = None) -> None:
+    """Initialize NVSHMEM from ``pg`` once, leaving an existing runtime intact."""
+    if not _INITIALIZED:
```
yueyiming2009: Refactor _INITIALIZED to better cover different pgs

src/liger_kernel/ops/cute/fused_linear_scaled_cross_entropy_tp.py
```diff
@@ -0,0 +1,251 @@
...
+            team_handle,
+        )
+        tvm_ffi.fused_linear_scaled_cross_entropy_configure_forward(
+            tokens,
+            local_vocab,
+        )
+    return team_handle
+
+
+class LigerFusedLinearScaledCrossEntropyLckTPFunction(torch.autograd.Function):
```
yueyiming2009, line 123: nit: why do you want to have Lck in the name?

src/liger_kernel/ops/cute/fused_linear_scaled_cross_entropy_tp.py
```diff
@@ -0,0 +1,251 @@
...
+        yield
+        return
+    with torch.cuda.device(tensor.device):
+        torch.cuda.set_device(tensor.device)
+        yield
+
+
+def _borrow_nccl_communicator(process_group: "ProcessGroup", device: torch.device) -> int:
+    """Borrow the current-device communicator without storing or owning it."""
+    backend = process_group._get_backend(device)
```
yueyiming2009: Could we avoid relying on private `ProcessGroupNCCL` APIs here (`_get_backend`, `_comm_ptr`, and `eager_connect_single_device`)? These are outside PyTorch’s compatibility contract and may change across supported PyTorch versions. Since the backward path already uses an NVSHMEM team, would using NVSHMEM for the forward collectives be viable? Alternatively, could LCK initialize and cache a second NCCL communicator for the supplied TP group? If borrowing the PyTorch communicator is materially better, it would be helpful to include quantitative setup/steady-state measurements and ensure that unavailable or changed private APIs cleanly select the fallback.

yueyiming2009 (approved):

README.md
```diff
@@ -250,6 +250,30 @@ The implementations share this public contract but use different schedules:
...
+```python
+from liger_kernel.ops import LigerFusedLinearScaledCrossEntropyTPFunction
+
+nll, entropy = LigerFusedLinearScaledCrossEntropyTPFunction.apply(
+    x, local_weight, target, tp_group, 1.0, -100, 1, True
+)
+```
+
+The TP frontend derives `vocab_start` from the process-group rank. BF16 inputs on Hopper use the optional
+`liger_cute_kernels` (LCK) implementation when its native core is available; other configurations use Liger's chunked
```
yueyiming2009, line 265: Could we generalize the terminology here and avoid the overloaded `LCK` acronym? The surrounding package and module descriptions still characterize `liger_cute_kernels` as MoE-specific infrastructure ported from LigerCommKernels, which makes this wording sound as though FLSCE depends on an MoE communication kernel. This PR is instead expanding the package into a shared home for native distributed CUTLASS/CuTe kernels. Please update the package/module documentation and metadata to reflect that broader role, and use a descriptive term such as “native CuTe backend (`liger_cute_kernels`)” consistently instead of `LCK`.

PR #1438, opened 2026-09-03
fix(fused_linear_jsd): project logits in FP32 before the matmul, not after

### LINKED ISSUE #1432: fused_linear_jsd rounds logits to the input dtype before the documented FP32 cast, costing up to 23% gradient error in bf16

JeanMaximilienCadic:
`fused_linear_jsd_forward` casts the student/teacher logits to FP32 *after* the projection matmul — but the matmul already ran in bf16/fp16, so the cast just relabels a value that's already been rounded to 8/11 mantissa bits. The "compute in FP32" comment above it is effectively a no-op.

JSD's gradient is a difference of nearly-equal distributions, so that rounding cancels catastrophically. At a realistic logit spread (std≈30) I measured bf16 gradients off by up to ~20% versus a true FP32 reference — well past the tolerance CE gets away with for the same pattern.

cuBLAS already accumulates the GEMM in FP32 internally, so the fix just keeps that accumulator instead of discarding it: `torch.mm(..., out_dtype=torch.float32)` on torch >= 2.8 / sm_80+, with an explicit upcast fallback otherwise. Same compute and memory, correct logits.

I also fixed the transformers-level test oracle, which shared the exact same cast-after-matmul defect and was matching the kernel bug-for-bug (so CI never caught it), and added a regression test at the issue's repro shape/scale.

Out of scope (separate follow-ups, as the issue notes): the same pattern in `fused_linear_distillation.py` / `fused_linear_ppo.py`, and the `grad_logits` downcast before the backward matmuls.

Verification (RTX 4090, torch 2.11+cu130, triton 3.6): grad error shrinks in every case, e.g. bf16 β=0 std=30 → grad_input 13.6%→7.5%, grad_weight 15.2%→5.1%.

Fixes #1432.

### DIFF
--- a/src/liger_kernel/ops/fused_linear_jsd.py
+++ b/src/liger_kernel/ops/fused_linear_jsd.py
@@ -23,6 +23,39 @@ from liger_kernel.utils import infer_device
 MAX_FUSED_SIZE = 4096 if infer_device() == "xpu" else 65536 // 2
 _TORCH_VERSION = Version(torch.__version__.split("+")[0])
 _ADDMM_SUPPORTS_OUT_DTYPE = _TORCH_VERSION >= Version("2.8.0")
+# torch.mm(..., out_dtype=...) landed alongside the addmm variant (2.8.0) and
+# has the same sm_80+ (Ampere) requirement.
+_MM_SUPPORTS_OUT_DTYPE = _ADDMM_SUPPORTS_OUT_DTYPE
+
+
+def _project_logits_fp32(input_chunk: torch.Tensor, weight: torch.Tensor) -> torch.Tensor:
+    """Project ``input_chunk @ weight.t()`` and return it in FP32.
+
+    The naive ``(input_chunk @ weight.t()).to(torch.float32)`` casts *after* the
+    matmul, so on a bf16/fp16 GEMM the output is already rounded to 8/11
+    mantissa bits before the cast ever runs -- the cast changes the container,
+    not the contents. cuBLAS already accumulates the GEMM in FP32 internally;
+    this just surfaces that accumulator instead of throwing it away on the
+    store.
+
+    ``torch.mm(..., out_dtype=torch.float32)`` (torch >= 2.8, CUDA sm_80+)
+    does this at no extra cost: a bf16xbf16 or fp16xfp16 product is exactly
+    representable in FP32 (8+8 / 11+11 mantissa bits both fit in 24), so it is
+    numerically identical to upcasting both operands first, while keeping the
+    low-precision tensor cores and avoiding an FP32 copy of the (potentially
+    huge, V x H) weight. When that path isn't available (older torch, no
+    CUDA, or CUDA < sm_80), fall back to the explicit upcast -- slower, but
+    still correct, unlike casting after the matmul.
+    """
+    if (
+        _MM_SUPPORTS_OUT_DTYPE
+        and input_chunk.is_cuda
+        and input_chunk.dtype in (torch.float16, torch.bfloat16)
+        and weight.dtype == input_chunk.dtype
+        and torch.cuda.get_device_capability(input_chunk.device)[0] >= 8
+    ):
+        return torch.mm(input_chunk, weight.t(), out_dtype=torch.float32)
+    return input_chunk.float() @ weight.float().t()
 
 
 def _max_capability() -> int:
@@ -133,9 +166,11 @@ def fused_linear_jsd_forward(
 
         # shape: chunk_size x V
         # For anything starting from logits to the final JSD loss, we do computation
-        # in FP32 to avoid losing numerical stability.
-        student_logits_chunk = (student_input_chunk @ student_weight.t()).to(torch.float32)
-        teacher_logits_chunk = (teacher_input_chunk @ teacher_weight.t()).to(torch.float32)
+        # in FP32 to avoid losing numerical stability. The projection itself must
+        # also run (or be recovered) in FP32: casting *after* a bf16/fp16 matmul
+        # only relabels an already-rounded result (see _project_logits_fp32).
+        student_logits_chunk = _project_logits_fp32(student_input_chunk, student_weight)
+        teacher_logits_chunk = _project_logits_fp32(teacher_input_chunk, teacher_weight)
 
         # log-softmax with temperature
         student_logits_chunk = student_logits_chunk / temperature
--- a/test/ops/test_fused_linear_jsd.py
+++ b/test/ops/test_fused_linear_jsd.py
@@ -63,15 +63,28 @@ def _fused_linear_jsd_ref(
     student_log_probs = torch.log_softmax(student_logits, dim=-1)
     teacher_log_probs = torch.log_softmax(teacher_logits, dim=-1)
 
-    student_probs = student_log_probs.exp()
-    teacher_probs = teacher_log_probs.exp()
-
-    m = beta * student_probs + (1.0 - beta) * teacher_probs
-    log_m = m.log()
-
-    # JSD = beta * KL(student || m) + (1 - beta) * KL(teacher || m)
-    loss = beta * (student_probs * (student_log_probs - log_m)).sum(dim=-1)
-    loss += (1.0 - beta) * (teacher_probs * (teacher_log_probs - log_m)).sum(dim=-1)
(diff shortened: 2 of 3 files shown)


kashif:
@JeanMaximilienCadic  there is already a PR that fixes this issue: #1433

PR #1465, opened 2026-09-15
Publish multi-architecture Liger Cute wheel on releases

kolehma8:
## Summary

- add an independent release workflow for `liger-cute-kernels` using the pinned CUDA 12.9.1 Ubuntu 22.04 build image and CUTLASS v4.4.1
- build SM90a and SM100f cores in parallel with four compiler jobs, then package both into one Python-ABI-independent wheel
- use the same public version as `liger-kernel` and publish only on release events; manual dispatches build artifacts without publishing
- stop bundling NVSHMEM shared libraries and depend on `nvidia-nvshmem-cu12==3.6.5`
- select the SM90a or SM100f core at runtime from the active GPU capability while retaining legacy single-core wheel compatibility
- validate FSLCE forward/backward exports, wheel contents, version, dependency metadata, manylinux tag, and the 100 MiB PyPI limit

## Testing done

- `ruff check` and `ruff format --check` on the changed Python files
- 17 targeted packaging/loader tests passed
- workflow YAML parsed successfully and `git diff --check` passed
- built both native cores on H200 pods with `LIGER_CUTE_BUILD_JOBS=4`:
  - SM90a: 45m09s, peak compiler RSS 8.69 GiB
  - SM100f: 2m42s, peak compiler RSS 7.42 GiB
- confirmed both cores export fused linear scaled cross-entropy forward and backward APIs
- packaged one dual-architecture wheel: 87,726,555 bytes (83.66 MiB), below PyPI's 100 MiB limit
- installed the wheel in an isolated target directory with `nvidia-nvshmem-cu12==3.6.5` and verified on H200:
  - selected `libliger_cute_kernels_sm90a.so`
  - no NVSHMEM libraries were bundled in the LCK package
  - NVSHMEM unique-ID initialization reported rank 0 / world size 1 and finalized successfully
- `twine check` passed without warnings

The H200 measurement pod uses glibc 2.38, so its validation artifact was tagged `py3-none-manylinux_2_38_x86_64`. The release workflow builds inside the digest-pinned Ubuntu 22.04 image, asserts glibc 2.35, and emits `py3-none-manylinux_2_35_x86_64`.


.github/workflows/publish-liger-cute-kernels.yml
```diff
@@ -0,0 +1,277 @@
+name: Publish Liger Cute Kernels on Release
+
+on:
```
vaibhavjindal, line 3: should we do:

on:
  push:
    branches:
      - main

.github/workflows/publish-nightly.yml
```diff
@@ -47,3 +47,7 @@ jobs:
     - name: Create release notes
       run: |
         echo "Nightly build published to PyPI with the name 'liger-kernel-nightly'."
+
```
vaibhavjindal: Should we remove this and have a separate run condition for publish-liger-cute-kernels-nightly.yml??

.github/workflows/publish-liger-cute-kernels.yml
```diff
@@ -0,0 +1,277 @@
...
+        name: liger-cute-kernels-wheel
+        path: liger_cute_kernels/dist/*.whl
+        if-no-files-found: error
+        retention-days: 7
+
+  publish-wheel:
+    name: Publish Liger Cute Kernels to PyPI
+    # Reusable nightly runs build and validate the wheel but never receive
+    # publishing credentials or execute this job.
+    if: github.event_name == 'release'
```
vaibhavjindal: Make sure it won't run on a nightly release.

vaibhavjindal (commented):
Added some comments

.github/workflows/publish-liger-cute-kernels.yml
kolehma8: Done in 38657c2. The LCK workflow now has its own push trigger for the main branch.

.github/workflows/publish-nightly.yml
kolehma8: Done in 38657c2. I removed the reusable-workflow job from publish-nightly.yml; the LCK workflow now runs independently on main pushes.

.github/workflows/publish-liger-cute-kernels.yml
kolehma8: Confirmed and strengthened in 38657c2. The upload job now requires both github.event_name == release and github.event.action == published, so main-push integration builds and manual dispatches skip Twine.

vaibhavjindal (approved):
LGTM!

PR #1467, opened 2026-09-15
Fix remaining LCK container build dependencies

kolehma8:
## Summary

- replace GitHub toolcache Python with Ubuntu 22.04's native Python 3.10 inside the CUDA build containers
- install `libibverbs-dev` so the PyPI NVSHMEM IBGDA headers can include `infiniband/mlx5dv.h`
- retain GitHub-hosted Python 3.12 and pip caching for the non-container wheel packaging job

## Failures addressed

PR #1466 fixed the first `setup-python` cache-hook failure, but the subsequent pre-merge workflow runs exposed two additional container dependencies:

1. GitHub's Python 3.12 toolcache binary requires glibc 2.38, while the digest-pinned CUDA Ubuntu 22.04 image provides glibc 2.35.
2. Both native architectures require `infiniband/mlx5dv.h` through NVSHMEM's IBGDA headers; Ubuntu supplies it in `libibverbs-dev`.

The native core is Python-ABI-independent, and `apache-tvm-ffi==0.1.14` provides a CPython 3.10 manylinux wheel.

## Validation

Pre-merge fork dispatch: https://github.com/kolehma8/Liger-Kernel/actions/runs/35032379309

- dependency installation and CUDA 12.9 configuration pass for both architectures
- SM100f core build and artifact upload pass in 6m34s
- SM90a all-config core/cubin build is still running
- the manual dispatch does not execute the Twine publication job


Mecoli1219 (dismissed):

kolehma8:
Validation update: fork run 35032379309 completed both native builds successfully (SM90a 1h00m22s, SM100f 6m34s). Packaging then exposed duplicate cubins because the uploaded artifact included the nested CMake tree. Commit 95bc894 moves the build tree outside release-core and scopes prebuilt cubin lookup to the SM90 directory. Repackaging the completed GitHub artifacts with this fix produced one 81,204,827-byte manylinux wheel containing both cores and exactly one SM90 cubin.

yueyiming2009 (approved):

